PR #10, opened 2015-11-24
NEW Rollback elements on page rollback

nyeholt:
- Rollback, but only if it's a Live => Stage rollback, because version based
  rollback is a little more difficult to determine
- Fixed an issue if the page class changes from one that has Elemental to one
  that doesn't

### DIFF
--- a/code/extensions/ElementPageExtension.php
+++ b/code/extensions/ElementPageExtension.php
@@ -116,25 +116,28 @@ class ElementPageExtension extends DataExtension {
 			}
 		}
 
-		$elements = $this->owner->ElementArea();
+		if ($this->owner->hasMethod('ElementArea')) {
+			$elements = $this->owner->ElementArea();
 
-		if(!$elements->isInDB()) {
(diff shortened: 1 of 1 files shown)


code/extensions/ElementPageExtension.php
```diff
@@ -208,4 +211,34 @@ public function onAfterPublish() {
...
+     * @param string $version
+     * @return null
+     */
+    public function onBeforeRollback($version) {
+        if ($version !== 'Live') {
+            // we don't yet have a smart way of rolling back to a specific version
+            return;
+        }
+        if($id = $this->owner->ElementAreaID) {
+			$widgets = Versioned::get_by_stage('BaseElement', 'Live', "ParentID = '$id'");
```
wilr: Cheers Marcus for the contribution. Something up with the formatting?

nyeholt:
Er, yeah, editor is now all ugly psr2. I didn't want to reformat all of it too!

PR #33, opened 2016-06-03
fix(FrontendFields): Fix calling getCMSFields on for frontend fields.

SilbinaryWolf:
- Calling getCMSFields on the frontend should not attempt to write a 'ElementArea'/'Page' if the page doesn't exist.
- Setup UnsavedRelationList for cases where ElementArea()->Elements() is empty. (ie. frontend fields / empty data object)


wilr:
Thanks :) Your project using frontend editing? Might be worth adding unit tests in the future for your use case to ensure any future changes don't break this.

PR #43, opened 2016-08-02
Don't allow unlinking for virtual elements

tamelion:


wilr:
@tamelion Thanks, IMHO a better check would be $record instanceof ElementVirtualLinked that way we'll capture subclasses (if someone does)

tamelion:
@wilr sounds sensible, thanks. Have pushed change.

PR #45, opened 2016-08-29
Versioneddataobject update

adrexia:
Update dependency to use name-spaced version of VersionedDataObject


wilr:
Thanks but reckon #46 was a slightly better approach

adrexia:
True. I think the other thing that was in the namespace was an extension so the chances of it appearing /needing to be used in the same file are quite unlikely

PR #48, opened 2016-09-02
Disable better buttons on Elemental

Leapfrognz:
As this interferes with versioning, note it also removes the prev/next and add record buttons.


Leapfrognz:
Implemented a better solution on https://github.com/unclecheese/silverstripe-gridfield-betterbuttons/pull/144

PR #50, opened 2016-09-18
fix(ElementalPageExtension): Make ElementArea gridfield insert before 'Metadata' if it exists.

SilbinaryWolf:
fix(ElementalPageExtension): Make ElementArea gridfield insert before 'Metadata' if it exists.


SilbinaryWolf:
Any particular reason for this not being pulled back in? Running a few 3.2 sites, and the ElementArea GridField looks ugly when its placed underneath the 'Metadata' ToggleCompositeFIeld.

I assume that field must either be gone for you DNADesign guys or it was removed in a later SS version.

wilr:
No reason apart from I didn't see it..

PR #63, opened 2016-12-14
Added multi theme support

mediabeastnz:
Added a small bit of code to the function that renders the element.
If now will check the current site config for a theme which should be set, this works nicely when you have multiple themes or are using Subsites which have different configs.


mediabeastnz:
I stumbled across this when I was using multiple themes... it throws a warning that it cannot find the template in the main theme but i didn't have an element template in my main theme on the secondary theme.

PR #71, opened 2017-04-30
Fix/publishing permissions

pitchandtone:


wilr:
What was the thinking behind removing versioned data objects? Just too buggy? 4.0's changeset feature would be worth having this (i.e bundling up a list of changes to blocks in 1 publish)

pitchandtone:
Sorry no, that was a mistake, let me pull some more together and push hit back.

Headline is we don't need versioned UI anywhere due to nested publish.

pitchandtone:
OK so now I've reviewed a bunch of other things and optimised a bunch of areas. Keen for comments @wilr @Leapfrognz

Leapfrognz:
Looks good. Keen for this to be merged asap. @wilr

SilbinaryWolf:
This will break any checks that use 'has_extension("VersionedDataObject")' as has_extension does not work properly with subclasses.

At least in 3.2 or lower.

pitchandtone:
@SilbinaryWolf We've created a pull request for that: https://github.com/silverstripe/silverstripe-cms/pull/1810

wilr:
This has been integrated into the master branch / ss4 support and partially improved by changes in core for dealing with children.

PR #74, opened 2017-05-09
Remove unnecessary page ID from preview link

tamelion:


adrexia:
After investigation, we think this might be caused by using either older version of cms/framework or mismatched versions. But this fix unfortunately breaks things for newer/matched silverstripe versions.

PR #82, opened 2017-06-28
FIX Update src directory names for PSR-4 integrity

robbieaverill:
Once you've implemented a PSR-4 autoloader in composer.json, the directory names should match the namespaces e.g. "Controllers" rather than "controllers" - this PR renames them.

---

Please note that the "SilverStripe" vendor namespace is reserved for core packages - perhaps consider changing your namespaces to "DNADesign" or similar


pitchandtone:
Have fixed namespace in another commit

PR #99, opened 2017-08-17
UPDATE: PSR-2 Compliance across all PHP files

adover:


wilr:
Looks like this is sorted in ss4

PR #108, opened 2017-10-03
Add missing namespaces, add lint config and update readme.

wilr:


CHANGELOG.md
```diff
@@ -0,0 +1,9 @@
+# Changelog
+
+All notable changes to this project will be documented in this file. This project adheres to [Semantic Versioning](http://semver.org/).
+
+## [2.x-dev]
+
+* SilverStripe 4.0 support.
+
+## [1.x-dev]
```
robbieaverill: Personally I'm not a fan of having changelogs in the code - they get forgotten about fairly often. I prefer putting release notes in the GitHub releases instead, but just my two cents

CONTRIBUTING.md
```diff
@@ -0,0 +1,25 @@
...
+
+See our [high level overview](http://silverstripe.org/contributing-to-silverstripe) on silverstripe.org on how you can
+help out.
+
+Or, for more detailed guidance, read one of the following pages:
+
+ * [Sharing your opinion and raising issues](http://docs.silverstripe.org/en/3.2/contributing/issues_and_bugs/)
+ * [Providing code, whether it's creating a feature or fixing a bug](http://docs.silverstripe.org/en/3.2/contributing/code/)
+ * [Writing and translating documentation](http://docs.silverstripe.org/en/3.2/contributing/translations/)
+ * [Translating user-interface elements](http://docs.silverstripe.org/en/3.2/contributing/translation_process/)
```
robbieaverill: We should use canonical URLs here (remove the 3.2 part)

README.md
```diff
@@ -1,13 +1,16 @@
 # SilverStripe Elemental
 
 [![Build Status](http://img.shields.io/travis/dnadesign/silverstripe-elemental.svg?style=flat-square)](https://travis-ci.org/dnadesign/silverstripe-elemental)
+[![Scrutinizer Code Quality](https://scrutinizer-ci.com/g/dnadesign/silverstripe-elemental/badges/quality-score.png?b=master)](https://scrutinizer-ci.com/g/dnadesign/silverstripe-elemental/?branch=master)
+[![Code Coverage](https://scrutinizer-ci.com/g/dnadesign/silverstripe-elemental/badges/coverage.png?b=master)](https://scrutinizer-ci.com/g/dnadesign/silverstripe-elemental/?branch=master)
+[![Build Status](https://scrutinizer-ci.com/g/dnadesign/silverstripe-elemental/badges/build.png?b=master)](https://scrutinizer-ci.com/g/dnadesign/silverstripe-elemental/build-status/master)
```
robbieaverill: We don't really need Scrutinizer build badges, just the code quality badge

phpcs.xml.dist
```diff
@@ -0,0 +1,25 @@
...
+        <exclude name="Squiz.Scope.MethodScope" />
+        <exclude name="Squiz.Classes.ValidClassName.NotCamelCaps" />
+        <exclude name="Generic.Files.LineLength.TooLong" />
+        <exclude name="PEAR.Functions.ValidDefaultValue.NotAtEnd" />
+    </rule>
+
+    <!-- PHP-PEG generated file not intended for human consumption -->
+    <exclude-pattern>*/SSTemplateParser.php$</exclude-pattern>
+</ruleset>
+
```
robbieaverill: Probably a wider discussion, but do we include these in each module? They're likely to change and become outdated I guess. Travis config is using the framework's version for CI builds, and the base SS project will include a copy as well, maybe we should omit it from modules?

.travis.yml
```diff
@@ -1,31 +1,37 @@
...
   - phpenv rehash
   - phpenv config-rm xdebug.ini
 
-  - composer install --prefer-dist
-  - composer require --prefer-dist --no-update silverstripe/installer 4.0.x-dev
-  - if [[ $DB == PGSQL ]]; then composer require --prefer-dist --no-update silverstripe/postgresql 2.0.x-dev; fi
-  - composer update
+  # Install composer dependencies
+  - composer validate
+  - composer require silverstripe/recipe-core 1.0.x-dev  --no-update
```
robbieaverill: I assume we need the CMS recipe for this module? Also relevant for the bootstrap path in phpunit.xml.dist

src/Models/BaseElement.php
```diff
@@ -569,8 +581,8 @@ public function forTemplate($holder = true)
         $config = SiteConfig::current_site_config();
 
         if ($config->Theme) {
-            Config::inst()->update('SSViewer', 'theme_enabled', true);
-            Config::inst()->update('SSViewer', 'theme', $config->Theme);
+            Config::inst()->update(SSViewer::class, 'theme_enabled', true);
+            Config::inst()->update(SSViewer::class, 'theme', $config->Theme);
```
robbieaverill: We should use `Config::modify()->set()` here as this API is deprecated.

Also the `theme` config property is deprecated, so we should instead push a theme to the `themes` array instead: `Config::modify()->merge(SSViewer::class, 'themes', [$config->Theme]);`

.travis.yml
```diff
@@ -1,31 +1,37 @@
 language: php
 
+dist: trusty
+
 env:
   global:
-    - COMPOSER_ROOT_VERSION=2.0.x-dev
+    - COMPOSER_ROOT_VERSION=4.0.x-dev
```
robbieaverill: My understanding is that this should match the module, rather than core

@tractorcow ?

CHANGELOG.md
robbieaverill: Discussed - we'll discuss this sort of thing another time

.travis.yml
tractorcow: Yes, this is simply the alias for "branch of the root project" which in this case is the module being checked out.

PR #110, opened 2017-10-09
CC-162/183 Modifications to elemental for v2

wilr:
- Strip out virtual blocks to separate module
- Update compatibility for SilverStripe 4


src/ElementalEditor.php
```diff
@@ -0,0 +1,120 @@
...
+     */
+    protected $title = '';
+
+    /**
+     * @var array
+     */
+    protected $types = [];
+
+    /**
+     * @param ElementalArea
```
robbieaverill:
```
     * @param string $name 
     * @param ElementalArea $area
```

src/ElementalEditor.php
```diff
@@ -0,0 +1,120 @@
...
+                    GridFieldPaginator::class,
+                    GridFieldPageCount::class,
+                    GridFieldAddExistingAutocompleter::class
+                ))
+                ->addComponent(new GridFieldOrderableRows('Sort'))
+                ->addComponent(new GridFieldDeleteAction(true))
+        );
+
+        if ($this->types) {
+            $adder = new ElementalGridFieldAddNewMultiClass('toolbar-header-left');
```
robbieaverill: Minor: would be great to use injector for this, in case anyone wants to extend it

```php
$adder = Injector::inst()->create(ElementalGridFieldAddNewMultiClass::class, 'toolbar-header-left');
```

src/Extensions/GridFieldDetailForm_ItemRequestExtension.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+namespace DNADesign\Elemental\Extensions;
+
+use SilverStripe\ORM\FieldType\DBField;
+use SilverStripe\Core\Extension;
+use DNADesign\Elemental\Models\BaseElement;
+
+class GridFieldDetailForm_ItemRequestExtension extends Extension
```
robbieaverill: While the core SS API is set now and is stuck with underscores, we don't need to follow suit - can we use `GridFieldDetailFormItemRequestExtension` or `GridFieldDetailForm\ItemRequestExtension` instead?

src/Models/BaseElement.php
```diff
@@ -13,80 +12,62 @@
...
-        'Parent' => ElementalArea::class,
-        'List' => ElementList::class // optional.
-    );
+    private static $has_one = [
+        'Parent' => ElementalArea::class
+    ];
 
     /**
-     * @var array $has_many
+     * @var array $extensions
```
robbieaverill: We don't actually need `$extensions` here as it's a description for the `@var` construct - it'd be better to remove it and the others in this class

src/Models/BaseElement.php
```diff
@@ -580,68 +430,66 @@ public function forTemplate($holder = true)
...
-
     /**
      * Renders the element in a custom template with the same name as the
      * current class. This should be the main point of output customization.
      *
      * Invoked from within ElementHolder.ss, which contains the "framing" around
      * the custom content, like a title.
      *
-     * @return string HTML
+     * @return HTMLText
```
robbieaverill: It also returns null

src/Models/BaseElement.php
```diff
@@ -688,106 +536,76 @@ public function getAnchor()
...
-
-        $usage->removeDuplicates();
-        return $usage;
     }
 
-    public function UsageSummary()
+    /**
+     * @param string $action
+     *
+     * @return string
```
robbieaverill: `|null`

src/Models/ElementalArea.php
```diff
@@ -98,105 +118,103 @@ public function renderSearchContent()
...
         return $controllers;
     }
 
     /**
-    * Return an ArrayList of pages with the Element Page Extension
-    *
-    * @return ArrayList
-    */
+     *
+     */
```
robbieaverill: This can just be removed

src/Models/ElementalArea.php
```diff
@@ -98,105 +118,103 @@ public function renderSearchContent()
...
     public function getOwnerPage()
     {
         if ($this->OwnerClassName) {
             $class = $this->OwnerClassName;
-            $elementalAreaRelations = ElementalAreasExtension::get_elemental_area_relations(singleton($class));
+            $elementalAreaRelations = Injector::inst()->get($class)->getElementalRelations();
```
robbieaverill: This will load as a singleton - is that what we want? If not, use `->create()`

src/Models/ElementalArea.php
```diff
@@ -98,105 +118,103 @@ public function renderSearchContent()
...
             $originalMode = Versioned::DRAFT;
         }
+
         Versioned::set_stage(Versioned::DRAFT);
-        $elementalPageTypes = self::elemental_page_types();
-        foreach ($elementalPageTypes as $elementalPageType) {
-            $elementalAreaRelations = ElementalAreasExtension::get_elemental_area_relations(singleton($elementalPageType));
+
+        foreach ($this->supportedPageTypes() as $class) {
+            $elementalAreaRelations = Injector::inst()->get($class)->getElementalRelations();
```
robbieaverill: Same here re: singleton

src/Models/ElementalArea.php
```diff
@@ -98,105 +118,103 @@ public function renderSearchContent()
...
+
             foreach ($elementalAreaRelations as $eaRelationship) {
                 $areaID = $eaRelationship . 'ID';
-                $page = $elementalPageType::get()->filter($areaID, $this->ID);
+                $page = $class::get()->filter($areaID, $this->ID);
```
robbieaverill: `$class` could be undefined at this point - can we tidy the logic up a bit so it's clearer that this wouldn't happen?

src/ElementalEditor.php
wilr: @robbieaverill GridFieldAddNewMultiClass doesn't implement `Injectable` infact, most of gridfield doesn't.

robbieaverill: Yeah, it doesn't need to to be created by the injector though. SS4 recommends this approach now - see https://docs.silverstripe.org/en/4/changelogs/4.0.0/#update-references-to-literal-classes-to-use-injector

src/Models/ElementalArea.php
wilr: Yep only need a singleton to get the hasOnes. Don't need to create anything.

src/Models/ElementalArea.php
wilr: Hmm, $class is in a if so pretty sure it can't be undefined but agree, this is a bit meh.

robbieaverill: I only skim read, but thought the condition where it's assigned is not the same as this section

PR #113, opened 2017-10-13
Tidying up GridField top level view

wilr:
screen shot 2017-10-13 at 3 06 06 pm

Based on Feedback from Paul.


src/Models/BaseElement.php
```diff
@@ -443,9 +443,9 @@ public function getRenderTemplates()
                 break;
             }
 
-            $templates[] = $value;
-            $templates[] = 'elements/' . DataObjectPreviewController::stripNamespacing($value);
-            $templates[] = DataObjectPreviewController::stripNamespacing($value);
+            $templates[] = $value . $suffix;
+            $templates[] = 'elements/' . DataObjectPreviewController::stripNamespacing($value) . $suffix;
```
robbieaverill: We don't need the elements folder convention any longer since everything is namespaced. We could possibly remove this functionality altogether?

wilr: 👍

templates/DNADesign/Elemental/Models/BaseElement_EditorPreview.ss
```diff
@@ -0,0 +1,15 @@
+<div class="elementalpreview">
+    <div class="elementalpreview--icon">$Icon</div>
+
+    <div class="elementalpreview--detail">
```
robbieaverill: I guess if we're following BEM that these dashes should be underscores

src/Models/BaseElement.php
```diff
@@ -443,9 +443,8 @@ public function getRenderTemplates()
                 break;
             }
 
-            $templates[] = $value;
-            $templates[] = 'elements/' . DataObjectPreviewController::stripNamespacing($value);
-            $templates[] = DataObjectPreviewController::stripNamespacing($value);
+            $templates[] = $value . $suffix;
+            $templates[] = DataObjectPreviewController::stripNamespacing($value) . $suffix;
```
robbieaverill: I left the same comment on #121 - with namespaced templates we actually _do_ want to keep the namespace rather than sanitising them, so maybe this logic can be removed now?

robbieaverill:
Can merge on green

sachajudd:
@wilr can we implement the updated designs in this P.R?

sachajudd:
Good to merge @robbieaverill

robbieaverill:
Thanks @sachajudd

PR #121, opened 2017-10-16
Enable preview in the CMS via the full page output

NightJar:
Viewing the content block without it's surrounding
context isn't ideal from a user's perspective.
So we'll instead render the page that the current
content block (element) belongs to, jumping to the
relevant section via an anchor.

### DIFF
--- a/src/Extensions/ElementalPageExtension.php
+++ b/src/Extensions/ElementalPageExtension.php
@@ -26,13 +26,13 @@ class ElementalPageExtension extends ElementalAreasExtension
     {
         $controller = Controller::curr();
         $request = $controller->getRequest();
-        if($request->getVar('ElementalPreview') !== null) {
+        if ($request->getVar('ElementalPreview') !== null) {
             $html = HTML4Value::create($tags);
             $xpath = "//meta[@name='x-page-id' or @name='x-cms-edit-link']";
             $removeTags = $html->query($xpath);
             $body = $html->getBody();
-            foreach($removeTags as $tag) {
-              $body->removeChild($tag);
+            foreach ($removeTags as $tag) {
+                $body->removeChild($tag);
             }
             $tags = $html->getContent();
         }
--- a/src/Extensions/ElementalPageExtension.php
+++ b/src/Extensions/ElementalPageExtension.php
@@ -36,6 +36,5 @@ class ElementalPageExtension extends ElementalAreasExtension
             }
             $tags = $html->getContent();
         }
-        return $tags;
     }
 }
--- a/src/Extensions/ElementalPageExtension.php
+++ b/src/Extensions/ElementalPageExtension.php
@@ -3,6 +3,8 @@
 namespace DNADesign\Elemental\Extensions;
 
 use DNADesign\Elemental\Models\ElementalArea;
+use SilverStripe\View\Parsers\HTML4Value;
+use SilverStripe\Control\Controller;
 
 class ElementalPageExtension extends ElementalAreasExtension
 {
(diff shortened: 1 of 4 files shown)


src/Extensions/ElementalPageExtension.php
```diff
@@ -19,4 +21,21 @@ class ElementalPageExtension extends ElementalAreasExtension
...
+            $html = HTML4Value::create($tags);
+            $xpath = "//meta[@name='x-page-id' or @name='x-cms-edit-link']";
+            $removeTags = $html->query($xpath);
+            $body = $html->getBody();
+            foreach ($removeTags as $tag) {
+                $body->removeChild($tag);
+            }
+            $tags = $html->getContent();
+        }
+        return $tags;
```
robbieaverill: we probably don't need to return here since it's modified by reference

src/Extensions/GridFieldDetailFormItemRequestExtension.php
```diff
@@ -22,4 +26,27 @@ public function updateBreadcrumbs($crumbs)
...
+            $navigator = SilverStripeNavigator::create($this->owner->record);
+            $field = LiteralField::create(
+                'SilverStripeNavigator',
+                $navigator->renderWith($template)
+            )->setAllowHTML(true);
+            $fields->push($field);
+            $form->addExtraClass('cms-previewable')
+                ->removeExtraClass('cms-panel-padded center');
+            Requirements::javascript(
+                'DNADesign/Elemental:javascript/Elemental.Preview.js'
```
robbieaverill: the vendor/module syntax will need to match composer.json - so it's probably `dnadesign/silverstripe-elemental:...`

src/Extensions/GridFieldDetailFormItemRequestExtension.php
```diff
@@ -22,4 +26,27 @@ public function updateBreadcrumbs($crumbs)
...
+            );
+        }
+        return $form;
```
robbieaverill: Same as earlier, we don't need to return here

src/Models/BaseElement.php
```diff
@@ -443,15 +442,29 @@ public function getRenderTemplates($suffix = '')
                 break;
             }
 
+            $namespacelessClassname = $this->stripNamespacing($value);
+
             $templates[] = $value . $suffix;
-            $templates[] = 'elements/' . DataObjectPreviewController::stripNamespacing($value) . $suffix;
-            $templates[] = DataObjectPreviewController::stripNamespacing($value) . $suffix;
+            $templates[] = 'elements/' . $namespacelessClassname . $suffix;
```
robbieaverill: @wilr has removed this line in #113, whichever doesn't get merged first will need to rebase anyway, but we can remove this line because templates are namespaced now anyway. In fact - this logic isn't really applicable any more, since we _do_ want namespaced classes when looking up template names

src/Models/BaseElement.php
```diff
@@ -443,15 +442,29 @@ public function getRenderTemplates($suffix = '')
...
         return $templates;
     }
 
     /**
+     * Strip all namespaces from class namespace
+     * @param string e.g. "\Fully\Namespaced\Class"
+     *
+     * @return string following the param example, "Class"
+     */
+    private function stripNamespacing($classname)
```
robbieaverill: We prefer protected over private

src/Models/BaseElement.php
```diff
@@ -443,15 +442,29 @@ public function getRenderTemplates($suffix = '')
...
+            $templates[] = 'elements/' . $namespacelessClassname . $suffix;
+            $templates[] = $namespacelessClassname . $suffix;
         }
 
         return $templates;
     }
 
     /**
+     * Strip all namespaces from class namespace
+     * @param string e.g. "\Fully\Namespaced\Class"
```
robbieaverill: `@param string $classname`

src/Extensions/ElementalPageExtension.php
NightJar: It's a tidy habit of mine as a while back it was introduced that you could do either (return or reference). But without the reference declaration here it no longer works... so yeah; it is pointless.

src/Extensions/GridFieldDetailFormItemRequestExtension.php
NightJar: Hmm, dead code actually, shifted from a CMS JS approach to the `MetaTags` extension in `ElementalPageExtension`. So I'll just remove it.

PR #123, opened 2017-10-19
API: Add history tab. Migrate content style variants to BaseElement

wilr:
History tab changes and report.

Blockers:

- [x] https://github.com/silverstripe/silverstripe-versioned/pull/60
- [x] https://github.com/silverstripe/silverstripe-framework/pull/7500
- [x] Peer review 
- [x] Tests for revert button functionality and report view.

Report blockers:

- [x] ~https://github.com/silverstripe/silverstripe-reports/pull/88~
- [x] ~UX review for report view.~

PR #135, opened 2017-10-30
NEW Add content block type and use reports

robbieaverill:
Adds "Block type report" and "Blocks in use report".

## Blockers

- [x] #123


NightJar:
Merge on green 👍 
Travis is having issues.

PR #148, opened 2017-11-09
Enable the new MeatballMenu gridfield-extension

NightJar:
As per design updates for the SilverStripe 4 release
each row now has actions on it that allow for things
such as publication to be enacted without the need to
browse deeper into the GridField item view; UX++

---
Depends on:
- [ ] https://github.com/symbiote/silverstripe-gridfieldextensions/pull/225

### DIFF
--- a/src/Extensions/ElementalPageExtension.php
+++ b/src/Extensions/ElementalPageExtension.php
@@ -5,7 +5,6 @@ namespace DNADesign\Elemental\Extensions;
 use DNADesign\Elemental\Models\ElementalArea;
 use SilverStripe\View\Parsers\HTML4Value;
 use SilverStripe\Control\Controller;
-use SilverStripe\Forms\FieldList;
 
 class ElementalPageExtension extends ElementalAreasExtension
 {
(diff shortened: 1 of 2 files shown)


src/Extensions/ElementalPageExtension.php
```diff
@@ -37,4 +38,25 @@ public function MetaTags(&$tags)
...
+            // We need MeatballMenu to handle the same routes. Detail form is required by
+            // \Symbiote\GridFieldExtensions\GridFieldAddNewMultiClass so can't just be removed,
+            // since it AddNewMultiClass is also utilised by Elemental.
+            '\SilverStripe\Forms\GridField\GridFieldDetailForm'
+        ]);
+        $cfg->addComponents(
+            new \Symbiote\GridFieldExtensions\GridFieldMeatballMenuComponent,
+            new \SilverStripe\Forms\GridField\GridFieldDetailForm
+        );
+        return $fields;
```
robbieaverill: Some initial comments:

* We should add these components to `ElementalEditor` where the GridField is first initialised
* Can we please refer to class names using `ClassName::class` and namespace imports rather than FCQN strings?
* Ideally use `Injector::inst()->create()` rather than creating new classes directly

NightJar:
As per the OP, this is dependent on another module, which is blocked for reasons noted in https://github.com/symbiote/silverstripe-gridfieldextensions/pull/225#issuecomment-346162598

This will be picked up and completed at a later time, likely when work on inline editing commences, as it is a more critical component of that workflow.

PR #149, opened 2017-11-09
"Discarded changes" message is no longer displayed onCreate()

mfendeksilverstripe:
Discarded changes popup message no longer shows up when creating a new element.

Related to:

https://github.com/dnadesign/silverstripe-elemental/issues/142


robbieaverill:
Nice! My initial thought is that this bug could/should be fixed in gridfieldextensions, perhaps across the board for all class types. This works nicely for now in this module anyway. What do you think?Thanks @mfendeksilverstripe.

mfendeksilverstripe:
@robbieaverill I agree. This change could be applied to `gridfieldextensions ` as well.

robbieaverill:
Logged as https://github.com/symbiote/silverstripe-gridfieldextensions/issues/226 - will need to be confirmed first

PR #154, opened 2017-11-13
Adds ability to have templates for specific element areas

### LINKED ISSUE #85: Ability to define element templates by area
This feature is handled on Content Blocks module by template names like: SlideshowBlock_{AreaName}.ss which seems like a reasonable implementation.

mak001:
Resolves #85

### DIFF
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -430,6 +430,7 @@ class BaseElement extends DataObject implements CMSPreviewable
                 break;
             }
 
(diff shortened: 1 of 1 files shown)


korthjp17:
@mak001 when creating a template for an element that's in a different element area, it seems both element areas are using the same template.

mak001:
@korthjp17 This is an underlying issue with elemental's `BaseElement->getAreaRelationName()`. #161 fixes it .

muskie9:
bump

PR #160, opened 2017-12-04
Fix for getOwnerPage always returning zero results.

chrispenny:
Anyone please correct me if I'm wrong, but no matter how many situations I tried, I couldn't get `$class::get()->filter($areaID, $this->ID);` to consistently return `DataObject` results when they did indeed exist.

Using `Versioned::get_by_stage();` looks to always be consistent, and has the added benefit of not requiring you to `set_stage()`.


ScopeyNZ:
I'm reverting this code with #653 . Assuming the stage and using `Versioned::get_by_stage()` is very dangerous. In my case it breaks archive mode on a "large project" that I'm working on. See the PR for a tiny bit more detail.

PR #161, opened 2017-12-04
Fixes getAreaRelationName() always returning the same relation

mak001:
if there are multiple element areas


robbieaverill:
@mak001 would you mind providing a bit more context around the error that this pull request fixes?

mak001:
Having multiple ElementAreas on a page will cause only one element area relation to be returned currently.
I created a test element which displays its area relation name and added one to each area on the page.
screenshot from 2018-01-12 15-03-01
without the fix, both element area relations return the first relation they come across. (which happens to be the Sidebar area here)
screenshot from 2018-01-12 15-03-20
With the fix it will get the correct relation name. (this is what is expected)
screenshot from 2018-01-12 15-07-16

muskie9:
bump

robbieaverill:
Logic looks sound. Will need a rebase @mak001

korthjp17:
Is this good to be merged? We are working on a project that will launch soon that is using this. It'd be helpful to have it in the main repo rather than referencing @mak001's PR branch in the composer requirements.

robbieaverill:
Merged :)

korthjp17:
Awesome, thank you!!

PR #164, opened 2017-12-13
fix(ElementList): Provide a better error message

silbinarywolf:
**WARNING: This is against branch `1`**

fix(ElementList): Provide a better error message when a class doesn't exist that's defined against an ElementList or ElementList subclass.

### DIFF
--- a/code/models/ElementList.php
+++ b/code/models/ElementList.php
@@ -105,6 +105,9 @@ class ElementList extends BaseElement
                 $sorted = array();
 
                 foreach ($list as $class) {
+                    if (!class_exists($class)) {
+                        throw new Exception($class.' is defined in '.get_class($this).'::allowed_elements config but doesn\'t exist.');
+                    }
                     $inst = singleton($class);
 
                     if ($inst->canCreate()) {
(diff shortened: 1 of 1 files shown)


robbieaverill:
Seems reasonable. It might be worth adding a unit test for it?

robbieaverill:
I think throwing an exception that didn't used to be there would be a breaking change, and we've already moved to a new major version of this module for SS4 support. To avoid this you could add a config static like `check_valid_element_types` which a dev could enable to then add this functionality. What do you think?

silbinarywolf:
If it's still convention to do the following for bugs that shouldn't halt execution, perhaps the following?
`user_error($messageHere, E_USER_WARNING)`

robbieaverill:
Yeah that might be OK. What do you think @pitchandtone?

pitchandtone:
Sounds fine to me, there's enough precedence to use user_error elsewhere.

silbinarywolf:
@robbieaverill @pitchandtone This might want to go into `master` as well if applicable?

robbieaverill:
Yep, we'll merge it up after merge. I think it would be worth converting it back to an Exception when it hits master - do you agree?

silbinarywolf:
@robbieaverill I'm really not fussed at whether it halts execution or not so... your call really.

robbieaverill:
Oh, haha - this class is moved to https://github.com/dnadesign/silverstripe-elemental-list for SS4, so master won't change

PR #168, opened 2018-01-08
Fix edit CMSEditLink when parent is not a page

cjsewell:
In BaseElement.php, CMSEditLink assumes the parent is always a page.
This generates an incorrect CMSEditLink for nested Elements ( e.g  [silverstripe-elemental-list ](https://github.com/dnadesign/silverstripe-elemental-list) )

This pull requests attempts to resolve the correct CMSEditLink when the Elements parent is not a page type.

### DIFF
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -8,6 +8,7 @@ use DNADesign\Elemental\Forms\HistoricalVersionedGridFieldItemRequest;
 use DNADesign\Elemental\Forms\TextCheckboxGroupField;
 use DNADesign\Elemental\Controllers\ElementController;
 use SilverStripe\CMS\Controllers\CMSPageEditController;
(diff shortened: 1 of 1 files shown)


robbieaverill:
Thanks for contributing @cjsewell! Would you be happy to add a unit test for this behaviour? You could add some test stubs (that implement the TestOnly interface) to set up your example relationship, then assert some results from a method that returns the edit link for your object.

nglasl:
+1 for this. I encountered this one recently, and found that the simplest solution was to just remove this anchor from the template https://github.com/dnadesign/silverstripe-elemental/blob/master/templates/DNADesign/Elemental/Models/BaseElement_EditorPreview.ss#L2. It already resolves correctly without it, so is there a specific reason that it exists to begin with?

robbieaverill:
I'm not sure. We could try removing it!

Issue #125, opened 2017-10-20
Missing .upgrade.yml


silbinarywolf:
When upgrading an existing project from 3.X to 4.X, this file is meant to help map old class names to new class names.

Looks like this doesn't exist currently for the core module and the additional modules here:
https://github.com/dnadesign/silverstripe-elemental-list
https://github.com/dnadesign/silverstripe-elemental-userforms

robbieaverill:
Thanks. We can add one to each repo, will need to split it up between them

robbieaverill:
Thanks @raissanorth!

PR #169, opened 2018-01-11
BUG Add upgrade.yml file and ignore in export

raissanorth:
This commit adds the upgrade.yml file as requested in #125


robbieaverill:
>adds the upgrade.yml to the gitattributes file to be ignored in exports

It doesn't, but that's good because we don't want it to

raissanorth:
Silly me!

PR #180, opened 2018-01-26
UPDATE Modify ElementHolder location/reference to enable theme templating

adover:
Currently you cannot modify the ElementHolder template in the theme due to the controller_template being set without a namespace. This enables that to happen. ElementHolder can then be customised in the <templateDir>/DNADesign/Elemental/Layout/ElementHolder.ss


robbieaverill:
Thanks for contributing! In future I'd suggest making your pull request from a branch other than `master` in your fork, in case we ask you to change the pull request target branch, which would end up with you messing up your master branch locally

PR #182, opened 2018-02-02
Pulls/2/elemental area publish permissions

### LINKED ISSUE #156: Pages with an ElementalArea can't be published by non-admins
You don't see an issue when logged in as admin, but as non-admins, one isn't able to publish a page when Elements are added to it.

Theory:

Because the Page is set to $owns the ElementalArea by the provided extension, the Page will try to publish the ElementalArea. The ElementalArea by default has no canPublish method / canEdit methods and publishing such pages as a non-admin will throw the error that "the user can't publish the ChangeSet" (along those lines).

I fixed it with this:

```
<?php

use SilverStripe\Security\Permission;

class ElementalAreaPublishExtension extends DataExtension {

  public function canPublish($member = NULL) {
    return Permission::checkMember($member, "CMS_ACCESS");
  }

}
```

This is for SS 4.

andrewandante:
Simple check of the parent page to determine whether the area can be edited (and thus published).

Considerations:

 - do we want a granular set of permissions for Elements?
 - do we want/need an equivalent `canView()` etc?

Fixes #156

### DIFF
--- a/src/Models/ElementalArea.php
+++ b/src/Models/ElementalArea.php
@@ -13,6 +13,7 @@ use SilverStripe\ORM\ArrayList;
 use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\HasManyList;
 use SilverStripe\ORM\UnsavedRelationList;
+use SilverStripe\Security\Permission;
 use SilverStripe\Versioned\Versioned;
 use SilverStripe\View\Requirements;
 use SilverStripe\View\SSViewer;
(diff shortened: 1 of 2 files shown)


robbieaverill:
I think we should assume that the content blocks belong to a page, therefore the page permissions should be sufficient. This PR looks good - is there anything left?

andrewandante:
Only thing maybe left is to add a `canView()` as well - otherwise it's good to go

NightJar:
I think it would be sensible to add. Although there isn't a controller endpoint/route for viewing a single element, I think it's probably good to tidy up the loose end & make the consistency & relation to a `Page` and it's permissions exceedingly obvious.

andrewandante:
@NightJar I tend to agree. Supplemented and re-pushed

Issue #165, opened 2017-12-14
Add indication of the state of a content blocks (draft/modified)


clarkepaul:
The state of a content block should be communicated through the dot above the block icon (similar to the Files area and site tree icon).
https://invis.io/VMDZ9YAGX#/265320689_Blocks_-_Future_-_Carousel_NTH

There should be text tooltip with the block type and state in text format, which will also be useful for accessibility reasons. Note that the published blocks do not communicate the state through text or visually.
Tooltip examples:
[File block: Modified] [Banner block: Draft] [Content block]

flamerohr:
Check out `VersionedGridFieldState` for a re-useable extension to help

jelicanin:
We made some progress on this issue.

Default view:
![default-view](https://user-images.githubusercontent.com/692844/35877872-daf7102a-0b76-11e8-816e-c4827b99937c.png)

Mouse hover on Published item:
![published-item-hover](https://user-images.githubusercontent.com/692844/35877871-dadb1668-0b76-11e8-9336-1de7055f49c5.png)

Mouse hover on Modified item:
![modified-item-hover](https://user-images.githubusercontent.com/692844/35877873-db199aaa-0b76-11e8-9e16-8e1f162c08a2.png)


Check out details [in our fork](https://github.com/fractaslabs/silverstripe-elemental).
I can make PR if you like it.

robbieaverill:
@jelicanin your changes look good, but it's essentially duplicating what GridFieldVersionedState does when you add it to the GridField.

I've been able to add the status badges reasonably quickly using this component:

![image](https://user-images.githubusercontent.com/5170590/35899836-98287a5a-0c33-11e8-9dd3-f79d756a78ff.png)


We'll need to do something sneaky to the status badge to get it to show up an orange dot on the block icon when it's in a draft/modified state though, since that's not a default behaviour of this component:

* Apply the component to one of the fields in the editor grid
* Give the component a `sr-only` class so that the label is present for screen readers but it doesn't modify the UI (the component doesn't provide any extension hooks, so we may need to overload the component in this module)
* Ideally, add a modifier class to the `elemental-preview__icon` element, e.g. `elemental-preview__icon--modified`
* Define some CSS to add the dot for that class (note there's an existing example in the CMS under the `.cms-tree.jstree .status-modified>a .jstree-pageicon:before` selector, visible in the main site tree)

[...]

PR #187, opened 2018-02-12
FIX Remove the aliased versioned GridFieldState component

robbieaverill:
Between SS 4.0 to 4.1 this component was copied to the versioned module and the original was deprecated. This means that the new component shows up in all GridFields by default again, so this is a patch to remove both the old and new component names.

This change leaves both components in place because we need to support SS 4.0 and 4.1+, so can't just use the new namespace.

Before:
image

This is only until #165 is completed, since that will actually reintroduce this component in one way or another. For now it's displayed incorrectly so removing it as per the current code logic.

cc @sachajudd


sachajudd:
Sounds good 👍

clarkepaul:
Do we have a followup issue to position the state as a circle @robbieaverill ?

robbieaverill:
@clarkepaul yep we're still going to do it on #165

PR #188, opened 2018-02-13
NEW Enable elemental content to return its owner page in CMS SiteTree searches

### LINKED ISSUE #163: Ensure that elements are searchable via the default CMS page search form
When searching in the site tree using the standard CMS search form, we need to ensure that elements under a page can have their content searched and returned as per normal under the page that owns them.

This would behave the same way as if the page wasn't using elemental at all.

We can possibly re-use the frontend search content PHP function `getElementsForSearch` (not sure what it's called off hand) for this.

robbieaverill:
Please note that this logic will not be performant in larger data sets, and is likely to change in future in favour of a more rigid and specific set of ORM joins.

I've marked the method which will definitely be removed once this enhancement is made as `@internal` with a warning not to use it, and have made it private just to make sure.

Better way to do this will be logged as a 2.1.x enhancement separately.

Resolves #163

### DIFF
--- a/src/Controllers/ElementSiteTreeFilterSearch.php
+++ b/src/Controllers/ElementSiteTreeFilterSearch.php
@@ -1,0 +1,112 @@
+<?php
+
+namespace DNADesign\Elemental\Controllers;
+
+use DNADesign\Elemental\Extensions\ElementalPageExtension;
+use SilverStripe\CMS\Controllers\CMSSiteTreeFilter_Search;
+use SilverStripe\CMS\Model\SiteTree;
+use SilverStripe\Forms\DateField;
+use SilverStripe\ORM\ArrayList;
+use SilverStripe\ORM\DataList;
+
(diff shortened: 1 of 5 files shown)


robbieaverill:
Follow up to make this better logged at https://github.com/dnadesign/silverstripe-elemental/issues/189

src/Controllers/ElementSiteTreeFilterSearch.php
```diff
@@ -0,0 +1,112 @@
...
+            }
+
+            switch ($name) {
+                case 'Term':
+                    $query = $query->filterAny([
+                        'URLSegment:PartialMatch' => $val,
+                        'Title:PartialMatch' => $val,
+                        'MenuTitle:PartialMatch' => $val,
+                        'Content:PartialMatch' => $val
+                        ] + $this->extraTermFilters); // NB: only modified line
```
NightJar: Thanks :)

PR #192, opened 2018-02-24
Fr translation

DrMartinGonzo:


robbieaverill:
Thanks for the pull request! We manage our non-English localisations through [Transifex](https://www.transifex.com/silverstripe/silverstripe-elemental/language/fr/) - would you mind submitting them there instead? The non-English files in these projects are automatically updated so changes like this would be undone during the next release process.

bummzack:
@robbieaverill couldn't you push from the `fr.yml` to transifex? Eg. something like:

    tx push -t -l fr

robbieaverill:
True, we can do that!

robbieaverill:
Pushed

PR #194, opened 2018-02-27
VersionedGridFieldState class does not exist

DrMartinGonzo:


robbieaverill:
Thanks for the PR!

The component was introduced into the framework in 4.0, then moved to Versioned in 4.1 and deprecated.

I'd be happy for you to change the translations to use the 4.0 keys, but we can't remove the 4.1 class name. In fact, it's likely that we'll introduce a 4.1 minimum dependency into this module soon.

I'm going to close this for this now - feel free to reopen if you'd like to, otherwise I hope that the existing code won't cause any more issues than IDE notices and missing translations.

DrMartinGonzo:
No real issues, thanks.
Sorry I missed the change in 4.1 !

PR #196, opened 2018-03-02
Set getType to use singular name by default

gorriecoe:
Set getType to use singular name by default so its consistent with silverstripe.
Removed translations so that doesn't override singular name unless explicitly set by extended element subclass.

### DIFF
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -105,9 +105,9 @@ class BaseElement extends DataObject implements CMSPreviewable
 
     private static $default_sort = 'Sort';
 
-    private static $singular_name = 'block';
+    private static $singular_name = 'Block';
 
-    private static $plural_name = 'blocks';
+    private static $plural_name = 'Blocks';
 
     private static $summary_fields = [
         'EditorPreview' => 'Summary'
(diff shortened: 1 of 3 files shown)


src/Models/BaseElement.php
```diff
@@ -376,7 +376,7 @@ public function getHistoryFields($checkLatestVersion = true)
      */
     public function getType()
     {
-        return _t(__CLASS__ . '.BlockType', 'Block');
+        return _t(__CLASS__ . '.BlockType', $this->i18n_singular_name());
```
robbieaverill: This doesn't work with the text collector unfortunately, since it's runs statically rather than executing dynamic logic. You can prove this by adding `function i18n_singular_name() { return 'foo'; }` and run `vendor/bin/sake dev/tasks/i18nTextCollectorTask '' flush=1 module=dnadesign/silverstripe-elemental` and observe that "foo" doesn't get set as the default value in `lang/en.yml`.

The reason why we have this method is because if you have a "Carousel Element" for example, the i18n singular name could be "Carousel Element" but we don't want to show "Carousel Element" in the dropdown for adding new blocks, or when showing the block type in GridFields. Instead we just want "Carousel". Referring to the element as this in other parts of the CMS is where we'd use the i18n singular name instead, but in certain contexts it doesn't make sense to use a full descriptive name like this (this is noted in the doc block of getType).

I realise it's an extra API to maintain for extensions of BaseElement, but for the sake of the UX we want to achieve we decided it was necessary (and not too much overhead) to implement.

Hope this makes sense! Thanks =)

PR #197, opened 2018-03-02
Added improvements to rendertemplates

gorriecoe:
Added extend updateRenderTemplates.

Added style based template suffix.

### DIFF
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -461,6 +461,11 @@ class BaseElement extends DataObject implements CMSPreviewable
                 break;
             }
 
+            $this->extend('updateRenderTemplates', $templates, $value, $suffix);
+            if ($style = $this->Style) {
+                $templates[] = $value . $suffix . '_'. $this->getAreaRelationName() . '_' . $style;
+                $templates[] = $value . $suffix . '_' . $style;
+            }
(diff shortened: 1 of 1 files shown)


gorriecoe:
I have taken a different approach to this now.  By splitting the templates into there classes we modify them easier.

robbieaverill:
How do you feel about refactoring this a little to reduce some of the duplication?

derralf:
will this (or at least the extension hook which would be very usefull) be merged/released anytime soon?

robbieaverill:
@derralf we won't be releasing a new major version of this module in the near future, so best to use the `3.0.x-dev` constraint. We could also cherry-pick this back into a 2.x release (minor) but this wouldn't likely happen for a few weeks or months.

derralf:
@robbieaverill maybe a dumb problem/question, but how do I install this with SS4.1?
`3.0.x-dev` requires `silverstripe/versioned-admin 1.x-@dev` which requires `silverstripe/admin 1.2.x-dev`
but `silverstripe/recipe-cms 1.1.0 ` (SS4.1) requires `requires silverstripe/admin 1.1.0@stable` so composer won't install/update to `3.0.x-dev`

robbieaverill:
3.0.x-dev looks like it requires SS 4.2 or higher

PR #215, opened 2018-04-06
Add PHP 7.2 and various SilverStripe recipe versions to Travis matrix

robbieaverill:

PR #217, opened 2018-04-06
API Bump CMS version, implement recursive duplication, 

robbieaverill:
Requires #214 and #216 to be merged first, then probably have this PR rebased.

---

In SS 4.1 we can use `$cascade_duplicates` to ensure that related objects to a page are duplicated when a page is duplicated.

This PR adds some tests to prove this fails in SS 4.0 without using the API, and that it works with SS 4.1.

It also:

* bumps the minimum CMS version to 4.1
* deprecates the ElementalDuplicationExtension, which as far as I can tell is not used by default
* fixes the wrong Config namespace in that extension


.travis.yml
```diff
@@ -24,7 +24,7 @@ before_script:
 
   # Install composer dependencies
   - composer validate
-  - composer require silverstripe/recipe-cms 1.0.x-dev  --no-update
+  - composer require silverstripe/recipe-cms 1.1.x-dev  --no-update
```
NightJar: Is this something we should introduce a `$RECIPE_VERSION` var to?
This is already stable so support will need to continue for the lower versions too

composer.json
```diff
@@ -16,7 +16,7 @@
     "minimum-stability": "dev",
     "prefer-stable": true,
     "require": {
-        "silverstripe/cms": "^4.0@dev",
+        "silverstripe/cms": "^4.1",
```
NightJar: 👍

NightJar:
I approve the changes, but you'll need to address the error `phpcs` is giving (apparently missing standard definition from framework - has it been moved in 4.1?).
Then merge on green though :)

.travis.yml
robbieaverill: See #215 (noted requirement)

robbieaverill:
See #216 (noted requirement)

robbieaverill:
💚

Issue #181, opened 2018-01-31
Readonly view


chillu:
# Overview

Readonly block views become important for viewing history of a page with all of its blocks. Currently that's not working. The SS4 page history only shows a "none" label, even with two (published) content blocks associated to that version.

![image](https://user-images.githubusercontent.com/111025/35604528-7fe862b6-06a7-11e8-8717-6fc63fcbd261.png)

# Acceptance Criteria

 * I can see all the information of the editable block overview, but as readonly fields
 * The position is displayed as an explicit readonly field
 * The view works with all core form fields in SilverStripe (excl. GridField, see https://github.com/silverstripe/silverstripe-framework/issues/3357)

# Notes

 * Designs: https://projects.invisionapp.com/share/5NEHOBLYZ#/screens/258945845_Blocks1_-_Page_History
 * Only needs to work in Entwine-based views (which can include React fields like UploadField or TreeDropdownField)
 * If there's significant work in getting form field readonly views going, split out into other cards
 * Excludes adding the link for "block history", see [listing versions of a block](https://github.com/silverstripe/silverstripe-versioned/issues/37)
* assuming 1 level deep

robbieaverill:
It’s worth noting that block history does work, just not page history at this point

robbieaverill:
I just tested the history for blocks on SS 4.1 and there have been some regressions. Will isolate and log an issue shortly.

NightJar:
Just adding tests for the PR at https://github.com/dnadesign/silverstripe-elemental/pull/224

robbieaverill:
Done in https://github.com/silverstripe/silverstripe-admin/pull/511 and #224

PR #230, opened 2018-05-14
NEW Revert a version (Block History)

raissanorth:
The changes in this PR enable reversion to an earlier block version. 

Note, that there are some styling issues caused by changes made in https://github.com/silverstripe/silverstripe-versioned-admin/pull/13

These include the form not expanding to full height and it currently not being scrollable. It is suggested that these are addressed in `versioned-admin`.


robbieaverill:
Can you please change the target branch to master and rebase?

client/src/boot/registerComponents.js
```diff
@@ -0,0 +1,20 @@
...
+      updater.component(
+        'HistoryViewer.Form_ItemEditForm',
+        readOneBlockQuery,
+        'ElementHistoryViewer'
+      );
+      }
+  );
+
+
+};
```
robbieaverill: This file can be removed, since it was renamed in #181 to `registerTransforms.js`

client/src/components/ElementHistoryViewer/ElementHistoryViewer.js
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * Transformer function which binds the readOneBlockQuery GraphQL query to the HistoryViewer
+ * component.
+ */
+const withElementsHistoryViewer = (HistoryViewer) => compose(
+  readOneBlockQuery
+)(HistoryViewer);
+
+export default withElementsHistoryViewer;
```
robbieaverill: We've removed this file deliberately, since it's handled purely in the registerTransforms file now. Is there a reason it was added back in?

client/src/state/history/revertToBlockVersionMutation.js
```diff
@@ -0,0 +1,43 @@
...
+    return {
+      actions: {
+        ...actions,
+        revertToVersion,
+      },
+    };
+  },
+  options: {
+    // Refetch versions after mutation is completed
+    refetchQueries: ['ReadHistoryViewerBlock']
```
robbieaverill: You've changed this query name to `readOneBlock`, will need to update it here too

raissanorth: I followed the same logic used in silverstripe/cms. 

The actual query has not been renamed:
```
query ReadHistoryViewerPage ($page_id: ID!, $limit: Int!, $offset: Int!) {
  readOnePage(
  ...
```

robbieaverill: Yep looks fine now after rebasing, it must have been an old change

PR #252, opened 2018-07-04
NEW BaseElement implements CMSPreviewable interface

raissanorth:
While `CMSPreviewable`'s functions were implemented in prior commits, the interface itself was not implemented. 

https://github.com/silverstripe/silverstripe-admin/issues/480 depends on BaseElement doing so.

### DIFF
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -17,6 +17,7 @@ use SilverStripe\Forms\FieldList;
 use SilverStripe\Forms\HiddenField;
 use SilverStripe\Forms\NumericField;
 use SilverStripe\Forms\TextField;
+use SilverStripe\ORM\CMSPreviewable;
 use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\FieldType\DBField;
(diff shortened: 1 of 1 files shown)


NightJar:
This is largely my fault; setting this up provided a great opportunity to test the preview toggle widget outside of the context of a page (as a whole). To be honest I completely don't remember removing this functionality - but it makes sense.

To that end if it's agreed that we don't need this functionality, it would be best to also remove all supporting methods before releasing 3.0.0 stable
e.g. `PreviewLink` and `isCMSPreview
https://github.com/dnadesign/silverstripe-elemental/blob/2a604077419b29fe62e8b3968f1559f4b7ebcd66/src/Models/BaseElement.php#L536-L565

raissanorth:
Thanks for your feedback @robbieaverill and @NightJar. I must admit that it does not make too much sense previewing a content block if that functionality is already available as part of the blocks page preview - it would only add value if the blocks element were displayed individually and we cannot assume that this will always be the case, right?! Therefore closing this PR.

PR #254, opened 2018-07-12
Replace some unit tests with behat tests and make required_extensions explicit in tests

robbieaverill:
This is to avoid encountering https://github.com/silverstripe/silverstripe-framework/issues/7978 when running in context with other modules like silverstripe-elemental-userforms that define their own test stubs.


tests/Reports/ElementTypeReportTest.php
```diff
@@ -1,55 +0,0 @@
...
-
-    public function testReportShowsBlockTypes()
-    {
-        $this->logInWithPermission('ADMIN');
-
-        $result = (string) $this->get('admin/reports/show/DNADesign-Elemental-Reports-ElementTypeReport')->getBody();
-
-        $this->assertContains('Content Type', $result, 'Table has headings');
-
-        $this->assertContains('Unused Element', $result, 'Unused elements are still shown in the report');
```
NightJar: I think covering this in your new behat replacement is important. It doesn't appear to be there, or have I missed something?

robbieaverill: Yes you're right in that this isn't in the new behat test. The stub it's checking is for phpunit, we'd need to make some stub elements available to the behat tests to make this work. Let's raise a new issue for it

NightJar: Ideal 👍

PR #259, opened 2018-07-13
FIX Check canView() permission before rendering BaseElements

### LINKED ISSUE #253: canView permissions aren’t checked on front-end
I’ve implemented `canView()` permissions on a per-element basis (mimicking `CanViewType` and `ViewerGroups` from `SiteTree`), out of the box these permissions aren’t checked before rendering.

It’s easy enough to implement these checks myself, just struck me as worth considering :)

raissanorth:
Fixes #253 by checking canView() permissions before rendering BaseElements.

### DIFF
--- a/tests/ElementalAreaTest.php
+++ b/tests/ElementalAreaTest.php
@@ -33,16 +33,36 @@ class ElementalAreaTest extends SapphireTest
         $this->assertEquals(2, $controllers->count(), 'Should be a controller per element');
     }
 
-
-    public function testViewPermissionsAreChecked()
+    public function testCanViewTestElementIsFalseWhenLoggedInAsCmsEditor()
     {
+        /** @var ElementalArea $area */
         $area = $this->objFromFixture(ElementalArea::class, 'area2');
-        $controllers = $area->ElementControllers();
-        $elements = $area->Elements();
+        // Content editors do not have permission to view the TestElement
+        $this->logInWithPermission('VIEW_DRAFT_CONTENT');
 
-        $this->assertEquals(1, $controllers->count(),
-            'Should be one controller only, since one of the elements is not viewable');
-        $this->assertEquals(2, $elements->count());
+        $controllers = $area->ElementControllers();
+        $this->assertCount(2, $area->Elements(), 'There are two elements in total');
+        $this->assertCount(
+            1,
+            $controllers,
+            'Should be one controller only, since TestElement is not viewable by non-admins'
+        );
+    }
+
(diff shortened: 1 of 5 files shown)


src/Controllers/ElementController.php
```diff
@@ -68,10 +68,14 @@ public function getElement()
...
      * {@link ElementController}.
      *
-     * @return string HTML
+     * @return null|string HTML
      */
     public function forTemplate()
     {
+        if (!$this->element->canView()) {
+            return null;
+        }
```
robbieaverill: I think I'd rather put `<% if $canView %>` checks into the template, e.g. around ElementHolder.ss

If continuing with this approach we'd need to update the unit test that's failing

kinglozzer: We’ve had issues in core before, I think with `Versioned::doPublish()`, where people bypass permission checks intentionally (e.g. CLI publishes and stuff like that) and expect things to work. I think this could fall into that category - if I call `$html = $element->forTemplate()` directly, I wouldn’t expect `null`.

I’d suggest filtering the elements in `ElementalArea::ElementControllers()`, something like

```php
$items = $this->Elements()->filterByCallback(function (BaseElement $item) {
    return $item->canView();
});
```

raissanorth:
Alrighty, reverted my prior changes and went with @kinglozzer's approach - thanks very much for your input!

tests/ElementalAreaTest.php
```diff
@@ -33,6 +33,19 @@ public function testElementControllers()
...
+    public function testViewPermissionsAreChecked()
+    {
+
+        $area = $this->objFromFixture(ElementalArea::class, 'area2');
+        $controllers = $area->ElementControllers();
+        $elements = $area->Elements();
+
+        $this->assertEquals(1, $controllers->count(),
+            'Should be one controller only, since one of the elements is not viewable');
+        $this->assertEquals(2, $elements->count());
```
robbieaverill: Minor: you can use assertCount(2, $elements)

tests/Src/TestElement.php
```diff
@@ -19,4 +20,9 @@ public function getType()
     {
         return 'A test element';
     }
+
+    public function canView($member = null)
+    {
+        return parent::canView($member) && $this->Viewable;
```
robbieaverill: This is testing the behaviour of a stub. Perhaps instead it could implement a real world permission check like “is admin”, then you can log in as a content author in the tests and assert that it’s not viewable

robbieaverill:
Follow up at #265

PR #260, opened 2018-07-16
FIX Elements in use report source list is now lazy loaded allowing for larger data sets

robbieaverill:
This PR moves the report column formatting into the report method for formatting rather than gathering the initial DataList, which allows for the list to be correctly lazy loaded instead of immediately loaded - crucial for large datasets.

Fixes https://github.com/dnadesign/silverstripe-elemental/issues/238

### DIFF
--- a/src/Reports/ElementsInUseReport.php
+++ b/src/Reports/ElementsInUseReport.php
@@ -3,11 +3,9 @@
 namespace DNADesign\Elemental\Reports;
 
 use DNADesign\Elemental\Models\BaseElement;
-use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Forms\GridField\GridField;
-use SilverStripe\ORM\ArrayList;
+use SilverStripe\ORM\DataList;
 use SilverStripe\Reports\Report;
-use SilverStripe\View\ArrayData;
 use SilverStripe\View\Requirements;
 
 class ElementsInUseReport extends Report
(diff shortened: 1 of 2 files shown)


src/Reports/ElementsInUseReport.php
```diff
@@ -84,14 +80,14 @@ public function columns()
...
-     * @param ArrayData $item
+     * @param BaseElement $item
      * @return string
      */
     protected function getEditLink($value, $item)
     {
         return sprintf(
             '<a class="grid-field__link" href="%s" title="%s">%s</a>',
-            $item->EditLink,
+            $item->CMSEditLink(),
```
robbieaverill: Technically this probably violates semver. If anyone is particularly upset about the low likelihood of this change causing issues, I'm OK with putting an if statement in to handle the old ArrayData object too.

tests/Reports/ElementsInUseReportTest.php
```diff
@@ -49,15 +50,8 @@ public function testSourceRecords()
...
-        $this->assertNotContains(BaseElement::class, $records->toArray(), 'BaseElement is excluded');
-
-        $this->assertContainsOnlyInstancesOf(ArrayData::class, $records);
-
-        foreach ($records as $record) {
-            $this->assertNotNull($record->Icon, 'Fields have an icon');
-            $this->assertNotNull($record->Type, 'Fields have a type');
-        }
+        $this->assertInstanceOf(DataList::class, $records);
+        $this->assertContainsOnlyInstancesOf(BaseElement::class, $records);
```
ScopeyNZ: Isn't this technically removing the granularity of "The report does NOT contain `BaseElement` (without extension) objects"?

robbieaverill: Yeah, I should add it back. Thanks

robbieaverill:
@frankmullenger perhaps you want this in the 2.1 release line? I'll assume you do and change the branch for that

robbieaverill:
To be fair it was wrong to begin with so this is just correcting it

Issue #272, opened 2018-07-29
Convert ElementalEditor GridField to a React based equivalent


robbieaverill:
Prior to beginning work on inline editing of content blocks, we'll need to:

Create a new react app - e.g. ElementalEditor and power as much of this with React, GraphQL and form factory/schema as possible now to pave the way for the react driven CMS in future.

We don't need to continue using gridfield for elemental, particularly because the editor only really has one column.

## Acceptance criteria

- [x] Keep current styling, but update where it makes sense
- [x] Re-build current functionality to be utilising React
- [x] Page will have one elemental area
- [x] Behat tests
- [x] Extensible
- drag and drop isn't expected to work on completion of this issue (see #285)

## Subtasks

- [x] Branch master -> 3 for 3.x support and make master 4.x @robbieaverill
- [x] Add necessary GraphQL query configuration and query HOC #274 @robbieaverill #274
- [x] Bind GraphQL query to ElementList container @robbieaverill #277
- [x] Create all empty React components and containers #275 @ScopeyNZ
- [x] Create skeleton ElementEditor container #275 @ScopeyNZ
- [x] Create skeleton ElementList container #275 @ScopeyNZ
- [x] Create skeleton Element container #275 @ScopeyNZ
- [x] Create skeleton Toolbar container #275 @ScopeyNZ
- [x] Create skeleton Header component #275 @ScopeyNZ @raissanorth
- [x] Create skeleton Content component #275 @ScopeyNZ
- [x] Create skeleton AddNew component #275 @ScopeyNZ
- [x] Implement current markup and CSS for ElementEditor @ScopeyNZ
- [x] Implement current markup and CSS for ElementList #284
- [x] Implement current markup and CSS for Element #284 @raissanorth
- [x] Implement current markup and CSS for Toolbar #281 @ScopeyNZ
- [x] Implement current markup and CSS for Header #278 @raissanorth
- [x] Implement current markup and CSS for Content #284 @raissanorth
- [x] Implement current markup and CSS for AddNew #281 @ScopeyNZ
- [x] Add start of upgrade guide (contribute to it ongoing) #283 
- [x] Enable delete action from block level
- [x] Edit ability (via GridFieldDetailForm) still works

robbieaverill:
Clarification on an acceptance criteria point:

>(A/C) Page will have one elemental area

Elemental currently supports multiple elemental areas per page, but only provides one out of the box.

[...]

PR #275, opened 2018-07-30
Initial work on converting elemental to react

ScopeyNZ:
This PR lays the foundation for converting the elemental block edit gridfield to react.

- `ElementalEditor` is converted to a react container (extending `FormField` now)
- Logical react component shells have been created
- The new `ElementalEditor` react base container component is entwined in.

This is work related to #272


client/src/components/ElementEditor/containers/ElementEditor.js
```diff
@@ -0,0 +1,14 @@
+import React, { Component, PropTypes } from 'react';
+
+class ElementEditor extends Component {
+  render() {
+    return (
+      <div />
+    );
+  }
+}
```
robbieaverill: I'm not 100% on the file paths here. I think I would've expected `client/src/components/ComponentName` or `client/src/containers/ContainerName`. What do you think? I realise that ElementEditor and HistoryViewer separation is a good idea though...

ScopeyNZ: I think it makes sense to keep the component domains separate (not sure if that's the right term but whatever). It feels slightly intense to have `import 'components/ElementEditor/containers/ElementEditor'` but I think it's better than combining `ElementalEditor and `HistoricElementView`.

Issue #276, opened 2018-07-30
Display versioned state of elements in react-based page


brynwhyman:
Ensure that the versioned state of blocks are returned to the blocks page.

This has been sub-tasked out from #272 due to the dependancies of two bugs: ~https://github.com/silverstripe/silverstripe-graphql/issues/153#issuecomment-408735721 and~ https://github.com/silverstripe/silverstripe-versioned/issues/147#issuecomment-408736435

robbieaverill:
I've triaged this in a way that assumes the blockers have been resolved

robbieaverill:
Yeah totally, what you've done is what we didn't do last time and why we created this ticket. The indicators were new features as far as I remember.

Done in #374 - thanks @raissanorth

PR #278, opened 2018-07-30
Update skeleton Header component

raissanorth:
Styling added, but component still needs to be supplied with props. Also, we need to make the delete and edit button work. 

Would appreciate some inital feedback. Thanks!


client/src/components/ElementEditor/ElementEditor.js
```diff
@@ -13,6 +13,7 @@ class ElementEditor extends PureComponent {
       <div className="element-editor">
         <ToolbarComponent />
         <ListComponent pageId={pageId} />
+        <HeaderComponent/>
```
robbieaverill: The HeaderComponent should be added to the Element, rather than this part, right?

client/src/components/ElementEditor/Header.js
```diff
@@ -1,14 +1,93 @@
...
+
     return (
-      <div />
+      <div className="element__heading">
+        <div className="element-header__info">
+          <div className="element-preview__icon">
+            <i className={fontIcon} />
+            {
+              modified &&
+              <span className={spanClassNames} title="Item has unpublished changes" />
```
robbieaverill: We need to make this string translatable. Same with the "Edit" and "Delete" bits further down

client/src/components/ElementEditor/Header.js
```diff
@@ -1,14 +1,93 @@
...
+            <span className="sr-only">
+              Edit
+            </span>
+          </a>
+          <button
+            type="submit"
+            className={deleteClassNames}
+            aria-label="Delete"
+            title="Delete"
+            data-url="admin/pages/edit/EditForm/8/field/ElementalArea"
```
robbieaverill: We'll need to inject this URL somehow, it can't really contain hard coded URLs or IDs

client/src/components/ElementEditor/Header.js
```diff
@@ -1,14 +1,93 @@
...
+Header.defaultProps = {
+  // For testing
+  title: 'Issa\'s Component',
+  elementType: 'Content',
+  fontIcon: 'font-icon-block-content',
+  modified: true,
+  draft: false,
+  moreActions: <div/>,
+  elementId: 1,
+  pageId: 8,
```
robbieaverill: Can you remove these please

client/src/styles/elementeditor.scss
```diff
@@ -0,0 +1,92 @@
+.element-editor {
```
robbieaverill: Now that we're making all of this stuff React driven we can name our stylesheets to match the React components that they're for - e.g. this file could be called `client/src/components/ElementEditor/Header.scss`

raissanorth:
@robbie Made requested changes.

client/src/components/ElementEditor/Element.js
```diff
@@ -1,21 +1,29 @@
...
     return (
       <div className="element-editor__element">
-        <p>#{ID}: {Title}</p>
+        <HeaderComponent
+          elementId={ID}
+          title={Title}
+          elementType={Type}
+          fontIcon={Icon}
+          modified={Modified}
+          draft={Draft}
```
robbieaverill: I'd like to make this more of a "versionedState" than individual props, but can be handled separately (see #276)

client/src/components/ElementEditor/Element.js
```diff
@@ -24,10 +32,19 @@ class Element extends PureComponent {
...
 
-export default Element;
+export { Element as Component };
+
+export default inject(
+  ['ElementHeader', ],
+  (HeaderComponent,) => ({
+    HeaderComponent,
+  }),
+  () => 'Element'
```
robbieaverill: I think we could use a more descriptive injector context here - how about `ElementEditor.ElementList.Editor` (since the parents have parts of that already)

client/src/components/ElementEditor/ElementEditor.js
```diff
@@ -25,8 +25,8 @@ ElementEditor.propTypes = {
 ElementEditor.defaultProps = {};
 
 export default inject(
-  ['ElementToolbar', 'ElementList'],
-  (ToolbarComponent, ListComponent) => ({
+  ['ElementToolbar', 'ElementList', ],
+  (ToolbarComponent, ListComponent, ) => ({
```
robbieaverill: Does the linter require this?

client/src/components/ElementEditor/Header.js
```diff
@@ -1,14 +1,85 @@
...
+        {/* <button */}
+        {/* type="submit" */}
+        {/* className={deleteClassNames} */}
+        {/* aria-label="Delete" */}
+        {/* title={i18n._t('Heading.DELETE', 'Delete')} */}
+        {/* data-url="admin/pages/edit/EditForm/8/field/ElementalArea" */}
+        {/* > */}
+        {/* <span className="btn__title" /> */}
+        {/* </button> */}
+        {/* </div> */}
```
robbieaverill: Hmm, maybe deleting it entirely would be cleaner for now...

client/src/components/ElementEditor/Header.scss
```diff
@@ -0,0 +1,92 @@
+.element-editor {
```
robbieaverill: Ok can we please refactor this stylesheet a little so that the class names are more specific? I'm imagining that the "block" in BEM in this component is `element-editor-header` always, and the "element" would be things like "heading", "title", "icon" etc

client/src/components/ElementEditor/Content.js
```diff
@@ -2,13 +2,19 @@ import React, { Component, PropTypes } from 'react';
...
   }
 }
 
 Content.defaultProps = {};
-Content.propTypes = {};
+Content.propTypes = {
+  summary: PropTypes.string,
+};
 
+export { Content as Component };
```
robbieaverill: This can be done in a separate pull request since it's not part of the Header. If we're gonna keep it here then I think we need to change BaseElement to return a string for the thumbnail URL and we render it from this end rather than using `dangerouslySetInnerHTML`

client/src/components/ElementEditor/Element.js
```diff
@@ -1,33 +1,52 @@
...
 /**
  * The Element component used in the context of an ElementEditor shows the summary
  * of an element's details when used in the CMS, including ID, Title and Summary.
  */
 class Element extends PureComponent {
   render() {
-    const { element: { ID, Title, Summary } } = this.props;
+    const { element: { ID, Title, Summary, Type, IconClass },
+      HeaderComponent,
+      ContentComponent } = this.props;
```
robbieaverill: I'm not a huge fan of this formatting, how would this be instead?

```js
const {
  element: { ID, Title, Summary, Type, IconClass },
  HeaderComponent,
  ContentComponent,
} = this.props;
```

client/src/components/ElementEditor/Element.js
```diff
@@ -1,33 +1,52 @@
...
 
-export default Element;
+export { Element as Component };
+
+export default inject(
+  ['ElementHeader', 'ElementContent'],
+  (HeaderComponent, ContentComponent) => ({
+    HeaderComponent, ContentComponent,
+  }),
+  () => 'Element'
```
robbieaverill: Can we please make this `ElementEditor.ElementList.Element`? The two parents above this are using `ElementEditor` and `ElementEditor.ElementList` respectively already

client/src/components/ElementEditor/Header.js
```diff
@@ -1,14 +1,55 @@
...
+    });
+  }
+
   render() {
+    const { id, title, elementType, fontIcon } = this.props;
+
     return (
-      <div />
+      <div className="element-editor-header">
+        <div className="element-editor-header-info">
```
robbieaverill: I think `element-editor-header__info` would be more BEM

client/src/components/ElementEditor/Header.js
```diff
@@ -1,14 +1,55 @@
...
+          <div className="element-editor-header-icon__container">
```
robbieaverill: `element-editor-header__icon-container`

client/src/components/ElementEditor/Header.js
```diff
@@ -1,14 +1,55 @@
...
+            <i className={fontIcon} id={`element-editor-header-icon${id}`} />
```
robbieaverill: Can we add a `element-editor-header__icon` as well as `fontIcon`?

client/src/components/ElementEditor/Header.js
```diff
@@ -1,14 +1,55 @@
...
+            <Tooltip
+              placement="top"
+              isOpen={this.state.tooltipOpen}
+              target={`element-editor-header-icon${id}`}
+              toggle={this.toggle}
+            >
+              {elementType}
+            </Tooltip>
+          </div>
+          <h3 className="element-editor-header-title">{title}</h3>
```
robbieaverill: `element-editor-header__title`

client/src/components/ElementEditor/Header.js
```diff
@@ -1,14 +1,55 @@
...
+};
+
+Header.propTypes = {
+  id: PropTypes.number,
+  title: PropTypes.string,
+  elementType: PropTypes.string,
+  fontIcon: PropTypes.string,
+};
 
+export { Header as Component };
```
robbieaverill: We don't really need this unless we're exporting default with HOCs applied

client/src/components/ElementEditor/Header.scss
```diff
@@ -0,0 +1,30 @@
+.element-editor-header {
+  display: flex;
+  justify-content: space-between;
+
+  .element-editor-header-title {
```
robbieaverill: See note in the component about the class names. The main benefit of using BEM class naming is that we don't need to nest the class definitions like this. If you agree and we update these then they can all be top level definitions with no nesting.

client/src/styles/bundle.scss
```diff
@@ -2,10 +2,16 @@
 @import "variables";
 
 // Thirdparty
+@import "~bootstrap/scss/functions";
```
robbieaverill: Were the functions and variables required for the existing SCSS to compile?

client/src/styles/bundle.scss
```diff
@@ -2,10 +2,16 @@
...
+@import "~bootstrap/scss/variables";
 
 // Elemental components
 @import "admin";
 @import "history";
 @import "reports";
 @import "textcheckboxgroup";
+@import "../components/ElementEditor/Content";
+@import "../components/ElementEditor/ElementEditor";
+@import "../components/ElementEditor/Header";
```
robbieaverill: Awesome! I see you've added the Header.scss file for example - is there a corresponding amount of dead SCSS in the original stylesheet that can be deleted now?

client/src/components/ElementEditor/Element.js
raissanorth: sure, will change.

client/src/components/ElementEditor/Element.js
raissanorth: of course.

client/src/components/ElementEditor/Header.js
raissanorth: Yeah, I have seen it written like that a few times before, indeed. It's confusing, as http://getbem.com/introduction/ quotes that particular style:
image (Which I am aware I haven't been following strictly either). Your call, shall I follow BEM BEM or SilverStripe BEM?!

client/src/styles/bundle.scss
raissanorth: It was before I stripped out the edit and delete button. Will re-check if this is still needed.

raissanorth: Yup, required indeed.

client/src/components/ElementEditor/Header.scss
raissanorth: I agree. Changes made, but not pushed yet.

client/src/components/ElementEditor/Header.js
robbieaverill: SilverStripe BEM === BEM 😄 (or should be at least)

In this case `element-editor-header` is the block, and `info`, `title`, `icon` etc is the element. See http://getbem.com/naming/

PR #279, opened 2018-07-30
FIX Ensure that SiteTree::ElementArea is always available for GraphQL queries

robbieaverill:
This fixes an issue with GraphQL query consistency when applied to SiteTree and the Elemental extension is applied to a subclass of SiteTree, by providing a consistent proxy method to use for the queries instead.

Issue: #272


src/Extensions/SiteTreeExtension.php
```diff
@@ -0,0 +1,25 @@
...
+ * of whether the specific implementation of SiteTree (subclass) has the elemental extension applied.
+ */
+class SiteTreeExtension extends DataExtension
+{
+    /**
+     * Returns the owner's `ElementalArea()` contents if it has the extension
+     *
+     * @return ElementalArea|null
+     */
+    public function ElementArea()
```
robbieaverill: Open to ideas here. Perhaps this naming might be more confusing than useful - we could rename it to `ElementalAreaProxy()` or something instead?

ScopeyNZ: Hmmn. It's a hard one. Perhaps something like `ElementalAreaIfExists`? If we assume the majority of pages don't have the elemental extension that name gives more indication of what it means...

PR #281, opened 2018-07-31
ADD React version of the elemental add block control

ScopeyNZ:
Here's a WIP PR for the add block react component that appears above the list of elements in an elemental area. It works, except the link where the add form is doesn't work - we've removed the gridfield and it was implemented as a gridfield component. I'm a bit out of my depth finding a good (quick) solution for this.

I also refactored the `ElementEditor` form field to instead be based off the `ElementalAreaField` field ensuring that we don't lose the existing functionality implemented in that field. `ElementalEditor` was a factory (albeit not very obvious) that isn't required if we're dropping gridfield.

Continues work towards #272


client/src/components/ElementEditor/AddNewButton.js
```diff
@@ -1,14 +1,70 @@
...
+    const { selectedType } = this.state;
+
+    if (!selectedType) {
+      return '#';
+    }
+
+    return this.props.baseEditHref + selectedType;
+  }
+
+  renderOptions() {
```
robbieaverill: Can we have some doc blocks plz

client/src/components/ElementEditor/AddNewButton.js
```diff
@@ -1,14 +1,70 @@
...
+      ));
+  }
+
+
   render() {
     return (
-      <div />
+      <InputGroup className="elemental-editor__add-new-block-control">
+        <Input type="select" className="no-change-track" onChange={this.onTypeChange}>
+          <option>(Select type to create)</option>
```
robbieaverill: Add translation: `i18n._t('AddNewButton.SELECT', 'Select type to create')`

client/src/components/ElementEditor/Toolbar.js
```diff
@@ -1,15 +1,28 @@
...
+  AddNewButtonComponent: PropTypes.oneOfType([PropTypes.node, PropTypes.func]).isRequired,
+};
 
-export default Toolbar;
+export default inject(
+  ['ElementAddNewButton'],
+  (AddNewButtonComponent) => ({
+    AddNewButtonComponent,
+  }),
+  () => 'ElementToolbar'
```
robbieaverill: Can you make this `ElementEditor.ElementToolbar` so it's under the context of being in an ElementEditor?

ScopeyNZ: Isn't this supposed to match the name which it's registered under Injector?

robbieaverill: Nah, this is the "context key" - essentially what this does is say "give me the add new button component, in the context of `ElementToolbar`." This allows devs to configure specific HOCs for `AddNewButtonComponent` in different contexts.

In this case it's important to highlight that the context is within `ElementEditor`

client/src/components/ElementEditor/AddNewButton.js
```diff
@@ -1,14 +1,98 @@
...
+   */
+  renderOptions() {
+    const { elementTypes } = this.props;
+
+    if (!elementTypes) {
+      return null;
+    }
+
+    return elementTypes.map(type => (
+      <option value={type.value}>{type.title}</option>
```
robbieaverill: I think they need a `key` as well - does React not complain about this when compiling? If not, don't worry

ScopeyNZ: I didn't get any linting issues or any warnings on compile but I'll recheck when I rebase.

PR #211, opened 2018-03-21
add ListingHelper to access iterator information like $Pos, $First, $Last and $EvenOdd

banglashi:
Hi there,

We had been developing a [similar module](https://github.com/arillo/silverstripe-elements) (mainly for SS3) where we had the necessity to access the iterator information, like $Pos, $First, $Last and $EvenOdd when iterating over blocks. This is a simple helper to achieve just this. Let me know if you need anything or you had a different approach in mind.

PR #282, opened 2018-07-31
Add First, Last, Pos, TotalItems and EvenOdd helpers

wilr:
When in an Element template the expected $First, $Last helpers aren't there so this replaces them with something.

### DIFF
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -229,8 +229,8 @@ class BaseElement extends DataObject
     }
 
     /**
-     * Increment the sort order if one hasn't been already defined. This ensures that new elements are created
-     * at the end of the list by default.
+     * Increment the sort order if one hasn't been already defined. This
+     * ensures that new elements are created at the end of the list by default.
      *
      * {@inheritDoc}
      */
@@ -239,7 +239,15 @@ class BaseElement extends DataObject
         parent::onBeforeWrite();
 
         if (!$this->Sort) {
-            $this->Sort = static::get()->max('Sort') + 1;
+            if ($this->hasExtension(Versioned::class)) {
+                $records = Versioned::get_by_stage(BaseElement::class, Versioned::DRAFT);
+            } else {
+                $records = BaseElement::get()->max('Sort');
+            }
+
+            $records = $records->filter('ParentID', $this->ParentID);
+
+            $this->Sort = $records->max('Sort') + 1;
         }
     }
 
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -830,4 +830,48 @@ class BaseElement extends DataObject
 
         return null;
     }
+
+    /**
+     * @return boolean
+     */
+    public function First()
+    {
+        return ($this->Parent()->Elements()->first()->ID === $this->ID);
+    }
+
+    /**
+     * @return boolean
+     */
+    public function Last()
+    {
+        return ($this->Parent()->Elements()->last()->ID === $this->ID);
+    }
+
+    /**
+     * @return int
+     */
+    public function TotalItems()
+    {
+        return $this->Parent()->Elements()->count();
+    }
+
+    /**
+     * Returns the position of the current element.
+     *
+     * @return int
+     */
+    public function Pos()
+    {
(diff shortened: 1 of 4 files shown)


src/Models/BaseElement.php
```diff
@@ -822,4 +830,28 @@ public function getPageTitle()
 
         return null;
     }
+
+    /**
+     * @return boolean
+     */
+    public function isFirstElement()
+    {
+        return !($this->Parent()->Elements()->filter('Sort:LessThan', $this->Sort)->exists());
```
robbieaverill: What do you reckon about `$this->Parent()->Elements()->first() === $this`?

wilr: Would `$this` be robust enough? I'm just thinking of objects perhaps changing at runtime with something like fluent etc.. 

Perhaps `$this->Parent()->Elements()->first()->ID === $this->ID`. ID is all that matters right?

robbieaverill: Yeah give that a go. Your tests should be good enough to try it out =D

robbieaverill:
Related: #211

wilr:
@robbieaverill What to you think like #211 should we do something at the ElementArea level or leave it as functions on `BaseElement` if we leave it as methods on BaseElement then perhaps it would be more expected behaviour that `First`, `Last` be the method names rather than `isFirstElement`

robbieaverill:
@wilr I'm happy to follow your judgement =) they sound like useful methods either way

wilr:
Ok my judgement is consistently probably out weighs risk since the behaviour for all the options is going to be the same unless you start to do funky things like building your own `ArrayList` of them. In which case you could override First, Last anyway if they didn't work.

Updated the code, tests and docs. Changed branch to 3

src/Models/BaseElement.php
```diff
@@ -822,4 +830,48 @@ public function getPageTitle()
...
+    /**
+     * @return boolean
+     */
+    public function Last()
+    {
+        return ($this->Parent()->Elements()->last()->ID === $this->ID);
+    }
+
+    /**
+     * @return boolean
```
robbieaverill: Minor: should be int

src/Models/BaseElement.php
```diff
@@ -822,4 +830,48 @@ public function getPageTitle()
...
+     */
+    public function TotalItems()
+    {
+        return $this->Parent()->Elements()->count();
+    }
+
+    /**
+     * Returns the position of the current element.
+     *
+     * @return boolean
```
robbieaverill: int

src/Models/BaseElement.php
```diff
@@ -822,4 +830,48 @@ public function getPageTitle()
...
+     */
+    public function Pos()
+    {
+        return ($this->Parent()->Elements()->filter('Sort:LessThan', $this->Sort)->count() + 1);
+    }
+
+    /**
+     * @return boolean
```
robbieaverill: string

src/Models/BaseElement.php
```diff
@@ -239,7 +239,15 @@ public function onBeforeWrite()
         parent::onBeforeWrite();
 
         if (!$this->Sort) {
-            $this->Sort = static::get()->max('Sort') + 1;
+            if ($this->hasExtension(Versioned::class)) {
+                $records = Versioned::get_by_stage(BaseElement::class, Versioned::DRAFT);
+            } else {
+                $records = BaseElement::get()->max('Sort');
```
ScopeyNZ: Sorry for the late review but it appears this is a bug?

wilr: @ScopeyNZ What seems to be the bug? I added this in as it was skipping draft version numbers and it was also using the global max rather than the local max.

ScopeyNZ: `$records = ...->max('Sort')` should return an int, but is treated like an object on the next line?

wilr: Ah yep, but only if versioned had been removed with is rare. I'll submit a fix to the 3 branch.

Thanks

ScopeyNZ: Sweet 👍

PR #284, opened 2018-08-01
Element and Content component updates

raissanorth:
Requesting some initial feedback, please:

- Updates GraphQL query to retrieve structured data for files and content rather then preformatted html. 

- Also fixes some minor errors from prior commits. 
- Adds some initial styling. 

- Relies on changes in https://github.com/creative-commoners/silverstripe-elemental-blocks/tree/elemental-react-app


src/Models/ElementContent.php
```diff
@@ -41,6 +41,11 @@ public function getSummary()
         return DBField::create_field('HTMLText', $this->HTML)->Summary(20);
     }
 
+    public function getContent()
+    {
+        return DBField::create_field('HTMLText', $this->HTML)->Summary(20);
```
ScopeyNZ: I think it makes sense to use `getSummary` here instead, and refer to it as `ContentSummary` or something in the JS - it might get confusing referring to it as content because you're just taking a summary :wink:

robbieaverill: Great idea. This is pretty similar to how `FormField::getSchemaData` communicates an arbitrary array of data from PHP to React form field components, then they're in charge of dealing with the data that their PHP counterparts provides. We could then combine `getIconClass` into this too

raissanorth:
For reference, this is what it looks like atm:

image

robbieaverill:
@raissanorth looks like we're missing the border radius on the image. In the 3.x elemental we have these styles for that:

```css
.elemental-preview__thumbnail-image img {
    border-radius: .192rem;
}
```

raissanorth:
@robbie thanks for pointing out the border-radius! i looked at the designs and thought there was no radius applied, eek! 👓  sorting it out now.

src/Extensions/SchemaScaffolderExtension.php
```diff
@@ -0,0 +1,25 @@
...
+            'serialize' => function ($value) {
+                return (object) $value;
+            },
+            'parseValue' => function ($value) {
+                return (array) $value;
+            },
+            'parseLiteral' => function ($ast) {
+                return $ast->value;
+            },
+        ]), 'ObjectType');
```
robbieaverill: My only concern with this is that it's applied globally to the DataObjectScaffolder. 

Essentially what this is doing is adding a new type, right? We could isolate this by pointing the scaffolding configuration at a `BaseElementTypeCreator` class (extends TypeCreator), which could then add this.

Otherwise, perhaps this needs to be added directly into core...

cc @unclecheese

src/Models/BaseElement.php
```diff
@@ -694,14 +699,48 @@ public function getSummary()
...
 
+
+    /**
+     * @return array
+     */
+    public function getSummaryData()
+    {
+        $summaryData = [];
+
+        $this->extend('setDefaultSummaryData', $summaryData);
```
robbieaverill: The usual convention would be `$this->extend('updateSummaryData', $summaryData);`

src/Models/BaseElement.php
```diff
@@ -694,14 +699,48 @@ public function getSummary()
...
+
+    /**
+     * This can be overridden on child elements.
+     *
+     * @return string
+     */
+    public function getFileURL()
+    {
+        return '';
+    }
```
robbieaverill: I think summary data should mean we can remove this now. Doing so might mean that the file block needs a custom extension for the `Element` component, but that would also be a good thing for us dogfooding this architecture

src/Extensions/SchemaScaffolderExtension.php
ScopeyNZ: It probably would make more sense to use `TypeCreator` - although I don't think it's going to change the fact it's applied globally. However it probably makes more sense to namespace the type a little unless it will become an addition to admin.

src/GraphQL/Types/ObjectType.php
```diff
@@ -20,6 +19,6 @@ public function onBeforeAddToManager(Manager $manager)
             'parseLiteral' => function ($ast) {
                 return $ast->value;
             },
-        ]), 'ObjectType');
+        ]);
```
robbieaverill: This is much better, IMO

client/src/components/ElementEditor/ElementEditor.scss
```diff
@@ -1,3 +1,9 @@
 .element-editor {
   background-color:  $white;
+
+  margin-top: calc(-1rem + -1px);
```
NightJar: I realise this is largely my fault, but I think this needs a comment to describe where these magic numbers come from. The -1rem may also have a variable - but in the least the `-1px` needs to be explained as matching the border-bottom-width of the field above it.

_config/graphql.yml
```diff
@@ -3,11 +3,13 @@ Name: elementalgraphqlconfig
...
 SilverStripe\GraphQL\Controller:
   schema:
+    types:
+      ObjectType: DNADesign\Elemental\GraphQL\Types\ObjectType
     scaffolding:
       types:
         # Expose common static fields for the ElementEditor component to use for preview summaries
         DNADesign\Elemental\Models\BaseElement:
-          fields: [ID, LastEdited, AbsoluteLink, Title, ShowTitle, Sort, Summary, Type, IconClass]
+          fields: [ID, LastEdited, AbsoluteLink, Title, ShowTitle, Sort, Type, IconClass, SummaryData]
```
robbieaverill: SummaryData could also include IconClass and Type, yeah?

client/src/components/ElementEditor/ElementEditor.scss
raissanorth: Sweet, will add a comment now.

robbieaverill: What's the computed value? It probably is a variable we can use

client/src/components/ElementEditor/tests/Content-test.js
```diff
@@ -0,0 +1,63 @@
...
+
+import React from 'react';
+import Content from '../Content';
+import Enzyme, { shallow } from 'enzyme';
+import Adapter from 'enzyme-adapter-react-15.4/build/index';
+
+Enzyme.configure({ adapter: new Adapter() });
+
+describe('Content', () => {
+  describe('conditionally render image', () => {
```
robbieaverill: `describe('functionName()', () => {` -> `render()`?

src/ORM/FieldType/DBObjectType.php
```diff
@@ -0,0 +1,21 @@
...
+use SilverStripe\ORM\FieldType\DBField;
+
+class DBObjectType extends DBField
+{
+    /**
+     * Add the field to the underlying database.
+     */
+    public function requireField()
+    {
+        // I am mad that I have to add this.
```
NightJar: ???

client/src/components/ElementEditor/tests/Content-test.js
```diff
@@ -0,0 +1,63 @@
...
+          fileTitle=""
+          content="Sample content"
+        />
+      );
+
+      expect(wrapper.find('img').length).toBe(0);
+    });
+  });
+
+  describe('conditionally render content summary', () => {
```
robbieaverill: If we are still describing the render() method then we can remove this line

client/src/components/ElementEditor/tests/ElementEditor-test.js
```diff
@@ -0,0 +1,28 @@
...
+import Enzyme, { shallow } from 'enzyme';
+import Adapter from 'enzyme-adapter-react-15.4/build/index';
+
+Enzyme.configure({ adapter: new Adapter() });
+
+describe('ElementEditor', () => {
+  const ToolbarComponent = () => <div />;
+  const ListComponent = () => <div className="elemental-editor__list" />;
+
+  it('should render ElementList and Toolbar', () => {
```
robbieaverill: Add `describe('render()', () => {` wrapper

client/src/components/ElementEditor/tests/Header-test.js
```diff
@@ -0,0 +1,59 @@
...
+import Header from '../Header';
+import Enzyme, { shallow } from 'enzyme';
+import Adapter from 'enzyme-adapter-react-15.4/build/index';
+
+Enzyme.configure({ adapter: new Adapter() });
+
+describe('Header', () => {
+  const Tooltip = () => <div />;
+
+  describe('render', () => {
```
robbieaverill:
```diff
- describe('render', 
+ describe('render()', 
```

src/GraphQL/Types/ObjectType.php
```diff
@@ -0,0 +1,24 @@
+<?php
+namespace DNADesign\Elemental\GraphQL\Types;
+
+use GraphQL\Type\Definition\CustomScalarType;
+use SilverStripe\GraphQL\TypeCreator;
+
+class ObjectType extends TypeCreator
```
robbieaverill: This could do with a PHPDoc block to explain what its for

src/Models/BaseElement.php
```diff
@@ -694,14 +699,37 @@ public function getSummary()
...
 
+
+    /**
+     * @return array
+     */
+    public function getSummaryData()
+    {
+        $summaryData = [];
+
+        $this->extend('updateSummaryData', $summaryData);
```
robbieaverill: The problem with adding an extension point here is that the extension runs before overloaded methods do.

Instead I think we need to do something like this:

```php
public function getSummaryData()
{
    $summaryData = $this->provideSummaryData();
    $this->extend('updateSummaryData', $summaryData);
    return $summaryData;
}

/**
 * Overload this method in order to customise the data fields for React elements, this data is serialised
 * and returned via the GraphQL API
 *
 * @return array
 */
protected function provideSummaryData()
{
    return [];
}
```

src/Models/BaseElement.php
```diff
@@ -694,14 +699,37 @@ public function getSummary()
...
+
+    /**
+     * This can be overridden on child elements.
+     *
+     * @return string
+     */
+    public function getContent()
+    {
+        return '';
+    }
```
robbieaverill: What's this method used for?

src/ORM/FieldType/DBObjectType.php
ScopeyNZ: Yes, it was experimental when I was writing this :joy:. Should be replaced with something like `// noop - This class exists only to bind to a custom GraphQL type`.

src/Models/BaseElement.php
```diff
@@ -694,14 +699,37 @@ public function getSummary()
...
+
     /**
      * Return the icon class name
      *
      * @return string
      */
     public function getIconClass()
     {
-        return $this->config()->get(‘icon’);
+        return $this->config()->get('icon');
```
robbieaverill: This is fixed in the latest master branch, rebase to pull it in =)

src/Models/ElementContent.php
```diff
@@ -41,6 +41,13 @@ public function getSummary()
         return DBField::create_field('HTMLText', $this->HTML)->Summary(20);
     }
 
+    public function getSummaryData()
+    {
+        $summaryData = [];
```
robbieaverill: See previous comment in BaseElement about this. Whichever method is used, this particular line should be `$summaryData = parent::getSummaryData();` rather than entirely overloading it

client/src/components/ElementEditor/ElementEditor.scss
raissanorth: Unfortunately not. 15px come from .form-group and we move up one pixel to hide the last form element's border, which doesn't' spread full width.

src/Models/BaseElement.php
ScopeyNZ: The original intention was to allow people to "extend" in defaults. I guess we could to the provider and all that but I'd rather just remove the extension point at this point if the API is likely to change with upcoming work (and I think it almost certainly will)

PR #289, opened 2018-08-04
[Add] Viewer Groups on BaseElement

JoshuaCarter:
Hey guys,

These changes enable ViewerGroups on individual elements so that individual elements can be hidden from public view and restricted to certain member groups, etc.

I'm keen to know if these are changes you want in the module and what would be needed for you to accept them.

BaseElement:
- Now has Groups (those that can view the element).
- InheritedPermissionsExtension enabled.
- Changed 'canView' to respect viewer settings.
- Existing ViewerGroups/EditorGroups removed.
- New 'ViewerGroups' tab added.
- Added 'Can Edit Type' to Content tab.

***This a remake of https://github.com/dnadesign/silverstripe-elemental/pull/288 to target the 3 branch. As a result of this feedback from @robbieaverill :*

> These changes seem like a good idea. I’ll have a better look at the permission checks on Monday. One thing to be wary of with this kind of thing is full page or partial caching around the elemental area, however it’s often not the responsibility of this module to implement that so it’d be down to devs to do it in a sensible way. Another thing is that master is being rebuilt at the moment, so maybe target the 3 branch (you can change the target branch for PRs via “edit” in top left, then rebase). Thanks for te PR!

### DIFF
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -15,11 +15,18 @@ use SilverStripe\Forms\CheckboxField;
 use SilverStripe\Forms\DropdownField;
 use SilverStripe\Forms\FieldList;
 use SilverStripe\Forms\HiddenField;
+use SilverStripe\Forms\ListboxField;
 use SilverStripe\Forms\NumericField;
+use SilverStripe\Forms\OptionsetField;
 use SilverStripe\Forms\TextField;
+use SilverStripe\Forms\TreeMultiselectField;
 use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\FieldType\DBField;
 use SilverStripe\ORM\FieldType\DBHTMLText;
+use SilverStripe\Security\Security;
+use SilverStripe\Security\Group;
+use SilverStripe\Security\InheritedPermissions;
+use SilverStripe\Security\InheritedPermissionsExtension;
 use SilverStripe\Security\Member;
 use SilverStripe\Security\Permission;
 use SilverStripe\Versioned\Versioned;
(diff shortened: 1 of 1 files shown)


robbieaverill:
Hey @JoshuaCarter, could you please fix the broken test?

robbieaverill:
Yeah fair enough. You're right, it's an existing test looking at canView permissions on elements which is now failing with your changes involved. In some respects this is good, because it points out regressions in behaviour. In another respect we'll need to look at the new changes and either update the existing tests and/or write new ones for the new code.

JoshuaCarter:
Cool, I'm happy to do whatever is needed provided I understand what is needed, which I can't guarantee xD

robbieaverill:
Thanks, my suggestion would be:

* Look at why the existing tests are failing, check whether the failure is a real change in behaviour or just a coincidence due to the way the tests are set up now. For example, they may have only scaffolded the minimum amount of data required in order to test the way canView works right now, so your change may be fine in practice but the tests fail because they're being too naive in their expectations.
* Update the tests so that they pass.
* Decide whether you need to add any new tests for your code. My suspicion is that the changes should really have some new tests to go with them. There are plenty of tests for canView etc that you could use for reference, for example in the CMS and blog modules.

See how you get on. I can't guarantee that we will have time to look at this immediately, but I can add it to my todo list if you get stuck.

Also happy to provide advice on the SilverStripe Users Slack channel.

robbieaverill:
Hey @JoshuaCarter, thanks again for the pull request. Please feel free to reopen it if you get a chance to get the tests passing or write some new ones =)

PR #301, opened 2018-08-06
NEW Implementing edit link on inline blocks and updating styling

ScopeyNZ:
Each block is now an anchor tag that links to the GridFieldDetailForm for the element. I updated the hover styling in this PR too.


client/src/components/ElementEditor/Element.js
```diff
@@ -11,15 +12,19 @@ class Element extends PureComponent {
...
+      i18n._t('ElementalElement.TITLE', 'Edit this {type} block'), { type: BlockSchema.type }
+    );
 
     if (!ID) {
       return null;
     }
 
     return (
-      <div className="element-editor__element">
+      <a className="element-editor__element" href={link} title={linkTitle}>
```
robbieaverill: Hmm, I don't know if I like this. Can you nest the anchor inside the div instead?

ScopeyNZ: I will have to remove all padding from the div and add it to the anchor - which is what that class did - so it made sense to move the class to the anchor tag, then the div was kind of pointless. Not really sure what a better solution to that would be?

robbieaverill: Ok, I'll get over it

PR #313, opened 2018-08-12
NEW Block schema is now extensible, getBlockSchema marked as internal

robbieaverill:
This change ensures that the block schema can be modified by extensions, and that overloaded methods will run before the extension does.


robbieaverill:
Done, cheers

Issue #309, opened 2018-08-08
Implement loading indicator while blocks load


ScopeyNZ:
With the new react editor the blocks now load asynchronously. While they load we need a loading indicator.

## Pull requests

 - [x] https://github.com/dnadesign/silverstripe-elemental/pull/316

robbieaverill:
FYI the admin module has a Loading component that can be rendered while the `loading` prop is true (GraphQL HOCs update this prop as necessary on the components that they are bound to).

Should be an easy fix.

brynwhyman:
@clarkepaul can you please check there is a loading indicator in the style guide or pattern library to use for this?

clarkepaul:
Yup will sort @brynwhyman

robbieaverill:
Loading indicator (React component in admin) from pattern library:

![image](https://user-images.githubusercontent.com/5170590/43809773-3feb5830-9b08-11e8-8893-84c9177a53e9.png)

clarkepaul:
We use that for entire page loads but I'm not sure if its as practical for sections of UI, will confirm with Jared.

robbieaverill:
@clarkepaul (cc @newleeland) we're already using this component for loading in sections of the UI, sometimes with multiple sections in one browser window at once. It's cool if that needs to change, perhaps we could create a new loading indicator for UI panels and mark this one as only to be used for full windows

newleeland:
It would be good to modernize the loading indicator but I think as long as we don't have 3+ loading indicators appearing per the loading of a view it should be fine.

robbieaverill:
In order to prevent a complex UI from having its subcomponents use the Loading component when loading async data, the component would need to share state between its child components somehow to tell them whether they're allowed to render their loading components or not.

I think this should be raised in core, and asset admin would be a good place to start (open the detail view for a file and refresh the page to see two loading indicators at the same time).

robbieaverill:
I had a chat the other day with @newleeland about loading indicators. There are five or six varieties in the CMS at the moment, so we decided to use the Loading React component which ships in the admin module (shown in screenshot above) for now and raise a core issue (TBC) to assess and consolidate loading indicators across the CMS.

robbieaverill:
PR at https://github.com/dnadesign/silverstripe-elemental/pull/316

PR #314, opened 2018-08-13
NEW Elements can be deleted via the inline delete button. Fix versioning state in readBlocks query

### LINKED ISSUE #305: Implement delete button
The existing elemental editor (PHP GridField) has a delete button for quickly removing blocks.

The new React version needs this too. This will need to be written as a GraphQL mutation applied via an HOC.

## Acceptance criteria

- [x] Delete button shows on elements
- [x] Clicking the delete button removes the row
- [x] The element does not appear in the list on reload

## Pull requests

- [x] https://github.com/dnadesign/silverstripe-elemental/pull/314

robbieaverill:
This pull request adds the ability to delete an element from the editor using a similar button to that of the PHP GridField delete action.

Other changes:

* `.element-editor__element` is now a `span` rather than an anchor, since we're nesting a button inside it. It has the necessary `role` and `tabIndex` as well as click and keypress handlers for accessiblity.
* `ReadBlocksForPage` GraphQL query's versioning mode was incorrectly reading from the "latest" stage, it should be reading from draft since it's in the CMS

After the delete mutation runs it refetches the list of elements, which sets the `loading` prop on the `ElementList` to true while it's running. Once we have a loading indicator (#309) this will work automatically.

Resolves #305


ScopeyNZ:
In this case do we have to bother with a refetch? Can't we just find a way to remove the element from the list and be done with it?

robbieaverill:
That would involve mutating the blocks array prop. I think there is a way to do this with apollo's imperative cache API but I'm not sure how to do it yet - this is what we've been doing so far for history viewer anyway

client/src/components/ElementEditor/Header.js
```diff
@@ -36,19 +63,39 @@ class Header extends Component {
           </div>
           <h3 className="element-editor-header__title">{title}</h3>
         </div>
+        <FormActionComponent
+          onClick={this.handleDelete}
+          title={null}
```
ScopeyNZ: This seems like a bug in admin 😒

ScopeyNZ:
Do you think we should include some sort of loading indication during deletion before this is merged? Currently you are just staring at the block you deleted for a bit before it actually disappears.

robbieaverill:
@ScopeyNZ yep, the loading prop is set by GraphQL but there's no indicator yet (noted in description)

robbieaverill:
Merging with @ScopeyNZ's approval

client/src/components/ElementEditor/Header.js
robbieaverill: To be fair, we could probably just omit the title prop

ScopeyNZ: You can - but you would also expect it to be passed through to the actual button that gets rendered...

ScopeyNZ: It already passed through a few attributes: https://github.com/silverstripe/silverstripe-admin/blob/1/client/src/components/FormAction/FormAction.js#L22-L26

PR #319, opened 2018-08-14
Fix infinite loop on write

Lolly101:
Fixes https://github.com/dnadesign/silverstripe-elemental/issues/318

### DIFF
--- a/src/Models/ElementalArea.php
+++ b/src/Models/ElementalArea.php
@@ -121,6 +121,12 @@ class ElementalArea extends DataObject
         return $controllers;
     }
 
+    public function onBeforeWrite()
+    {
+        $this->getOwnerPage();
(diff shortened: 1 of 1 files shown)


robbieaverill:
Looks good, tests pass - 👌 

@Lolly101 Do you want this to be released as a patch fix in the 2.x release line? If so I'll retarget this pull request against 2.0.x or 2.1.x, otherwise it's unlikely to receive a new release since we're currently supporting elemental 3.x by default.

Lolly101:
Yes 2.x please because we're not using SilverStripe 4.2 yet

robbieaverill:
Which version are you using? 2.0 or 2.1?

Lolly101:
2.1 thanks :)

PR #345, opened 2018-08-22
ENH Add action dropdown. Move delete button.

raissanorth:
Partially resolves https://github.com/dnadesign/silverstripe-elemental/issues/322 

The ACs for the above ticket include styling. Outstanding is the amendment of the size of the toggle button icon, which depends on https://github.com/silverstripe/silverstripe-admin/pull/610#pullrequestreview-148328966, so I'll make a seperate PR for this.

Changes in this PR:

-  add an action menu 
- move the action menu into it's own container within the header given that a 'caret' button will be added shortly
- move the delete button into the actionmenu's dropdownmenu and change the form action into a simple button, since a separate task will look into adding children to the action menu

**Depends on https://github.com/silverstripe/silverstripe-admin/pull/610** to prevent the delete button events to bubble up the element itself (the latter would cause a redirect to the edit page)


client/src/components/ElementEditor/Element.js
```diff
@@ -20,9 +20,14 @@ class Element extends Component {
   /**
    * Take the user to the GridFieldDetailForm to edit the record
    */
-  handleClick() {
+  handleClick(event) {
     const { link } = this.props;
 
+    // TODO Replace with something more unique than a class name, perhaps add an ID?
+    if (event.target.className === 'action-menu__toggle btn btn--no-text btn--icon-xl font-icon-dot-3 btn btn-secondary') {
```
robbieaverill: To start with you could check whether the target's [classList](https://developer.mozilla.org/en-US/docs/Web/API/Element/classList) contains `action-menu__toggle`

robbieaverill: I think we did something similar with the Delete button - you should maybe check that. I think it has some code in the click handler for the delete button that stops propagation so the click event doesn't bubble to the parent component as well

client/src/components/ElementEditor/Header.js
```diff
@@ -63,14 +71,22 @@ class Header extends Component {
...
-          onClick={this.handleDelete}
-          title={null}
-          icon="trash-bin"
-          attributes={{
-            title: i18n._t('ElementHeader.DELETE', 'Delete'),
-          }}
-        />
+        <div className="element-editor-header__actions">
+          <ActionMenuComponent
+            id={this.props.id}
```
robbieaverill: Should we prefix this with something unique to the actions menu maybe like `element-editor-actions-${id}` or something? Note `id` is destructured already at this point

client/src/components/ElementEditor/Element.js
ScopeyNZ: I think the issue here is that we can't easily alter the click handler for the toggle button. That's taken care of by the ActionMenu component as a part of admin. Perhaps that's something to add to silverstripe/silverstripe-admin#610

robbieaverill: I see. Maybe the click handler needs a callback prop you can pass in e.g. `beforeHandleClick` which is given the event argument so we can tell it to stop propagation

client/src/components/ElementEditor/Header.js
raissanorth: Good point. Changed it as suggested.

ScopeyNZ: Also FYI - IDs that start with a number is invalid HTML!

raissanorth: 😮

client/src/components/ElementEditor/Element.js
```diff
@@ -20,7 +20,7 @@ class Element extends Component {
   /**
    * Take the user to the GridFieldDetailForm to edit the record
    */
-  handleClick() {
```
robbieaverill: We don't need this any more =)

raissanorth: Do you say that because this functionality is moving into the caret button in the header? And we don't want to replicate this if the user clicks anywhere on the element?!

ScopeyNZ: I think it's mentioning that you've added `event` but you're not using it anymore 🙂

raissanorth: ah, i removed that since.

Issue #266, opened 2018-07-19
API to define inline editing display


brynwhyman:
Blocks will be inline editable by default and will need to have an opt-out approach to in-line editing. Some blocks (eg. userforms) will not allow editing in-line as they are too complex.

This should probably be done as a `private static` declaration on block types, default is "in-line editable". Docs must be updated for block authors and some UI elements will be changed based on this setting.

A/Cs:
- [x] Opt-out configuration at block level for whether this block type is in-line editable
- [x] Docs updated to introduce this setting, and mention that it will not support GridField for now.
- [x] Blocks that are not marked as "inline-editable" will:
  - [x] Not have an options action dropdown
  - [x] Not have expand/collapse functionality
  - [x] Have the expand chevron pointed to the right instead
  - [x] Direct you to a `GridFieldDetailForm` when clicked

Options dropdown will be implemented in #322
Expand collapse functionality in #295 (probably)
GridDetailForm on click is already implemented

robbieaverill:
@chillu would be good to get your thoughts on this: if we're creating a new API like `getOverviewFields()` in order to determine the few form fields that are displayed when you initially load up inline editing of a block, where should the rest of the fields that `getCMSFields()` defines should be in `Root.Main` go?

If we add `Content` as a tab in the more actions dropdown (alongside Settings and History) then we'd be displaying the fields that are common between `getCMSFields()` and `getOverviewFields()` twice. If that's OK, then that's fine - what do you think?

The other question would be around us moving away from using GridField to power this (#273). Initially we won't have inline editing, so a CMS user would still click on an element in the editor view and be taken to an edit screen - I assume for simplicity that this would continue to be a GridFieldDetailForm? If that's the case, do we need to continue to use GridField as a backing for the ElementEditor, or can we still use parts of the GridField architecture without GridField being the entry point?

Another example would be the userforms block, which as per initial designs will not initially have inline edit capability even when inline editing is implemented. You'd set a message like "Click here to edit this element" or something similar, and be taken through to a GridFieldDetailForm to edit the data.

If we want to replace the GridFieldDetailForm implementation, perhaps we need to create an `admin/blocks` area which can be linked through to in order to edit content blocks...

Your thoughts would be great.

chillu:
> I assume for simplicity that this would continue to be a GridFieldDetailForm

[...]

PR #346, opened 2018-08-23
NEW In-line edit is an opt-out configuration setting

ScopeyNZ:
Partially addresses #266 (First two points). Once #345 is merged there's a little bit of follow up for that.


src/Models/BaseElement.php
```diff
@@ -128,6 +130,14 @@ class BaseElement extends DataObject
      */
     private static $disable_pretty_anchor_name = false;
 
+    /**
+     * Set to false to prevent an in-line edit form from showing in an elemental area. Instead the element will be
+     * clickable and a GridFieldDetailForm will be used.
+     *
+     * @var bool
```
robbieaverill: Add `@config`- at some stage we'll be using this annotation for the API docs site

PR #353, opened 2018-08-27
NEW Expand block preview

raissanorth:
Resolves https://github.com/silverstripe/silverstripe-elemental-blocks/issues/43

Block is not expandable:

screen shot 2018-08-27 at 3 56 44 pm

Block is expandable and expanded; the preview is hidden

screen shot 2018-08-27 at 3 53 34 pm

Block is expandable and _not_ expanded; the preview is shown
screen shot 2018-08-27 at 3 54 05 pm

Changes in this PR...

- add a `FormBuilder` component
- add a `Summary` component
- both of the above are injected into `Content` and rendered by it
- move the prior `Content` unit tests into the `Summary` test file, since the latter is handling this logic now
- add new unit tests for the `Content` component
- add unit tests for `Element` component

Design review by @clarkepaul still outstanding. 

The expand button has been tested on Edge 17, IE 11 (browser stack), Chrome 11, Firefox 61, and Safari 11 . It is keyboard accessible on all of these browsers, expect Safari. 

Also, the styling is slightly off on Edge 17:

image


client/src/components/ElementEditor/Content.js
```diff
@@ -1,37 +1,55 @@
...
+import { inject } from 'lib/Injector';
 
 class Content extends PureComponent {
   render() {
-    const { fileUrl, fileTitle, content } = this.props;
+    const {
+      fileUrl,
+      fileTitle,
+      content,
+      previewExpanded,
```
robbieaverill: It feels like this should be a context argument

E.g. `if (context === 'preview') { return <SummaryComponent ...props />; } else { return <FormBuilderComponent ...props />}`

client/src/components/ElementEditor/Element.js
```diff
@@ -13,17 +13,32 @@ class Element extends Component {
...
   /**
-   * Take the user to the GridFieldDetailForm to edit the record
+   * Expand the element to show the  preview
+   * If the element is not inline-editable, take user to the GridFieldDetailForm to edit the record
    */
-  handleClick() {
-    const { link } = this.props;
+  handleExpand() {
+    const {
+      element: { InlineEditable },
```
robbieaverill: maybe just import the `element` and use that` - the fact that the InlineEditable prop is capitalised is misleading, because usually capitalisation in React is used for component names, this sort of implies that it could be renderable

client/src/components/ElementEditor/Element.js
```diff
@@ -66,11 +86,16 @@ class Element extends Component {
           title={Title}
           elementType={BlockSchema.type}
           fontIcon={BlockSchema.iconClass}
+          link={link}
+          caretClickCallback={(event) => this.handleExpand(event)}
```
robbieaverill: `handleExpand` is already bound to `this` in the constructor - I think you could use `caretClickCallback={this.handleExpand}` here

client/src/components/ElementEditor/Header.js
```diff
@@ -39,6 +40,20 @@ class Header extends Component {
     }
   }
 
+  /**
+  * Handle the opening/closing of the block preview
+  */
```
robbieaverill: Indentation slightly out

client/src/components/ElementEditor/Header.js
```diff
@@ -52,12 +67,23 @@ class Header extends Component {
...
     const deleteButtonClassNames = classNames('dropdown-item', 'element-editor__actions-delete');
 
+    const expandTitle = i18n._t('ElementHeader.EXPAND', 'Show editable fields');
+    const expandButtonClassNames = classNames(
+      'dropdown-item',
+      'element-editor__expand',
+      'btn',
+      { 'font-icon-right-open-big': !expandable,
+        'font-icon-up-open-big': expandable && previewExpanded,
+       'font-icon-down-open-big': expandable && !previewExpanded });
```
robbieaverill: Can you please format this object a little - each entry on a new line with consistent indentation

raissanorth:
Note to self, rewrite the Behat tests!

client/src/components/ElementEditor/Content.js
robbieaverill: I see - this is loosely coupled to the container's state name... leave it as is +D

client/src/components/ElementEditor/Element.js
raissanorth: Cool, I'll go for your last suggestion. But for consistency I'll do the same with ID, Title, and BlockSchema.

robbieaverill: Yes please, thanks!

PR #357, opened 2018-08-27
FIX Resolving various console warning

ScopeyNZ:
As titled. There's a few React warnings that come up in the console when using this module (with dev built JS) - This resolves all of those that this module is responsible for.s


ScopeyNZ:
Currently GraphQL gives IDs as strings. I've converted the expectation in this module so that IDs are strings. I think GraphQL would need changes if we wanted to keep them as integers - but I may be missing something.

robbieaverill:
Needs a rebase now, soz

ScopeyNZ:
Done 🙂

PR #360, opened 2018-08-28
NEW Elements can be published via the inline publish button

### LINKED ISSUE #297: Add "more actions" item to publish block individually
Subtask from https://github.com/dnadesign/silverstripe-elemental/issues/290

* "Publish" option is in more actions menu
* GraphQL mutation given the element ID publishes the latest draft version from stage to live (versioned module contains these GraphQL operations)
* Apollo cache updates to update the state of the Element (ideally) or fallback to allowing the "read all blocks" query to re-run and reload all blocks

raissanorth:
Changes in this PR:

- modify the `ReadBlocksForPage` query to pull in `Version` and `IsPublished` (BaseElement casts the returned String to a Boolean)

- `IsPublished` is passed to the `Header` component and used to conditionally show the Publish button

- `Version` is passed to the `Header` component which feeds it into the new `publishBlockMutation`, making use of `copyBlockToStage`

The dropdown menu with the delete and publish options looks as follows:

image

Resolves #297


client/src/components/ElementEditor/Header.js
```diff
@@ -40,6 +42,26 @@ class Header extends Component {
...
+
+    const { id, version, actions: { handlePublishBlock } } = this.props;
+
+    const publishMessage = i18n._t(
+      'ElementHeader.CONFIRM_PUBLISH',
+      'Are you sure you want to publish this block?'
+    );
+
+    // eslint-disable-next-line no-alert
+    if (handlePublishBlock && confirm(publishMessage)) {
```
ScopeyNZ: I feel like we don't normally confirm publishes. I think we can drop the confirm here. /cc @clarkepaul

ScopeyNZ:
Nice work! Smashing out the PRs.

client/src/components/ElementEditor/Header.js
raissanorth: Good point! Let's see what @clarkepaul says.

raissanorth: @ScopeyNZ Thanks for spotting this. I checked with @clarkepaul and we do not need confirmation since we haven't confirmed publishing elements before. I will take this bit out.

client/src/components/ElementEditor/Header.js
```diff
@@ -66,31 +88,43 @@ class Header extends Component {
...
+    const { id, expandable, ActionMenuComponent, isPublished } = this.props;
 
     // Don't show the menu when inline editing is not enabled
     if (!expandable) {
       return null;
     }
 
     const deleteTitle = i18n._t('ElementHeader.DELETE', 'Delete');
-    const deleteButtonClassNames = classNames('dropdown-item', 'element-editor__actions-delete');
+    const publishTitle = i18n._t('ElementHeader.PUBLISH', 'Publish');
```
robbieaverill: I'm not convinced these need to be in variables since we aren't re-using them

client/src/components/ElementEditor/Header.js
```diff
@@ -66,31 +88,43 @@ class Header extends Component {
...
           onClick={this.handleDelete}
           title={deleteTitle}
           type="button"
-          className={deleteButtonClassNames}
+          className={classNames('element-editor__actions-delete', actionItemButtonClassName)}
         >
           {deleteTitle}
         </button>
+
+        {!isPublished &&
```
robbieaverill: How about we move this button into its own method - `renderPublishButton()` - it can return null if `isPublished` is true

src/Models/BaseElement.php
```diff
@@ -78,6 +78,7 @@ class BaseElement extends DataObject
     private static $casting = [
         'BlockSchema' => DBObjectType::class,
         'InlineEditable' => DBBoolean::class,
+        'IsPublished' => DBBoolean::class
```
robbieaverill: Is this casting required? What value type comes out if we don't define it?

client/src/components/ElementEditor/Header.js
raissanorth: Cool, can easily change that.

client/src/components/ElementEditor/Header.js
raissanorth: Good idea.

client/src/components/ElementEditor/Header.js
raissanorth: When you say 'these' do you mean only the `deleteButtonClassNames` or the `publishTitle`/`deleteTitle` also? For the latter I reckoned it be better then calling the translate function twice.

robbieaverill: The button classes can stay I reckon, but the two titles don't need to be variables if they're only used once. If they are used more than once then they're fine in a variable

raissanorth: Used more than once so I'll keep them as they are.

client/src/state/editor/publishBlockMutation.js
```diff
@@ -0,0 +1,44 @@
+import { graphql } from 'react-apollo';
+import gql from 'graphql-tag';
+
+// GraphQL query for saving a specific block
+const mutation = gql`
+mutation PublishBlock($blockId:ID!, $fromStage:VersionedStage!, $toStage:VersionedStage!, $fromVersion:Int!) {
+  copyBlockToStage(Input: {
```
maxime-rainville, line 7: @raissanorth @ScopeyNZ @robbieaverill Sorry for tagging you people on this 2 years after the fact, but does anyone recall a reason why we're using `copyBlockToStage` mutation here instead of `publishBlock`?

`copyBlockToStage` doesn't fire the onBeforePublish/onAfterPublish hooks which makes it difficult for developers to extend elements. 

https://github.com/dnadesign/silverstripe-elemental/issues/779

ScopeyNZ: Yeah. We had to support a version of SS4 that didn't have that mutation scaffolder I think.

maxime-rainville: That make sense. That's probably out of support by now anyway.  I'll try to dig up when that got added.

PR #361, opened 2018-08-28
NEW Add box shadow to focussed and hovered elements in the editor

### LINKED ISSUE #270: UX to highlight selected block
## User Story

As a CMS user, I want an indication of which block preview is expected to be selected, so that I can confirm I’m focusing on the correct block preview before clicking it.

## Acceptance Criteria

- [x] Mouse-hover and click ( :focus) of a content block gives shadow animation
- [x] If the “:focus” state is moved to a different block then the shadow is removed from the first block to the second block, as per wireframe: https://projects.invisionapp.com/share/ZRL4T2HQAGB#/screens/301920915 
- [x] The blocks are keyboard accessible

## Pull requests

- [ ] #361

robbieaverill:
Resolves #270


ScopeyNZ:
I know @newleeland mentioned "just use something existing" but the designs had a much more defined shadow on the top. I think this is something we should address now rather than later if the UX peeps are just going to want changes.

Maybe @newleeland can provide some feedback before merging?

robbieaverill:
Sure. We have five variations of box shadows in core already, if this needs a sixth we should add it there

PR #365, opened 2018-08-29
Unpublish an element via inline button

raissanorth:
Changes in this PR
- add a mutation to unpublish individual elements
- add a button to the dropdown menu that triggers the unpublishing of individual elements (this is conditionally rendered and only appears when the item is live or has been modified since the latest publish

- ~it fixes a bug that was introduced in b49e2230e877caa32cc48bd44e45d6df4320b5a2: rather than isPublished we need to retrieve isLiveVersion~ - In addition to `isPublished`, `isLiveVersion` was added; the former checks if there exists _a_ live version of this element (and remains true as long as there is at least one), whereas the latter checks if the current version is _the_ live version


client/src/components/ElementEditor/Header.js
```diff
@@ -58,6 +60,21 @@ class Header extends Component {
...
 
+  /**
+   * Handle unpublishing of a block, passing the block ID and version number in
+   */
+  handleUnpublish(event) {
+    event.stopPropagation();
+
+    const { id, version, actions: { handleUnpublishBlock } } = this.props;
+
+    // eslint-disable-next-line no-alert
```
robbieaverill: There's no alert any more =)

client/src/components/ElementEditor/Header.js
```diff
@@ -198,12 +239,13 @@ Header.propTypes = {
...
   version: PropTypes.number,
-  isPublished: PropTypes.bool,
+  isLiveVersion: PropTypes.bool,
   elementType: PropTypes.string,
   fontIcon: PropTypes.string,
   actions: PropTypes.shape({
     handleDeleteBlock: PropTypes.func.isRequired,
-    handlePublishBlock: PropTypes.func
+    handlePublishBlock: PropTypes.func,
+    handleUnpublishBlock: PropTypes.func
```
robbieaverill: If you add a trailing comma then the next PR won't show this line as changed (like this one has)

raissanorth:
@robbie I have checked and once an element is 'unpublished' the value of `IsPublished` is false, even if there have been prior published versions. `Unpublish` removes the latest live version, and `isPublished` checks if there are any live version - however none exist in the table.

I have added `isLiveVersion`. There can be cases where an element is published and then modified which causes `isLiveVersion` to be false, whereas `isPublished` is still true. We want to render the publish button to allow users copy this modified version to the live stage.

client/src/components/ElementEditor/Header.js
```diff
@@ -111,6 +124,31 @@ class Header extends Component {
...
+  /**
+   * Render the unpublish button if the current state of the button is published
+   *
+   * @returns {DOMElement|null}
+   */
+  renderUnpublishButton() {
+    const { isPublished } = this.props;
+    const unpublishTitle = i18n._t('ElementHeader.UNPUBLISH', 'Unpublish');
+
+    if (!isPublished) {
```
robbieaverill: This prop has been renamed to isLiveVersion - does this need updating too?

raissanorth: No, I don't think so. We won't to always give the CMS user the option to unpublish the publish version. By using `isPublished` here instead of `isLiveVersion` we ensure that they can also unpublish in the case that a version has been published and then modified without being published. In this case `isPublished` equals true, whereas `isLiveVersion` equals false.

client/src/state/editor/readBlocksForPageQuery.js
```diff
@@ -21,7 +21,11 @@ query ReadBlocksForPage($id:ID!) {
             Title
             BlockSchema
             InlineEditable
+            IsLiveVersion
+<<<<<<< Updated upstream
+=======
             IsPublished
+>>>>>>> Stashed changes
```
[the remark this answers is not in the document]
raissanorth: 🤦‍♀️

src/Models/BaseElement.php
```diff
@@ -78,7 +78,7 @@ class BaseElement extends DataObject
     private static $casting = [
         'BlockSchema' => DBObjectType::class,
         'InlineEditable' => DBBoolean::class,
-        'IsPublished' => DBBoolean::class
+        'IsLiveVersion' => DBBoolean::class
```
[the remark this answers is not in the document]
raissanorth: Indeed.

Issue #370, opened 2018-09-03
UX improvements for "Add Blocks" widget with many block types


ScopeyNZ:
Follow up from #338.

#338 has implemented the content of a popover that will be used to add blocks to a page. Currently all blocks will show in the popover. This will be an issue given project may have many block types.

We should at least implement a max-height for this widget and allow scrolling (or similar) when there are many blocks. 

Designs: https://projects.invisionapp.com/dsm/silver-stripe/silver-stripe/asset/components/5b90499d4d45900010a7be60

clarkepaul:
@ScopeyNZ yup @newleeland is on it.

robbieaverill:
I guess we also need to show a limited number of blocks to start with (9 maybe, if it's three columns?). Then the search feature can show more.

We probably need a way for devs to give priority to specific block types for their projects, e.g. where a "carousel block" is more used than a "content block" so they'd want to make sure it shows up by default rather than having to search for it each time.

We could do that with a numeric config flag which could be increased in YAML to give a higher priority, but I'm not sure how that would work with GraphQL queries which expect DB values, and it would also mean having to filter a list of blocks in PHP rather than in the ORM query which might have a small performance hit. We could also just return the list of block types in PHP with no ORM queries at all, then problem solved.

ScopeyNZ:
The long term goal sounds like it will allow CMS authors to categorise blocks. The short term "MVP" is to show a number in two columns (12 I think?) and have it scrollable. I think @newleeland is still finalising.

clarkepaul:
Added designs to top comment for further breakdown of ACs

raissanorth:
Maximum-height was added as part of https://github.com/dnadesign/silverstripe-elemental/pull/406. 

Before I close this ticket I would like to add an automated way to test this, i.e. making it easy to add at least a dozen different block types. 
I believe @chillu, you are working on a way to add elements to pages for testing purposes. Will this work include different types of blocks, you reckon?

chillu:
I'm only building a block generator which uses existing block types: https://github.com/silverstripe/silverstripe-frameworktest/issues/49. You're correct that you should check this with a number of blocks. You could either adapt https://github.com/silverstripe/silverstripe-frameworktest/blob/master/code/tasks/FTPageTypeCreatorTask.php, or just do it manually. But don't let this card be blocked by my (somewhat unrelated) block generator work

raissanorth:
Hi @newleeland,
[...]

PR #371, opened 2018-09-03
NEW Add content for 'Add Block' popover widget

raissanorth:
This is the ground work for adding a popover to add a new 'Add Block' widget for the Elemental Editor. There is yet some work remaining, as decisions from the Style Council are still outstanding - see #370

There are also styling issues with Edge and IE, but these can be addressed as part of the ongoing work related to this feature:
image

The buttons are keyboard accessible on the latest versions of Edge, IE, Firefox, Chrome, and Safari.

Note that this does not return a popover itself, but the content that could be rendered within a Popover component.


client/src/components/ElementEditor/AddElementPopoverContent.js
```diff
@@ -0,0 +1,70 @@
...
+ */
+class AddElementPopoverContent extends Component {
+  constructor(props) {
+    super(props);
+
+    this.renderElementButtons = this.renderElementButtons.bind(this);
+    this.handleLinkClick = this.handleLinkClick.bind(this);
+  }
+
+  handleLinkClick() {
```
ScopeyNZ: This was removed right? There's still a few references to this stuff. Sad to see it go 😢

client/src/components/ElementEditor/AddElementPopoverContent.js
```diff
@@ -0,0 +1,70 @@
...
+  elementTypes: PropTypes.arrayOf(PropTypes.shape({
+    title: PropTypes.string,
+    icon: PropTypes.string,
+  })),
+
+};
+
+AddElementPopoverContent.defaultProps = {
+  elementTypes: [
+    { title: 'banner', icon: 'font-icon-block-banner' },
```
ScopeyNZ: The default props should not have test data in it any more

client/src/components/ElementEditor/AddElementPopoverContent.js
raissanorth: Same here.

client/src/components/ElementEditor/AddElementPopoverContent.js
```diff
@@ -0,0 +1,58 @@
...
+
+    this.renderElementButtons = this.renderElementButtons.bind(this);
+  }
+
+  renderElementButtons() {
+    const { elementTypes } = this.props;
+
+    return elementTypes.map((elementType) =>
+      (
+        <Button className={
```
ScopeyNZ: You will need to add a `key=` in here. There is an identifier (of sorts) on the `elementType` you can use.

client/src/components/ElementEditor/AddElementPopoverContent.js
```diff
@@ -0,0 +1,58 @@
...
+        >
+          {elementType.title}
+        </Button>
+      )
+    );
+  }
+
+  render() {
+    return (
+      <div className="element-editor__add-element-popover-content">
```
ScopeyNZ: You're going to hate me for this one. I think we should refactor these class names. I'm terrible at this myself because [BEM confuses me](http://getbem.com/naming/). It's supposed to be like this:

```
.block__element--modifier
```

Modifiers should only be things like `--hidden` or `--disabled`

We should probably refactor this so `element-editor-add-element` is the "block" and `content` is the "element". Ie. `className="element-editor-add-element__content"`. And then for the next div `className="element-editor-add-element__button-container"`.

PR #373, opened 2018-09-04
NEW Show CMS field (top-level) tabs in more actions dropdown

### LINKED ISSUE #323: Add options for tabs within the more actions dropdown
The designs indicate that the tab options that usually appear in a `GridFieldDetailForm` for blocks will need to be available in a dropdown on blocks. These need to be available per block without considerable performance degradation.

To handle this, I propose that we check tabs per block type (extension of `BaseElement`) and cache the result. There's is a possibility that the data saved on a block may affect the tabs that are shown, but when a block is expanded we can update the tab options dynamically (logged at #324).

A/Cs:
- [x] Tabs are calculated per block type and cached so that performance is not significantly affected
- [x] All tabs that usually show when editing a saved block with a GridFieldDetailForm are included
- [x] Tab options are included in the action menu on a block (implemented by #322)

PR: #373

ScopeyNZ:
This introduces tab names to the "more actions" dropdown on individual elements in the elemental editor. This includes:
- Server-side caching of tab names per element type
- A new type for "element type" - If there's suggestions for better naming I'm all ears

Resolves #323


client/src/components/ElementEditor/ElementList.js
```diff
@@ -10,16 +11,18 @@ class ElementList extends Component {
...
     if (!blocks) {
       return null;
     }
 
+
     return blocks.map((element) => (
       <ElementComponent
         key={element.ID}
         element={element}
+        editTabs={elementTypes.find(type => element.BlockSchema.type === type.title).tabs || []}
```
robbieaverill: I feel like this deserves to be its own method

client/src/components/ElementEditor/Header.js
```diff
@@ -185,6 +187,16 @@ class Header extends Component {
...
   }
 
+  renderEditTabs() {
+    const { editTabs } = this.props;
+
+    if (!editTabs || !editTabs.length) {
+      return null;
+    }
+
+    return editTabs.map((tab) => <button key={tab} className="dropdown-item">{tab}</button>);
```
robbieaverill: Do we want to add extra class options to this now, incase eg. someone wants to put icons into it in their site?

robbieaverill: Cancel, that'd mean passing an array of objects instead of strings. Ignore

src/Forms/ElementalAreaField.php
```diff
@@ -115,11 +116,15 @@ public function getSchemaDataDefaults()
 
         $blockTypes = [];
 
+        /** @var ElementTabProvider $tabProvider */
+        $tabProvider = Injector::inst()->get(ElementTabProvider::class);
```
robbieaverill: You could use injector to provide this as a service and add the relevant getter/setter for it

src/Services/ElementTabProvider.php
```diff
@@ -0,0 +1,123 @@
...
+use DNADesign\Elemental\Models\BaseElement;
+use Psr\SimpleCache\CacheInterface;
+use SilverStripe\Core\ClassInfo;
+use SilverStripe\Core\Config\Configurable;
+use SilverStripe\Core\Flushable;
+use SilverStripe\Core\Injector\Injector;
+use SilverStripe\Forms\TabSet;
+
+/**
+ * Provides tab names for
```
robbieaverill: ...?

src/Services/ElementTabProvider.php
```diff
@@ -0,0 +1,123 @@
...
+     * ElementTabProvider constructor.
+     * @param CacheInterface $cache
+     */
+    public function __construct(CacheInterface $cache = null)
+    {
+        if (!$cache) {
+            $cache = Injector::inst()->get(CacheInterface::class . '.ElementTabCache');
+        }
+        $this->cache = $cache;
+    }
```
robbieaverill: I do prefer this kind of proper DI, but we should use Injector to handle this instead. You'll need to add the config for it as well as the getter/setter for `$this->cache`

src/Services/ElementTabProvider.php
```diff
@@ -0,0 +1,123 @@
...
+
+    /**
+     * Get the top level tab names for the given element class
+     *
+     * @param string $elementClass
+     * @return array Array of the tabs for the element
+     */
+    public function getTabsForElement($elementClass)
+    {
+        if (null !== ($tabs = $this->cache->get($this->getCacheKey($elementClass)))) {
```
robbieaverill: When you've added a getter, use `$this->getCache()`

src/Services/ElementTabProvider.php
```diff
@@ -0,0 +1,123 @@
...
+     * @see FlushMiddleware
+     */
+    public static function flush()
+    {
+        /** @var static $self */
+        $self = singleton(static::class);
+
+        $self->cache->clear();
+
+        if (self::config()->regenerate_on_flush) {
```
robbieaverill: It might be preference, but I'd rather see it written like this:

```diff
- self::config()->regenerate_on_flush
+ $this->config()->get('regenerate_on_flush')
```

src/Services/ElementTabProvider.php
```diff
@@ -0,0 +1,123 @@
...
+
+    /**
+     * Generate top level tab names for the given element class (and cache them)
+     *
+     * @param string $elementClass
+     * @return array
+     */
+    protected function generateTabsForElement($elementClass)
+    {
+        // We set an ID because some tabs only show if ->isInDb() is true
```
robbieaverill: Be careful: https://github.com/silverstripe/silverstripe-framework/pull/7799

src/Services/ElementTabProvider.php
ScopeyNZ: Okay I'll update. My preference is still to use `self` I'd say. `config()` is marked as static and `self` is more descriptive about where the config is coming from.

src/Services/ElementTabProvider.php
ScopeyNZ: Hmmmn. This is just to get around `link_tracking` and `file_tracking` - which should hopefully not show by default soon. Might be able to remove it. Additionally it looks like that PR includes a way to "fake it" so we might be able to update this.

src/Services/ElementTabProvider.php
ScopeyNZ: Updated - It really doesn't sit well with me that I'm removing this constructor but I understand it's the Injector way...

src/Services/ElementTabProvider.php
ScopeyNZ: Also in this case it's a static method - so I updated it to `$self->config()->get()`.

ScopeyNZ:
Updated from feedback 🙂

src/Forms/ElementalAreaField.php
robbieaverill: Ok, that's fine then - just add a comment

src/Services/ElementTabProvider.php
robbieaverill: I've removed the tabs in #374 because they break the ability set an element Title

client/src/components/ElementEditor/Header.js
```diff
@@ -185,6 +187,16 @@ class Header extends Component {
     );
   }
 
+  renderEditTabs() {
```
robbieaverill: Please add a doc block

ScopeyNZ:
Updated again and rebased.

_config/config.yml
```diff
@@ -25,3 +25,10 @@ Symbiote\GridFieldExtensions\GridFieldAddNewMultiClassHandler:
 SilverStripe\Core\Injector\Injector:
   SilverStripe\CMS\Controllers\CMSSiteTreeFilter_Search:
     class: DNADesign\Elemental\Controllers\ElementSiteTreeFilterSearch
+  Psr\SimpleCache\CacheInterface.ElementTabCache:
+    factory: SilverStripe\Core\Cache\CacheFactory
+    constructor:
+      namespace: "ElementTabCache"
+  DNADesign\Elemental\Services\ElementTabProvider:
+    properties:
+      cache: %$Psr\SimpleCache\CacheInterface.ElementTabCache
```
robbieaverill: I think it's always safest to quote injector definitions like this, I've had problems without quoting them before. Also we could put these config definitions into a new cache.yml file

src/Services/ElementTabProvider.php
robbieaverill: Ah sorry I missed that. That's fine. I do think it's a bit nicer this way since the config is obviously part of the same instance, even though in practice it wouldn't matter. Cheers

PR #374, opened 2018-09-05
ENHANCEMENT Show state of elements draft/live/modified

raissanorth:
Styling based on current implementation:

image

Note that the CSS changes minimally increase the distance between the header title and the summary content.


raissanorth:
Not sure what is causing the Behat test failures. Could be a timing issue. That particular test runs fine locally:

image

raissanorth:
Added a box-shadow to the circle icon. Unfortunately it is a lot more subtle than the one in the designs, since I tried to reuse existing variables for consistency:

image

tests/Behat/features/edit-block-element.feature
```diff
@@ -32,5 +32,7 @@ Feature: Edit elements in the CMS
       And I press the "Publish" button
     Then I should see a "Published content block" message
     When I go to "/admin/pages/edit/show/6"
-      Then I should see "Eve's Block"
+      And I wait until I see the ".element-editor__element" element
```
robbieaverill: I think I mentioned this on another PR - this assertion isn't designed to select a class - you're basically saying "wait until I see the [any element] element" where it's designed to be more specific than that e.g. "wait until I see the [specific] element".

We could add another step like "wait until the elements have loaded" instead maybe?

robbieaverill:
By the way, I realise re: the class names that you're re-using existing styles from the PHP GridField equivalent, but I think we should migrate them to the React structure and update the legacy reports GridField implementation to use those instead

client/src/components/ElementEditor/Header.scss
```diff
@@ -40,4 +40,30 @@
...
+  height: 8px;
+  left: 22px;
+  position: relative;
+  width: 8px;
+  z-index: 1;
+}
+
+.element-editor__element {
+  &--draft {
+    .element-editor-header__version-state {
```
ScopeyNZ: Isn't this not BEM?

robbieaverill: Yeah not really. We could compromise and have the versioned state on both parts? I do think it’s important to have it on the element too though

ScopeyNZ: Yes I think that would be an appropriate compromise. I can feel it's going to be a new pet peeve of mine if we use BEM but make styles with specificity >`0 1 0`

PR #375, opened 2018-09-06
FIX Updating styling to more accurately match given designs

ScopeyNZ:
I sat down with @clarkepaul and we went over the designs in detail. Here's a whole bunch of styling changes which makes everything follow the designs a little more accurately.

Pic:

image

We also decided to make the caret not a button, and leave that role to the element which already has the role of the button. This means that the caret has been removed from the tab index - The title already provides the necessary screen reader prompts.


robbieaverill:
+1 for the branch name

client/src/components/ElementEditor/Header.scss
```diff
@@ -4,9 +4,13 @@
   justify-content: space-between;
 
   &__title {
+    font-size: 15px;
```
robbieaverill: It's a shame this has to be 15. We have `$h4-font-size` (14px) and `$font-size-lg` (1.231rem / 16px)

robbieaverill:
Technically a major change since removing the handle caret click method, but fine because we're still breaking stuff!

robbieaverill:
Actually, I'm gonna add a tiny bit of margin on the left of the caret:

image

Shoot me if you want to @clarkepaul!

robbieaverill:
Now looks like this:

image

Issue #368, opened 2018-09-03
ElementalArea gridfield performance issues


mfendeksilverstripe:
Version: `^3.0.0 (5334f5709878a3502e1cee431c58a28c36f9a843)`

It seems that elemental area gridfield does not scale well with the number of blocks contained inside it. I tested this on a page with 56 blocks. Here are the results:

- `28.4s` CMS page load with `ElementalArea` field removed
- `51.6s` CMS page load with `ElementalArea` field removed

That's quite a difference. It's probably worth investigating why is the difference so large.

robbieaverill:
For reference, I'm testing this (in CWP 2.1) on a page with 100 blocks in one ElementalArea.

Response times: 11.2s, 11.5s, 11.18s

It's a long way off what you're reporting, but that could be because you're using container blocks like layout blocks/the list block. 

It's still too slow.

robbieaverill:
Ok so 50% of the request time when loading the CMS page edit form is calling `ElementalArea::getOwnerPage()` - I've made a PR at #376 to use a cacheable call to do this which reduces that by 50% - a good start.

robbieaverill:
Ok I've tried a few other things:

* Partial caching on the editor preview's icon
* Internal instance caching for:
  * `BaseElement::getPage()`
  * `ElementalArea::getOwnerPage()`
  * `BaseElement::getAreaRelationName()`
  * `BaseElement::CMSEditLink()`

None of these have made a very significant change in response time at all.

The bulk of the remaining time is in `BaseElement::getEditorPreview()`, which generates the GridField UI for each element in the CMS (icon, title, type, summary). All of this is unique to each element, and aside from trying to cache the icon I don't think we can do much more with it.

@mfendeksilverstripe thoughts?

mfendeksilverstripe:
@robbieaverill Tried to test it out. I upgraded to `3.0.x-dev` (`0e3c17aa866b7d954fdd369f9c5681a4e95956a3`), however I couldn't get the page loaded in CMS because of this issue:

```
[Emergency] Uncaught SilverStripe\ORM\Connect\DatabaseException: Couldn't run query:

SELECT DISTINCT "SiteTree"."ClassName", CASE WHEN "SiteTree_Localised_en_NZ"."ID" IS NOT NULL THEN
"SiteTree_Localised_en_NZ"."LastEdited" ELSE "SiteTree"."LastEdited" END AS "LastEdited", CASE WHEN
[...]

PR #376, opened 2018-09-07
FIX Use cachable call to get page version from draft stage

robbieaverill:
Profiling suggests that 50% of the request calls are spent getting the owner page. This change allows Versioned to cache those calls, reducing the CMS page edit form load time by ~50%

This also removes around ~300ms with 100 blocks by internally caching some of the most used ORM calls.

Issue: #368

### DIFF
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -92,6 +92,14 @@ class BaseElement extends DataObject
      */
     protected $controller;
 
+    /**
+     * Cache various data to improve CMS load time
+     *
+     * @internal
+     * @var array
+     */
+    protected $cacheData;
+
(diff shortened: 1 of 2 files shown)


chillu:
@unclecheese 300ms on 100 blocks doesn't sound like it's delivering the kind of improvements that large users are after. How does this relate to the 12-18s which the original ticket reported?

robbieaverill:
There were two fixes here - one was to use a cacheable call to get the DataObject which shaved off ~50% load time. The second was internal instance caching (~300ms reduction) to reduce load in repeated calls to getSomething(), we implemented both then reverted the first one because it wasn't fluent compliant, but it should still be covered by the second one.

PR #377, opened 2018-09-10
FIX Ensure Versioned query has table name prefix so it will work with modules like fluent

robbieaverill:
Issue: #368


robbieaverill:
Bah, this just exposes more of the mess with test stubs. I'll change it back - we did internal class caching in #376 anyway so it's less relevant

Issue #285, opened 2018-08-02
Implement React based drag and drop for reordering elements in the editor


robbieaverill:
In elemental <= 3.x the GridField sortable component is used to allow reordering elements.

In the React equivalent editor (not using GridField components) we'll need to add a React component to allow this.

### Acceptance Criteria
- [x] Styling as per one of design directions (pending decision): https://projects.invisionapp.com/dsm/silver-stripe/silver-stripe/asset/components/5b8c6cbfc5ffe300176d8e22
- [x] Blocks can be dragged up or down the hierarchy of the blocks on a page, both collapsed or expanded blocks can be dragged. 
- [x] On hover content blocks have a ‘drag handle’ icon that is visible for mouse operated UI’s which can be used to click and drag, as well as showing an icon on hover, there is a shadow below the block as per designs.
- [x] Dragging can be activated anywhere on the block where there isn’t another action and the block is pressed for x time (use smart defaults from drag-drop library). If dragging is in process (On “click-and-hold” or “touch-and-hold”) of a block, the drag handle and shadow beneath the block appears. 
- [x] I receive an indication from the interface as to where I can drop the block. Highlighted by an indicator line
- [ ] Re-ordered blocks will become “modified” if previously published, the block will require publishing for the effects of re-ordering to be visible on the live site.

### Options:

* https://github.com/react-dnd/react-dnd: already used in silverstripe/asset-admin
* https://github.com/atlassian/react-beautiful-dnd

The handling of the reordering could be done with a GraphQL mutation on ElementalArea, providing the list of element IDs in the desired new sort order after a drop is completed. This could be a GraphQL type like `ElementIDList`, but might be a little unstructured in terms of what GraphQL expects.

Another option might be to put the mutation directly into the element that is dragged and dropped, providing a mutation signature like this:

```graphql
mutation updateElementSortOrder($id:ID!, $before:ID, $after:ID) {
  setElementSortOrder(Input: {
    ID: $id
    MoveAfter: $after
    MoveBefore: $before
  }) {
    ID
  }
}
```

In this case the element would need to set its sort order to +1 or -1 of the MoveAfter/MoveBefore element, then set new sort orders for all elements in the list to respect the updates.

Just possible ideas at this stage.

brynwhyman:
@newleeland can you please confirm this is the correct wireframe? There's a similar DnD wireframe not included here

brynwhyman:
Refer to asset-admin DnD styling for this

[...]

PR #380, opened 2018-09-11
NEW Add custom mutation to allow changing the sort position of indivi…

raissanorth:
…dual elements

The changes in this PR add a custom mutation to allow sorting of individual elements without having to  fire multiple queries. This is based on the second suggestion in #285.

Thanks @ScopeyNZ for your help!


src/GraphQL/SortBlockMutationCreator.php
```diff
@@ -0,0 +1,93 @@
...
+            if (!$afterBlock) {
+                throw new \InvalidArgumentException(sprintf(
+                    '%s#%s not found',
+                    BaseElement::class,
+                    $afterBlockID
+                ));
+            }
+            if ($afterBlock->ParentID !== $parentId) {
+                throw new \InvalidArgumentException(sprintf(
+                    '%s# not found',
```
ScopeyNZ: In this case the message should be something that indicates "Trying to sort block after a block in a different elemental area" or words to that effect.

src/GraphQL/SortBlockMutationCreator.php
```diff
@@ -0,0 +1,93 @@
...
+            $operator = '-';
+            $filter = "Sort <= $sortAfterPosition && Sort > $blockPosition";
+            $newBlockPosition = $sortAfterPosition;
+        }
+
+        DB::query("
+            UPDATE Element
+            SET Sort = (Sort $operator 1)
+            WHERE $filter && ParentId = $parentId 
+        ");
```
robbieaverill: I don't agree with this approach for two reasons:

* We should be avoiding direct SQL queries, and at worst should be using SQLUpdate with quoted table and field names so it will work with DB engines other than MySQL
* Updating the sort order of other elements in the same parent will mean they don't have new versions created for them

I think we need to use the ORM for this

src/GraphQL/SortBlockMutationCreator.php
```diff
@@ -0,0 +1,93 @@
...
+            'ID' => ['type' => Type::nonNull(Type::id())],
+            'AfterBlockID' => ['type' => Type::nonNull(Type::id())],
+        ];
+    }
+
+    public function resolve($object, array $args, $context, ResolveInfo $info)
+    {
+        $block = BaseElement::get_by_id($args['ID']);
+
+        if (!$block) {
```
robbieaverill: You also need to check for canEdit() with the `$context['currentUser']` ([docs](https://github.com/silverstripe/silverstripe-graphql#authentication))

src/GraphQL/SortBlockMutationCreator.php
robbieaverill: Here's the relevant code in the CMS that does this with the ORM for pages: https://github.com/silverstripe/silverstripe-cms/blob/4/code/Controllers/CMSMain.php#L817-L826

robbieaverill: Actually that's a bad example because it's doing exactly what you're doing here, but deliberately:

```php
                    // Nodes that weren't "actually moved" shouldn't be registered as
                    // having been edited; do a direct SQL update instead
```

We need to do the opposite for elements because changing the sort order is a versionable operation

ScopeyNZ: I disagree. If you have ~60 blocks and you move the last one to the beginning I don't think anyone would expect a new version on every single block. If any should have a new version, just the block that was reordered. 

Frankly, versioning sort is something that probably needs to be addressed separately. When you revert to a version with an old sort, the sort on other blocks will need to be changed.

We're doing this deliberately. Running 60 queries for updating sort is ridiculous when you can run two. Refactoring it to use `DBUpdate` is fair.

ScopeyNZ:
@raissanorth I added a commit adding optimistic cache updates for this mutation.

client/src/state/editor/sortBlockMutation.js
```diff
@@ -20,6 +21,42 @@ const config = {
...
+          // Add it back after the target
+          const end = edges.slice(targetBlockIndex + 1);
+          edges = edges.slice(0, targetBlockIndex + 1);
+          edges.push(movedBlock);
+          edges = edges.concat(end);
+        }
+
+        // Add it back to the full result
+        newData.readOnePage.ElementalAreaIfExists.Elements.edges = edges;
+        store.writeQuery({ query: readBlocksQuery, data: newData, variables });
```
robbieaverill: Cool 😄

client/src/state/editor/sortBlockMutation.js
```diff
@@ -20,6 +21,42 @@ const config = {
         blockId,
         afterBlockId,
       },
+      optimisticResponse: {
+        ID: blockId,
+        __typename: 'Block',
+      },
+      update: store => {
+        const variables = readBlocksConfig.options({ pageId: 6 }).variables; // Fix me
```
ScopeyNZ: I need to finish this 🤦‍♂️

src/GraphQL/SortBlockMutationCreator.php
robbieaverill: >When you revert to a version with an old sort, the sort on other blocks will need to be changed.

That's a good point

client/src/state/editor/sortBlockMutation.js
ScopeyNZ: I finished the implementation of this. Now the function takes `pageId` as a parameter (because it needs to know what `readBlocksForPage` query to update). The params are different from the query but that's fine I assume.

ScopeyNZ:
Hey @raissanorth I added a test for the new mutation. I ran this test with Fluent so it seems fine. In any case, the fields and filters we use shouldn't be translated by Fluent so I'm happy now. 

Feel free to look over my commit and offer any feedback. Ready for a re-review @robbieaverill .

robbieaverill:
>the fields and filters we use shouldn't be translated by Fluent so I'm happy now

I think it's pretty likely that someone will want to localise their sort order for a page between different locales

src/GraphQL/SortBlockMutationCreator.php
```diff
@@ -0,0 +1,102 @@
...
+            $operator = '-';
+            $filter = "\"Sort\" <= $sortAfterPosition AND \"Sort\" > $blockPosition";
+            $newBlockPosition = $sortAfterPosition;
+        }
+
+        $table = DataObject::getSchema()->tableName(BaseElement::class);
+
+        $query = SQLUpdate::create()
+           ->setTable("\"$table\"")
+           ->assignSQL('"Sort"', "\"Sort\" $operator 1")
```
robbieaverill: Just checking - did you test this with fluent localising the Sort field? If it breaks it'll be here and because there's no table name in the column e.g. `"Foo"."Field"` then it'll throw ambiguity errors

src/GraphQL/SortBlockMutationCreator.php
```diff
@@ -0,0 +1,102 @@
...
+           ->addWhere([$filter, '"ParentID"' => $parentId]);
```
robbieaverill: We need to also add the table name to ParentID here to avoid ambiguity

src/GraphQL/SortBlockMutationCreator.php
```diff
@@ -0,0 +1,102 @@
...
+
+        $query->execute();
+
+        $block->Sort = $newBlockPosition;
+        $block->write();
```
robbieaverill: This class has zero documentation about why we've made the decisions we've made about how sorting should be done - could you please add some? For example, I know we're using manual SQL updates because we don't want every element to have a new version when reordered (even though I disagree with this decision)

tests/GraphQL/SortBlockMutationCreatorTest.php
```diff
@@ -0,0 +1,60 @@
...
+
+    protected static $extra_dataobjects = [
+        TestElement::class
+    ];
+
+    protected function setUp()
+    {
+        parent::setUp();
+
+        $this->logInWithPermission('ADMIN');
```
robbieaverill: It should be logged in as admin by default when you have a fixture file - you can remove this

tests/GraphQL/SortBlockMutationCreatorTest.php
```diff
@@ -0,0 +1,60 @@
...
+        $this->assertIdsInOrder([2,3,1,4]);
+
+        $this->runMutation(4, 2);
+        $this->assertIdsInOrder([2,4,3,1]);
+
+        $this->runMutation(1, 0);
+        $this->assertIdsInOrder([1,2,4,3]);
+
+        $this->runMutation(3, 2);
+        $this->assertIdsInOrder([1,2,3,4]);
```
robbieaverill: This could be a data provider, it's four separate unit tests

robbieaverill: I see you've done this to continually reorder them in one go, I'll leave it

robbieaverill:
I've pushed an update commit which addresses all my concerns and also does some simple SQL injection prevention checks that I know will otherwise come up at some point in a review =)

PR #383, opened 2018-09-12
FIX Re-enable Behat tests by adding to FeatureContext

### LINKED ISSUE #382: Complete behat test for inline editing
#372 as a part of #327 involves adding an inline edit form for elements to allow editing directly in the `ElementalArea`. This had accompanying end to end tests (behat), one of which proved a little too weak to be included in the merge.

This needs to be fixed with a more robust solution and re-enabled (currently it is commented).

https://github.com/dnadesign/silverstripe-elemental/blob/f25de3e5adf3df474cd7aadef420ab4680313070/tests/Behat/features/element-editor.feature#L25-L37

raissanorth:
Resolves #382 

This PR adds three custom functions to the feature context to allow the tests to assert that the edit form exists and/or is visible to the CMS user. As a pleasant side effect the custom functions should make the tests more readable as well.

### DIFF
--- a/tests/Behat/Context/FeatureContext.php
+++ b/tests/Behat/Context/FeatureContext.php
@@ -7,10 +7,11 @@ use SilverStripe\BehatExtension\Context\SilverStripeContext;
 if (!class_exists(SilverStripeContext::class)) {
     return;
 }
+
 class FeatureContext extends SilverStripeContext
 {
     /**
-     * @Then /^I should( not |\s*)see the edit form for block (\d+)$/i
+     * @Then /^I should( not |\s+)see the edit form for block (\d+)$/i
      */
     public function iShouldSeeTheEditFormForBlock($negative, $position)
     {
@@ -21,10 +22,10 @@ class FeatureContext extends SilverStripeContext
         $form = $block->find('css', '.element-editor-editform');
 
         if ($iShouldNotSee) {
-            assert(!$form || !$form->isVisible(), 'I see the form! Try again later.');
+            assertTrue(!$form || !$form->isVisible(), 'I see the form! Try again later.');
         } else {
             assertNotNull($form, 'Edit form not found');
-            assert($form->isVisible());
+            assertTrue($form->isVisible());
         }
     }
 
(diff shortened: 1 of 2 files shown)


tests/Behat/Context/FeatureContext.php
```diff
@@ -9,6 +9,25 @@
 }
 class FeatureContext extends SilverStripeContext
 {
+    /**
+     * @Then /^I should( not |\s*)see the edit form for block (\d+)$/i
```
robbieaverill: Super minor, but I think ` not |\s*)` should be `\s+` otherwise it might match `I shouldsee the` when it should only match `I should see`

tests/Behat/Context/FeatureContext.php
```diff
@@ -142,4 +174,18 @@ protected function getArchiveButton($position)
 
         return $button;
     }
+
+    /**
+     * Returns the caret button for a specific block
+     *
+     * @param $block
```
robbieaverill: Needs a type

tests/Behat/Context/FeatureContext.php
```diff
@@ -35,6 +54,19 @@ public function iClickOnBlock($position)
...
+    public function iClickOnTheCaretButtonForBlock($position)
+    {
+        $block = $this->getSpecificBlock($position);
+        $button = $this->getCaretButton($block);
+        assertNotNull($button, 'Caret button for block ' . $position . ' was not found in the page.');
+        $button->click();
+    }
+
+
+
```
robbieaverill: Too many spaces

tests/Behat/features/element-editor.feature
```diff
@@ -22,19 +22,20 @@ Feature: View types of elements in an area on a page
...
-  #    Then I should not see "Title (displayed if checked)"
-  #      # The form should still exist, just be hidden from the user
-  #      And the "HTML" HTML field should contain "Some content"
-  #      # The content shows in the preview when the form is not shown
-  #      And I should see "Some content"
+    Scenario: I can preview a block and hide the form again
+      Given I see a list of blocks
+      Then I should see block 1
+      # The entire block should be clickable to reveal the form
+      Given I click on block 1
```
robbieaverill: I think this and the next Given might be better read as When

PR #385, opened 2018-09-12
API Move elemental editor actions menu items into their own HOCs

### LINKED ISSUE #384: Element editor's actions menu items should live in their own components
We're adding more and more action menu items at the moment, so the `Header` component is filling up quickly with things that aren't directly related to a "header", but to an actions menu within it.

I think we should move each of the actions into their own components

robbieaverill:
This adds an ElementActions component which is responsible for rendering the menu with the items in it, and adds Injector transforms with individual HOCs applying each action to the menu, i.e. archive, publish etc

Resolves #384


robbieaverill:
Discussing with @unclecheese at the moment, we agree that the injection route is the right way so that user code could add more actions if they want.

We should try and centralise some of the logic in the HOCs here though, so perhaps we provide some props like title, handle (callback), disabled (bool) and name and the rendering is handled by something abstract.

client/src/components/ElementActions/ArchiveAction.js
```diff
@@ -0,0 +1,46 @@
...
+ * Adds the elemental menu action to archive a block of any state
+ */
+const ArchiveAction = (WrappedComponent) => (props) => (
+  <WrappedComponent {...props}>
+    {props.children}
+
+    <AbstractAction
+      title={i18n._t('ArchiveAction.ARCHIVE', 'Archive')}
+      extraClass="element-editor__actions-archive"
+      onClick={(event) => {
```
ScopeyNZ: How about:

```js
ArchiveAction = (Thing) => (props) => {
	const handleClick = () => {...}
	return (
		...
		onClick={handleClick}
		...
	)
}
```
?

client/src/components/ElementActions/ArchiveAction.js
```diff
@@ -0,0 +1,46 @@
...
+        }
+
+        // eslint-disable-next-line no-alert
+        if (handleArchiveBlock && confirm(archiveMessage)) {
+          handleArchiveBlock(id);
+        }
+      }}
+    />
+  </WrappedComponent>
+);
```
ScopeyNZ: It took us a bit of time to get our head around this. Partially because of `WrappedComponent` vs `WrappingComponent`. Dylan has suggested perhaps a nicer way to handle this, which gives the wrapped "`MenuComponent`" more flexibility:

```js
const HOC = (MenuComponent) => (props) => {
	const newProps = {
		...props,
		actions: [
			...props.actions,
			<AbstractAction {...} /> OR { title: 'thing', handleClick: () => {...} } // <- Shaped object type
		]
	};

	return <MenuComponent {...newProps} />
}
```

Edit tabs could be added to the `MenuComponent` actions too, and perhaps something for separators too - reactstrap has a `<Separator>` component for separators, and if we go the "shaped object type" approach we could have nested arrays?

This gives MenuComponent more flexibility to choose how actions render. We'd like to hear your thoughts.

client/src/components/ElementEditor/ElementActions.js
```diff
@@ -0,0 +1,83 @@
...
+      <ActionMenuComponent
+        id={`element-editor-actions-${id}`}
+        className={'element-editor-header__actions-dropdown'}
+        dropdownMenuProps={{ right: true }}
+        dropdownToggleClassNames={dropdownToggleClassNames}
+        toggleCallback={(event) => event.stopPropagation()}
+      >
+        { this.renderEditTabs() }
+        { !editTabs || !editTabs.length || <DropdownItem divider /> }
+        { children }
```
ScopeyNZ: We need to check that children has length before rendering the divider too. Additionally, this might require some more thought for the situation where both are empty 🤔

client/src/components/ElementActions/ArchiveAction.js
ScopeyNZ: Really this doesn't make a difference, I was just suggesting for readability. Not really fussed here.

client/src/components/ElementActions/ArchiveAction.js
```diff
@@ -0,0 +1,46 @@
...
+import React from 'react';
+import { compose } from 'redux';
+import AbstractAction from 'components/ElementActions/AbstractAction';
+import archiveBlockMutation from 'state/editor/archiveBlockMutation';
+import i18n from 'i18n';
+
+/**
+ * Adds the elemental menu action to archive a block of any state
+ */
+const ArchiveAction = (WrappedComponent) => (props) => (
```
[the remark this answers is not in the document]
robbieaverill: The reason I chose this is because it's heavily used as an example in the [React HOCs documentation](https://reactjs.org/docs/higher-order-components.html)

client/src/components/ElementActions/AbstractAction.js
```diff
@@ -0,0 +1,34 @@
+import React, { PropTypes } from 'react';
+import classNames from 'classnames';
+
+/**
+ * Renders an action item for the "more actions" dropdown on elements
+ */
+const AbstractAction = (props) => {
```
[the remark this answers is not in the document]
robbieaverill: They are injected, therefore should be injectable (see registerTransforms.js). If injector transformations can't be modified by usercode then I think we probably need to bring that up in core

client/src/components/ElementActions/ArchiveAction.js
robbieaverill: Sure

client/src/components/ElementActions/ArchiveAction.js
robbieaverill: Yeah I'd like to use the AbstractAction as much as possible for this menu - I've already added to the method that renders the CMS tab actions

robbieaverill: I've pushed a new commit - let me know if that satisfies your suggestion :)

client/src/components/ElementActions/ArchiveAction.js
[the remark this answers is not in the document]
robbieaverill: Resolved -> renamed to MenuComponent as suggested further down in the feedback

robbieaverill:
@ScopeyNZ how about we merge now and refactor/improve going forward? #continualimprovementoverperfection

ScopeyNZ:
Sure.

PR #386, opened 2018-09-12
FIX Add namespaces to inline edit form field names

### LINKED ISSUE #381: Inline form fields need to be psuedo-namespaced
With #372 all form fields for inline element editing are given HTML names from their element, e.g. `Title`, `HTML` etc.

This not only conflicts between elements but also with the page that the forms sit in. Changing a basic content block's Title then overrides the page's Title when saving (#328).

I propose that we prefix element CMS field names with something unique to them, e.g.:

```diff
- input type="text" name="Title"
+ input type="text" name="PageElements[123][Title]"
```

Where 123 is the element's ID

## Pull requests

- [ ] https://github.com/dnadesign/silverstripe-elemental/pull/386

robbieaverill:
This ensures that inline edit forms don't conflict with each other or the page that they're rendered in.

I would've liked the format to be `PageElements[123][FieldName]` but either the silverstripe-admin module or react-form package have trouble matching the form state value against the field when it has brackets in the name like this. The data is all in the state correctly, but doesn't end up in the form field.

At least for now we can look for all request vars that start with `PageElements_` and process those, rather than looping an array from HTML inputs.

Resolves #381

### DIFF
--- a/src/Forms/EditFormFactory.php
+++ b/src/Forms/EditFormFactory.php
@@ -56,6 +56,6 @@ class EditFormFactory extends DefaultFormFactory
         foreach ($fields->dataFields() as $field) {
             $namespacedName = sprintf(self::FIELD_NAMESPACE_TEMPLATE, $elementID, $field->getName());
             $field->setName($namespacedName);
-        };
+        }
     }
 }
--- a/src/Forms/EditFormFactory.php
+++ b/src/Forms/EditFormFactory.php
@@ -4,16 +4,27 @@ namespace DNADesign\Elemental\Forms;
 
 use SilverStripe\Control\RequestHandler;
 use SilverStripe\Forms\DefaultFormFactory;
+use SilverStripe\Forms\FieldList;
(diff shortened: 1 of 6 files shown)


src/Forms/EditFormFactory.php
```diff
@@ -30,4 +41,21 @@ protected function getFormFields(RequestHandler $controller = null, $name, $cont
...
+     * @param array $context
+     */
+    protected function namespaceFields(FieldList $fields, array $context)
+    {
+        $elementID = $context['Record']->ID;
+
+        foreach ($fields->dataFields() as $field) {
+            $namespacedName = sprintf(self::FIELD_NAMESPACE_TEMPLATE, $elementID, $field->getName());
+            $field->setName($namespacedName);
+        };
```
ScopeyNZ: Been writing too much javascript 😂

robbieaverill: Ha! True

Issue #390, opened 2018-09-14
History Tab is broken on first load


eehondas:
Hello,
I was looking around at the improvements that have been shipped with Silverstripe 4.2.
First of all thank you for the implementation of the History Tab on the Elemental Page and on the elements itself.
I think the functionality is very good from a user standpoint.
I believe there is a small bug on the History Tab of an element detailed view (say ElementContent for example):
The history tab is accessible before the element is saved for the first time (created), because of this when trying to access the History Tab on first load it the CMS gets stuck in an infinite loop of loading (I assume because it is looking for Versions of that specific object that don't exist yet).
Anyway, I believe the fix is as quick as checking for the presence of an ID here:
```php
            // Support for new history viewer in SS 4.2+
            if (class_exists(HistoryViewerField::class)) {
                Requirements::javascript('dnadesign/silverstripe-elemental:client/dist/js/bundle.js');

                $historyViewer = HistoryViewerField::create('ElementHistory');
                $fields->addFieldToTab('Root.History', $historyViewer);

                $fields->fieldByName('Root.History')
                    ->addExtraClass('elemental-block__history-tab tab--history-viewer');
            }
```
While I'm at it I wanted to ask: seeing that the History functionality of a normal Silverstripe Page and an Elemental Page are now completely different, do you guys plan to roll-out this more-advanced functionality with the preview to the SiteTree History Tab itself? Are there plans for the HistoryViewerField to be used in a normal DataObject?

Thanks guys.

robbieaverill:
Hi @eehondas, thank you for the feedback!

Yes you’re right, we should check the element exists before pushing the history field in getCMSFields. Would you mind making a PR for it?

With SilverStripe 4.3 the new history viewer will be enabled by default on all pages. We didn’t enable it by default in 4.2 because it doesn’t have compare mode at that version, so is a feature regression over the legacy history controller. You can still opt into it everywhere if you want to, by enabling an extension (see the example in this repository)

ScopeyNZ:
Hi @eehondas. 3.0 is perfect. We can release this fix as a bugfix (eg. `3.0.1`) and we can merge up to `3` for another minor release if applicable (`3.1.0`). Master is for the next major (`4.0.0`). Hopefully that makes the branching strategy a little clearer?

ScopeyNZ:
Closed in #391

PR #391, opened 2018-09-16
Check object exists before pushing History Tab to BaseElement

eehondas:
Hello,
This is a quick fix relative to issue #390 .
I tested it and it does prevent the history tab from showing (duh!).
I used a separate logic block rather than adding an && condition because I assume the inner logic block will be removed in future instances and this way you can just take away the inner condition. Not rocket science obviously.

Thanks for the module again.

### DIFF
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -300,15 +300,18 @@ class BaseElement extends DataObject
                 $fields->removeByName('Style');
             }
 
-            // Support for new history viewer in SS 4.2+
-            if (class_exists(HistoryViewerField::class)) {
-                Requirements::javascript('dnadesign/silverstripe-elemental:client/dist/js/bundle.js');
+            // Check object exists before pushing History Tab
+            if($this->ID) {
+              // Support for new history viewer in SS 4.2+
+              if (class_exists(HistoryViewerField::class)) {
+                  Requirements::javascript('dnadesign/silverstripe-elemental:client/dist/js/bundle.js');
 
-                $historyViewer = HistoryViewerField::create('ElementHistory');
-                $fields->addFieldToTab('Root.History', $historyViewer);
+                  $historyViewer = HistoryViewerField::create('ElementHistory');
+                  $fields->addFieldToTab('Root.History', $historyViewer);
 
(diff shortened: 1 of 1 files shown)


src/Models/BaseElement.php
```diff
@@ -300,15 +300,18 @@ public function getCMSFields()
...
-            if (class_exists(HistoryViewerField::class)) {
-                Requirements::javascript('dnadesign/silverstripe-elemental:client/dist/js/bundle.js');
-
-                $historyViewer = HistoryViewerField::create('ElementHistory');
-                $fields->addFieldToTab('Root.History', $historyViewer);
-
-                $fields->fieldByName('Root.History')
-                    ->addExtraClass('elemental-block__history-tab tab--history-viewer');
+            // Check object exists before pushing History Tab
+            if($this->ID) {
```
ScopeyNZ: Can you use `$this->isInDB()` here? It's pretty similar logic but the method might be updated to more accurately determine if the object is persisted.

ScopeyNZ: Also, combining the conditions is nicer because then you're only updated one line instead of 12.

ScopeyNZ:
Thanks for the PR @eehondas! I left a few requests and the automated tests are only failing due to some PHPCS (linting) failures pasted here for clarity:

```
 304 | ERROR | [x] Expected 1 space(s) after IF keyword; 0 found
 306 | ERROR | [x] Line indented incorrectly; expected 16 spaces, found 14
 307 | ERROR | [x] Line indented incorrectly; expected at least 20 spaces
     |       |     found 18
 309 | ERROR | [x] Line indented incorrectly; expected at least 20 spaces
     |       |     found 18
 310 | ERROR | [x] Line indented incorrectly; expected at least 20 spaces
     |       |     found 18
 312 | ERROR | [x] Line indented incorrectly; expected at least 20 spaces
     |       |     found 18
 314 | ERROR | [x] Line indented incorrectly; expected 16 spaces, found 14
```

If you wouldn't mind fixing that up I can get this merged for you 🙂

eehondas:
Sure will, sorry, I thought the different lines would make life easier but it's actually the opposite.

PR #397, opened 2018-09-19
FIX FormBuilderLoader now keeps its state across remounts so we can remove it rather than hide

ScopeyNZ:
This relies on silverstripe/silverstripe-admin#654

With changes to FormBuilderLoader it can be unmounted and remounted without losing form state inbetween. This was done in prep for reordering forms as part of drag and drop. It makes sense to remove the custom logic around hiding the form (instead of removing it) that we had to handle not losing form state previously.


robbieaverill:
I think the behat tests need to be updated - there's a test that is checking for the functionality that you've removed

ScopeyNZ:
I've updated the PR to fix the Behat tests. It now relies on two more PRs now:

silverstripe/silverstripe-framework#8461
silverstripe/silverstripe-admin#676

Nothing's ever easy...

ScopeyNZ:
I reran the tests after Robbie merged those two core PRs - All green 🙂

ScopeyNZ:
Merging with Robbie and Raissa's approval.

PR #401, opened 2018-09-20
NEW Show message if element list is empty

### LINKED ISSUE #351: Show a message when no blocks exist
**Version: 4.x-dev**

image

When no blocks exist yet we should show a message to hint to a user they need to add blocks to their page. 

Refer to designs:

image

https://projects.invisionapp.com/dsm/silver-stripe/silver-stripe/asset/components/5b90499d4d45900010a7be60

raissanorth:
Resolves #351 

Changes in this PR render a message in case the element list is empty, once loading has finished. Behat and Jest tests were added. 

image


client/src/components/ElementEditor/ElementList.js
```diff
@@ -29,10 +31,14 @@ class ElementList extends Component {
   renderBlocks() {
     const { ElementComponent, blocks } = this.props;
 
+    // Blocks can be either null or an empty array
```
robbieaverill: If blocks is null do you not want to show the message below?

ScopeyNZ: No. Blocks will be null/undefined while the GraphQL query is running I think. It will be an empty array if there's nothing

raissanorth: Yes, what @ScopeyNZ is saying is correct. I should have made that clearer. By returning nothing on blocks being null we prevent that the message _and_ the loading component are rendered at the same time, which looks really ugly.

NightJar: Perhaps reflecting this information in the comment would be good :)

PR #402, opened 2018-09-21
NEW Add toast messages when publishing and unpublishing blocks

### LINKED ISSUE #364: Toast messaging when saving a block
Implement toast messages to provide further notification that block versioning is complete. Wireframes: https://projects.invisionapp.com/share/S4MQXACDB2P#/screens/307488194_Save-Publish-Unpublish-Archive

ACs:
- Toast message appears when individual blocks are:
   - [x] Saved
   - [x] Published
   - [x] Unpublished

Note: the wireframe also shows a saving animation with the block icon, this functionality is captured in: https://github.com/dnadesign/silverstripe-elemental/issues/292

raissanorth:
Partially resolves #364 (since saving hasn't been implemented yet). 

Uses jQuery to push toast messages - as a temporary solution - as suggested in the issue's comments:

image


client/src/components/ElementActions/PublishAction.js
```diff
@@ -3,6 +3,7 @@ import { compose } from 'redux';
 import AbstractAction from 'components/ElementActions/AbstractAction';
 import publishBlockMutation from 'state/editor/publishBlockMutation';
 import i18n from 'i18n';
+import jQuery from 'jquery';
```
robbieaverill: I think we discussed in Slack, you could probably get this from `window.jQuery` if you wanted to keep it a bit slimmer

client/src/components/ElementActions/PublishAction.js
```diff
@@ -11,10 +12,18 @@ const PublishAction = (MenuComponent) => (props) => {
...
 
-    const { id, version, actions: { handlePublishBlock } } = props;
+    const { id, title, version, actions: { handlePublishBlock } } = props;
 
     if (handlePublishBlock) {
       handlePublishBlock(id, 'DRAFT', 'LIVE', version);
+      jQuery.noticeAdd({
+        text: i18n._t(
+          'UnpublishAction.SUCCESS_NOTIFICATION',
+          `Published '${title} successfully`
```
robbieaverill: Missing a single quote at the end of the variable. Also we should make this a placeholder e.g. `Published {title} successfully` then use `i18n.inject()` - that way the translation value is constant rather than dynamic

client/src/components/ElementActions/UnpublishAction.js
```diff
@@ -11,10 +12,18 @@ const UnpublishAction = (MenuComponent) => (props) => {
...
 
-    const { id, actions: { handleUnpublishBlock } } = props;
+    const { id, title, actions: { handleUnpublishBlock } } = props;
 
     if (handleUnpublishBlock) {
       handleUnpublishBlock(id);
+      jQuery.noticeAdd({
+        text: i18n._t(
+          'UnpublishAction.SUCCESS_NOTIFICATION',
+          `Removed '${title}' from the published page`
```
robbieaverill: Same comment as above - better to use a placeholder in this translation value

tests/Behat/Context/FeatureContext.php
```diff
@@ -182,6 +221,38 @@ protected function getArchiveButton($position)
         return $button;
     }
 
+    /**
+     * Returns the publish button for a specific block if it exists
+     *
+     * @param $position
```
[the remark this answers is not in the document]
raissanorth: Thanks for pointing that out. Made a PR https://github.com/dnadesign/silverstripe-elemental/pull/405 to fix.

Issue #294, opened 2018-08-05
Add block to page (inline)


brynwhyman:
### User Story

As a CMS user, I want to be able to add new content blocks within a page level between current content blocks, so that I can do this quickly and in context with the other page blocks.

### Acceptance Criteria

- [x] Hover-over between current page blocks presents styling as per wireframe: https://projects.invisionapp.com/share/ZRL4T2HQAGB#/screens/301918525
- [x] Clicking a “+” button in between blocks presents a pop-up with all blocks available in a format that’s easy to view
- [x] Clicking a new block type adds the block to the present gap between blocks. Styling animation pushes current page blocks below new block as per wireframe: https://projects.invisionapp.com/share/ZRL4T2HQAGB#/screens/301918523 
- [x] As per designs, when hovering between blocks there is a indicator that a block can be added between/above/below existing blocks (delayed interaction so the line isn’t triggered as you move the cursor across the screen).  https://invis.io/BYMQXG07QZH#/307488192_Adding_Blocks-_In-Page

### Notes
- Pop-up work is also covered in: https://github.com/dnadesign/silverstripe-elemental/issues/293

NightJar:
Done https://github.com/dnadesign/silverstripe-elemental/pull/443
in tandem with #340

PR #421, opened 2018-10-01
NEW Allow adding blocks inline via HoverBar by utilising AddElementPo…

raissanorth:
…pover

Addresses three of the four ACs in #294 

Hover bar rendered with .2s delay from within two element blocks:

image

Hover bar rendered instantaneously when hovering over the more actions area between two elements:

image

The placement of the popover in relation to the hover bar and in particular the add button disaligns with the original designs - see https://projects.invisionapp.com/d/main#/console/12297909/318343221/inspect - since the approach to raise the z-index of the hover bar's plus icon over that of the popover causes the unintended effect that the icon is overlaying the popover in the case that there is not enough space to render the popover below an element:

image

To avoid the default margin between target - the hover bar - and ReactStrap's ` Popover` has been maintained, as shown below:

image

For consistency, the top margin has been re-added to the Add block button at the top of the Elemental list as shown below:

image


raissanorth:
There are some cross-browser compatibility issues which are logged at [placeholder].

PR #427, opened 2018-10-02
ENHANCEMENT: Writeable local cache API

unclecheese:
Allows imperative writes to the cache so that components can be eagerly loaded instead of lazily writing to local cache in accessors. The advantage is that cached values can be bulk assigned (e.g. in a single query).

Tests forthcoming upon approval of general approach.

### DIFF
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -676,7 +676,8 @@ class BaseElement extends DataObject
             }
         }
 
-        $this->cacheData['area_relation_name'] = $result;
+        $this->setAreaRelationNameCache($result);
+
         return $result;
     }
 
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -375,7 +375,7 @@ class BaseElement extends DataObject
      * @param string $name
      * @return $this
      */
-    public function setAreaRelationNameCache(string $name)
+    public function setAreaRelationNameCache($name)
     {
         $this->cacheData['area_relation_name'] = $name;
 
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -372,6 +372,17 @@ class BaseElement extends DataObject
     }
 
     /**
+     * @param string $name
+     * @return $this
+     */
+    public function setAreaRelationNameCache(string $name)
+    {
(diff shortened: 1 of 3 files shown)


robbieaverill:
Looks good to me

ScopeyNZ:
I'm a little worried about `->Elements()->add()`. Maybe this will be most noticeable during tests but we'll need some way to bust the area cache (and perhaps always do it when adding to the HasManyList...)

unclecheese:
Have rebased. Diff is a bit clearer now. :)

unclecheese:
@ScopeyNZ Yeah, it's a bit of "use, but understand the risks". If you're caching the elements, you're implicitly making them immutable.

ScopeyNZ:
> @ScopeyNZ Yeah, it's a bit of "use, but understand the risks". If you're caching the elements, you're implicitly making them immutable.

Right on re-reading I can see it's not automatically populated... Understood.

src/Models/ElementalArea.php
```diff
@@ -87,6 +88,41 @@ public function forTemplate()
...
+    public function setOwnerPageCached(DataObject $page)
+    {
+        $this->cacheData['owner_page'] = $page;
+
+        return $this;
+    }
+
+    /**
+     * A cache-aware accessor for the elements
+     * @return ArrayList|DataList
```
ScopeyNZ: Could you keep the existing type hint? So make it `ArrayList|HasManyList|BaseElement[]`?

robbieaverill: PSR-5 (draft) is introducing type collections 😃 `ArrayList<BaseElement>`

ScopeyNZ: To this effect you can also remove the `@method` hint at the top.

src/Models/BaseElement.php
```diff
@@ -371,6 +371,17 @@ public function getController()
         return $this->controller;
     }
 
+    /**
+     * @param string $name
+     * @return $this
+     */
+    public function setAreaRelationNameCache(string $name)
```
robbieaverill: Oh, we can't type hint scalars in PHP 5.6 (which we support)

PR #428, opened 2018-10-03
NEW Edit tab click changes tab state 

raissanorth:
Related issue: 
- https://github.com/dnadesign/silverstripe-elemental/issues/325

Related PRs: 

- https://github.com/silverstripe/silverstripe-admin/pull/669
- https://github.com/silverstripe/silverstripe-framework/pull/8440

This is work in progress for the above issue. The related PRs connect `Tabs` to the redux-form store. 

I would like some initial feedback on the approach taken. Since we are dispatching an action I am not sure if bubbling up the edit tab click from `ElementActions` to `Element`, saving the active tab in `Element`'s local state and passing it down as a prop to `InlineEditForm` is the most elegant solution.

Outstanding work includes styling of the selected edit tab and cursor, as well as expanding the form on edit tab click.


client/src/components/ElementEditor/ElementActions.js
```diff
@@ -9,6 +9,22 @@ import AbstractAction from 'components/ElementActions/AbstractAction';
...
 class ElementActions extends Component {
+  constructor(props) {
+    super(props);
+    this.handleEditTabsClick = this.handleEditTabsClick.bind(this);
+  }
+
+  /**
+   * Set the active tab
+   *
+   * @param activeTab
```
robbieaverill: Missing type

client/src/components/ElementEditor/ElementList.js
```diff
@@ -48,7 +48,7 @@ class ElementList extends Component {
     }
 
     return blocks.map((element) => (
-      <div>
+      <div key={element.ID}>
```
robbieaverill: I'm curious about whether giving it the same key as its child element might cause rendering issues... I don't know enough about how this works in React though

client/src/components/ElementEditor/InlineEditForm.js
```diff
@@ -1,9 +1,57 @@
...
+import { change } from 'redux-form/lib';
+import Config from 'lib/Config';
 
 class InlineEditForm extends PureComponent {
+  constructor(props) {
+    super(props);
+
+    const sectionKey = 'DNADesign\\Elemental\\Controllers\\ElementalAreaController';
+    const section = Config.getSection(sectionKey);
+    const formNameTemplate = section.form.elementForm.formNameTemplate;
```
robbieaverill: You can use `loadElementSchemaValue('formNameTemplate')` for this (see `client/src/state/editor/loadElementSchemaValue.js`)

client/src/components/ElementEditor/InlineEditForm.js
```diff
@@ -1,9 +1,57 @@
...
+    this.updateFormTab = this.updateFormTab.bind(this);
+  }
+
+  componentDidMount() {
+    this.updateFormTab();
+  }
+
+  /**
+   * Update the active tab on tab actions menu button click event.
+   * @param event onChangeEvent from a input field.
```
robbieaverill: Missing `prevProps` declaration

client/src/components/ElementEditor/Header.js
```diff
@@ -94,9 +94,18 @@ class Header extends Component {
...
+          className="element-editor-header__actions"
+
+        >
+          {expandable &&
+          <div
+            role="none"
+            onClick={(event) => event.stopPropagation()
+            }
+          >
+            <ElementActionsComponent {...this.props} /></div>
```
ScopeyNZ: What's this wrapping `div` doing? The `</div>` should be a on a new line.

client/src/components/ElementEditor/Element.js
```diff
@@ -2,9 +2,15 @@
 
 import React, { Component, PropTypes } from 'react';
 import { elementType } from 'types/elementType';
+import { bindActionCreators, compose } from 'redux';
 import { inject } from 'lib/Injector';
 import i18n from 'i18n';
 import classNames from 'classnames';
+import { connect } from 'react-redux';
+import { change } from 'redux-form';
```
ScopeyNZ: Is this (and the redux-form dependancy) actually needed?

client/src/components/ElementEditor/Element.js
```diff
@@ -130,23 +183,59 @@ class Element extends Component {
...
+  const elementId = ownProps.element.ID;
+  const elementFormSchema = loadElementSchemaValue('schemaUrl', elementId);
+
+  const stateValue = state.form.formState.element[elementName].values.Root;
+
+  // Search out a default value for the active tab if it is not already in the state.
+  // {@see Tabs.getDefaultActiveKey}
+  const filterFieldsForTabs = (field) => field.component === 'Tabs';
+  const defaultValue = state.form.formSchemas[elementFormSchema].schema.fields
+    .filter(filterFieldsForTabs)[0].children[0].name;
```
ScopeyNZ: You could use `find` instead of `filter` here. Also, it'd probably be good to defensively program in for the possibility there's no tabs - although admittedly there's probably some work to do elsewhere for this...

client/src/components/ElementEditor/ElementList.js
ScopeyNZ: In this case I think we can remove the key from the child component. I'm not 100% on this either but I'm pretty sure it's only used for the top level dom elements within a loop.

client/src/components/ElementEditor/Element.js
raissanorth: Without them we cannot update the form state for the active tab. Unless you can come up with a more elegant solution, we do need them. 'connect' allows us to connect the Element component to the Redux store and change allows us to dispatch an action.

ScopeyNZ: Yep - I was too tired last night - missed the use of it within `mapDispatchToProps` 👍

client/src/components/ElementEditor/ElementList.js
raissanorth: "Adjacent JSX elements must be wrapped in an enclosing tag" 😉

Issue #340, opened 2018-08-21
Add a block when clicking a button in the "Add Block" popover


ScopeyNZ:
#338 implements an "Add Block" popover. When the various block types shown in this popover are clicked, a block needs to be added to the relevant elemental area.

### A/Cs
- [x] Clicking on a block in the popover will add a block to the relevant elemental area
- [x] Block is already saved but unpublished
- [ ] ~~Animations as specified by the designs are implemented~~

ScopeyNZ:
@clarkepaul For accessibility do you think we should add an A/C to make the newly added block the next thing that you tab to?

clarkepaul:
Tough one, you might want to be adding multiple blocks (one after the other). I'm thinking to remain within the popover, after selecting a new block. Happy to hear what others think. Agree it should be an AC though.

clarkepaul:
Three options as I see it:
1. Close the popover, focus moves to the "Add" button
2. Close popover, focus moves to newly added block
3. Keep popover open, focus is on popover ready for another block to be added

I took at look at a few examples from others and they didn't really consider this, although clicking through the experience feels quite intuitive if you go to the new block as you mentioned @ScopeyNZ (opt. 3).

ScopeyNZ:
> the experience feels quite intuitive if you go to the new block as you mentioned @ScopeyNZ (opt. 3).

This was my initial thought. To be honest I didn't consider the idea of even leaving the popover open after clicking to add a block.

ScopeyNZ:
@raissanorth has done a "part 1" PR at #404. There's still a few things that need to be addressed:

- [ ] Padding and positioning is a tiny bit off. Currently it just uses default popover positioning/padding
- [ ] Clicking "Add block" in the designs has it just adding the block ready for in-line editing (with some fancy animations)
  - @robbieaverill @NightJar @raissanorth How do you think this should be done? With a `CreateBlock` mutation that just takes the block type? A save would have a lot more options and block type is not one of those. Creating seems quite different.
  - @clarkepaul @newleeland How should this work with non in-line editable blocks? Do you expect people to add, then click on the block for editing or should it take you straight to a "GridFieldDetailForm"?

robbieaverill:
Sounds good to me. A custom mutation which takes in the block type makes sense to me. We can either push an optimistic update to the Apollo cache with the new block or refetch the blocks again in order to re-render the new one

brynwhyman:
[...]

PR #429, opened 2018-10-03
Add mutation to support adding blocks

NightJar:
Without the need to add extra details on a `GridFieldDetailForm`. This paves the way for #340 and #294 to be implemented in React.

### DIFF
--- a/src/GraphQL/AddElementToAreaMutation.php
+++ b/src/GraphQL/AddElementToAreaMutation.php
@@ -1,0 +1,76 @@
+<?php
+
+namespace DNADesign\Elemental\GraphQL;
+
+use DNADesign\Elemental\Models\BaseElement;
+use DNADesign\Elemental\Models\ElementalArea;
+use DNADesign\Elemental\Services\ReorderElements;
+use GraphQL\Type\Definition\ResolveInfo;
+use GraphQL\Type\Definition\Type;
+use InvalidArgumentException;
+use SilverStripe\Core\Injector\Injector;
+use SilverStripe\GraphQL\MutationCreator;
+use SilverStripe\GraphQL\OperationResolver;
+
+class AddElementToAreaMutation extends MutationCreator implements OperationResolver
+{
+    public function attributes()
+    {
+        return [
+            'name' => 'addElementToArea',
+            'description' => 'Adds an Element to an ElementalArea, optionally after another Element'
+        ];
+    }
+
+    public function type()
+    {
+        return $this->manager->getType('Block');
+    }
+
+    public function args()
+    {
+        return [
+            'ClassName' => ['type' => Type::nonNull(Type::string())],
+            'ElementalAreaID' => ['type' => Type::nonNull(Type::id())],
+            'AfterElementID' => ['type' => Type::id()],
+        ];
+    }
+
+    public function resolve($object, array $args, $context, ResolveInfo $info)
+    {
+        $elementClass = $args['ClassName'];
+        $elementalAreaID = $args['ElementalAreaID'];
+        $afterElementID = isset($args['AfterElementID']) ? $args['AfterElementID'] : null;
+
+        if (!is_subclass_of($elementClass, BaseElement::class)) {
+            throw new InvalidArgumentException("$elementClass is not a subclass of " . BaseElement::class);
+        }
+
+        $elementalArea = ElementalArea::get()->byID($elementalAreaID);
+
+        if (!$elementalArea) {
+            throw new InvalidArgumentException("Invalid ElementalAreaID: $elementalAreaID");
+        }
+
+        if (!$elementalArea->canEdit($context['currentUser'])) {
+            throw new InvalidArgumentException("The current user has insufficient permission to edit ElementalAreas");
+        }
+
+        $newElement = Injector::inst()->create($elementClass);
+
+        if (!$newElement->canEdit($context['currentUser'])) {
+            throw new InvalidArgumentException(
+                'The current user has insufficient permission to edit Elements'
+            );
+        }
+
+        $elementalArea->Elements()->add($newElement);
+
+        if (!is_null($afterElementID)) {
+            $reorderer = Injector::inst()->create(ReorderElements::class, $newElement);
+            $reorderer->reorder($afterElementID);
+        }
+
+        return $newElement;
+    }
+}
--- a/src/GraphQL/SortBlockMutationCreator.php
+++ b/src/GraphQL/SortBlockMutationCreator.php
@@ -2,9 +2,11 @@
 namespace DNADesign\Elemental\GraphQL;
 
 use DNADesign\Elemental\Models\BaseElement;
+use DNADesign\Elemental\Services\ReorderElements;
 use GraphQL\Type\Definition\ResolveInfo;
 use GraphQL\Type\Definition\Type;
 use SilverStripe\Core\Convert;
+use SilverStripe\Core\Injector\Injector;
 use SilverStripe\GraphQL\MutationCreator;
 use SilverStripe\GraphQL\OperationResolver;
 use SilverStripe\ORM\DataObject;
@@ -40,73 +42,23 @@ class SortBlockMutationCreator extends MutationCreator implements OperationResol
 
     public function resolve($object, array $args, $context, ResolveInfo $info)
     {
-        /** @var BaseElement $block */
-        $block = BaseElement::get_by_id($args['ID']);
-
-        if (!$block) {
-            throw new \InvalidArgumentException(sprintf(
+        $element = BaseElement::get()->byID($args['ID']);
+        
+        if (!$element) {
+            throw new InvalidArgumentException(sprintf(
                 '%s#%s not found',
                 BaseElement::class,
                 $args['ID']
             ));
         }
 
-        if (!$block->canEdit($context['currentUser'])) {
-            throw new \InvalidArgumentException(
-                'Changing the sort order of blocks is not allowed'
+        if (!$element->canEdit($context['currentUser'])) {
+            throw new InvalidArgumentException(
+                'Changing the sort order of blocks is not allowed for the current user'
             );
         }
 
-        $parentId = $block->ParentID;
-        $blockPosition = (int) $block->Sort;
-
-        $sortAfterPosition = 0;
-        $afterBlockID = $args['AfterBlockID'];
-        if ($afterBlockID) {
-            /** @var BaseElement $afterBlock */
-            $afterBlock = BaseElement::get_by_id($afterBlockID);
-
-            if (!$afterBlock) {
-                throw new \InvalidArgumentException(
-                    'Trying to sort block after a block in a different elemental area'
-                );
-            }
-            if ($afterBlock->ParentID !== $parentId) {
-                throw new \InvalidArgumentException(sprintf(
-                    '%s#%s not found',
-                    BaseElement::class,
-                    $parentId
-                ));
-            }
-
-            $sortAfterPosition = (int) $afterBlock->Sort;
-        }
-
-        // We are updating records with SQL queries to avoid the ORM triggering the creation of new versions
-        // for each element that is affected by this reordering.
-        $tableName = Convert::raw2sql(DataObject::getSchema()->tableName(BaseElement::class));
-
-        if ($sortAfterPosition < $blockPosition) {
-            $operator = '+';
-            $filter = "\"$tableName\".\"Sort\" > $sortAfterPosition AND \"$tableName\".\"Sort\" < $blockPosition";
-            $newBlockPosition = $sortAfterPosition + 1;
-        } else {
-            $operator = '-';
-            $filter = "\"$tableName\".\"Sort\" <= $sortAfterPosition AND \"$tableName\".\"Sort\" > $blockPosition";
-            $newBlockPosition = $sortAfterPosition;
-        }
-
-        $query = SQLUpdate::create()
-           ->setTable("\"$tableName\"")
-           ->assignSQL('"Sort"', "\"$tableName\".\"Sort\" $operator 1")
-           ->addWhere([$filter, "\"$tableName\".\"ParentID\"" => $parentId]);
-
-        $query->execute();
-
-        // Now use the ORM to write a new version of the record that we are directly reordering
-        $block->Sort = $newBlockPosition;
-        $block->write();
-
-        return $block;
+        $reorderingService = Injector::inst()->create(ReorderElements::class, $element);
+        return $reorderingService->reorder($args['AfterBlockID']);
     }
 }
--- a/src/Services/ReorderElements.php
+++ b/src/Services/ReorderElements.php
@@ -1,0 +1,120 @@
+<?php
+
+namespace DNADesign\Elemental\Services;
+
+use DNADesign\Elemental\Models\BaseElement;
+use InvalidArgumentException;
+use SilverStripe\Core\Convert;
+use SilverStripe\ORM\DataObject;
+use SilverStripe\ORM\Queries\SQLUpdate;
+
+class ReorderElements
+{
+    /**
+     * @var BaseElement
+     */
+    protected $element;
+
+    /**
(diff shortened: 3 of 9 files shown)


src/GraphQL/AddElementToAreaMutation.php
```diff
@@ -0,0 +1,64 @@
...
+        $elementalAreaID = $args['ElementalAreaID'];
+        $afterElementID = isset($args['AfterElementID']) ? $args['AfterElementID'] : null;
+
+        if (!is_subclass_of($elementClass, BaseElement::class)) {
+            throw new InvalidArgumentException("$elementClass is not a subclass of " . BaseElement::class);
+        }
+
+        $elementalArea = ElementalArea::get()->byID($elementalAreaID);
+
+        if (!$elementalArea) {
```
robbieaverill: You should also check `$elementalArea->canEdit()` here, or some other kind of permission check using `$context['currentUser']`

src/GraphQL/AddElementToAreaMutation.php
```diff
@@ -0,0 +1,64 @@
...
+            throw new InvalidArgumentException("Invalid ElementalAreaID: $elementalAreaID");
+        }
+
+        $newElement = $elementClass::create();
```
robbieaverill: I think `Injector::inst()->create($elementClass)` would be better

src/GraphQL/AddElementToAreaMutation.php
```diff
@@ -0,0 +1,64 @@
...
+        $elementalArea->Elements()->add($newElement);
+
+        if (!is_null($afterElementID)) {
+            $reorderer = new ReorderElements($newElement);
```
robbieaverill: Use Injector to get this class 👍

src/GraphQL/SortBlockMutationCreator.php
```diff
@@ -1,14 +1,15 @@
...
-use SilverStripe\GraphQL\MutationCreator;
-use SilverStripe\GraphQL\OperationResolver;
 use SilverStripe\ORM\DataObject;
+use GraphQL\Type\Definition\Type;
 use SilverStripe\ORM\Queries\SQLUpdate;
+use GraphQL\Type\Definition\ResolveInfo;
+use SilverStripe\GraphQL\MutationCreator;
+use DNADesign\Elemental\Models\BaseElement;
+use SilverStripe\GraphQL\OperationResolver;
+use DNADesign\Elemental\Services\ReorderElements;
```
robbieaverill: As an aside, it'd be cool if your IDE could sort these alphabetically to reduce noise in future PRs from people whose IDEs do sort them

src/GraphQL/SortBlockMutationCreator.php
```diff
@@ -40,73 +41,17 @@ public function args()
...
-           ->addWhere([$filter, "\"$tableName\".\"ParentID\"" => $parentId]);
-
-        $query->execute();
-
-        // Now use the ORM to write a new version of the record that we are directly reordering
-        $block->Sort = $newBlockPosition;
-        $block->write();
-
-        return $block;
+        $reorderingService = new ReorderElements($element);
```
robbieaverill: Use Injector to get this class 👍

src/Services/ReorderElements.php
```diff
@@ -0,0 +1,93 @@
...
+use SilverStripe\ORM\DataObject;
+use SilverStripe\ORM\Queries\SQLUpdate;
+use DNADesign\Elemental\Models\BaseElement;
+
+class ReorderElements
+{
+    /**
+     * @var BaseElement $block
+     */
+    private $element;
```
robbieaverill: We always prefer protected over private

src/Services/ReorderElements.php
```diff
@@ -0,0 +1,93 @@
...
+use InvalidArgumentException;
+use SilverStripe\Core\Convert;
+use SilverStripe\ORM\DataObject;
+use SilverStripe\ORM\Queries\SQLUpdate;
+use DNADesign\Elemental\Models\BaseElement;
+
+class ReorderElements
+{
+    /**
+     * @var BaseElement $block
```
robbieaverill: `$block` is wrong, but you don't actually need it here - that's the description of a `@var` tag

src/Services/ReorderElements.php
```diff
@@ -0,0 +1,93 @@
...
+     */
+    private $element;
+
+    public function __construct(BaseElement $element)
+    {
+        if (!is_subclass_of($element, BaseElement::class)) {
+            throw new InvalidArgumentException(sprintf(
+                'Invalid ' . BaseElement::class . ' passed to ' . __CLASS__
```
robbieaverill: In the new mutation you're throwing an exception which contains the passed class name, do you want to use the same thing here? I.e. `get_class($element) . ' is not an instance of BaseElement` or whatever?

src/Services/ReorderElements.php
```diff
@@ -0,0 +1,93 @@
...
+            ));
+        }
+
+        if (!$element->canEdit()) {
```
robbieaverill: We're missing the context here, this will have problems with using non-session based authentication since the member is always passed in the `$context` args

src/Services/ReorderElements.php
```diff
@@ -0,0 +1,93 @@
...
+use DNADesign\Elemental\Models\BaseElement;
+
+class ReorderElements
+{
+    /**
+     * @var BaseElement $block
+     */
+    private $element;
+
+    public function __construct(BaseElement $element)
```
robbieaverill: Needs a doc block

src/Services/ReorderElements.php
```diff
@@ -0,0 +1,93 @@
...
+                'Changing the sort order of blocks is not allowed for the current user'
+            );
+        }
+
+        $this->element = $element;
+    }
+
+    /**
+     * @param
+     */
```
robbieaverill: Needs an update

src/Services/ReorderElements.php
```diff
@@ -0,0 +1,93 @@
...
+            ));
+        }
+
+        if (!$element->canEdit()) {
+            throw new InvalidArgumentException(
+                'Changing the sort order of blocks is not allowed for the current user'
+            );
+        }
+
+        $this->element = $element;
```
robbieaverill: Do you want to add a getter and setter for this?

src/Services/ReorderElements.php
```diff
@@ -0,0 +1,93 @@
...
+
+        if ($elementToBeAfterID) {
+            /** @var BaseElement $afterBlock */
+            $afterElement = BaseElement::get()->byID($elementToBeAfterID);
+
+            if (!$afterElement) {
+                throw new InvalidArgumentException(sprintf(
+                    '%s#%s not found',
+                    BaseElement::class,
+                    $parentId
```
robbieaverill:
```diff
- $parentId
+ $elementToBeAfterID
```

src/Services/ReorderElements.php
```diff
@@ -0,0 +1,93 @@
...
+        } else {
+            $operator = '-';
+            $filter = "\"$tableName\".\"Sort\" <= $sortAfterPosition AND \"$tableName\".\"Sort\" > $currentPosition";
+            $newBlockPosition = $sortAfterPosition;
+        }
+
+        $query = SQLUpdate::create()
+        ->setTable("\"$tableName\"")
+        ->assignSQL('"Sort"', "\"$tableName\".\"Sort\" $operator 1")
+        ->addWhere([$filter, "\"$tableName\".\"ParentID\"" => $parentId]);
```
robbieaverill: Indentation is out (might be causing the failing builds)

tests/GraphQL/AddElementToAreaMutationTest.php
```diff
@@ -0,0 +1,92 @@
...
+    protected static $fixture_file = 'AddElementToAreaMutationTest.yml';
+
+    protected static $extra_dataobjects = [
+        TestElement::class,
+    ];
+
+    protected function setUp()
+    {
+        parent::setUp();
+        ElementalArea::create()->write();
```
robbieaverill: Could this go into a fixture file instead?

tests/GraphQL/AddElementToAreaMutationTest.php
```diff
@@ -0,0 +1,92 @@
...
+use InvalidArgumentException;
+use SilverStripe\Dev\SapphireTest;
+use GraphQL\Type\Definition\ResolveInfo;
+use DNADesign\Elemental\Models\ElementalArea;
+use DNADesign\Elemental\Tests\Src\TestElement;
+use DNADesign\Elemental\GraphQL\AddElementToAreaMutation;
+
+class AddElementToAreaMutationTest extends SapphireTest
+{
+    protected static $fixture_file = 'AddElementToAreaMutationTest.yml';
```
robbieaverill: Your fixture file is empty - I realise this is because you've reported having problems with `$usesDatabase = true`, but maybe document it. I've never had a problem with that property, so at some point somebody is going to wonder why this is here and remove it =)

tests/Services/ElementTabProviderTest.php
```diff
@@ -0,0 +1,29 @@
...
+
+use SilverStripe\Dev\SapphireTest;
+
+class ElementTabProviderTest extends SapphireTest
+{
+    protected function setUp()
+    {
+        parent::setUp();
+        $this->markTestIncomplete('Does not yet test anything, remains to be implemented');
+    }
```
robbieaverill: I'd say remove this rather than adding incomplete test classes

tests/Services/ReorderElementsTest.php
```diff
@@ -0,0 +1,44 @@
...
+
+namespace DNADesign\Elemental\Tests\Services;
+
+use SilverStripe\Dev\SapphireTest;
+use DNADesign\Elemental\Tests\Src\TestElement;
+use DNADesign\Elemental\Services\ReorderElements;
+
+class ReorderElementsTest extends SapphireTest
+{
+    protected static $fixture_file = '../GraphQL/SortBlockMutationCreatorTest.yml';
```
robbieaverill: -1 for coupling test fixtures, can you copy it so this class can own it?

src/Services/ReorderElements.php
NightJar: Ah cool, didn't realise ta

src/GraphQL/SortBlockMutationCreator.php
NightJar: Interesting note: I have sorted them.
But weirdly this is the order. Rather than alphabetically (which I also expected) it seems to go in triangle form, shortest to longest... which is just... weird.

tests/GraphQL/AddElementToAreaMutationTest.php
NightJar: I think if I combine it with the previous suggestion we might knock two off the list with one change :)

tests/Services/ElementTabProviderTest.php
NightJar: I've implemented this as I noticed it was not tested - more as a beginning and something that will constantly be of note until it becomes a thing. I should raise an issue for it (it's not fleshed out as it is not the focus of this work)

src/GraphQL/SortBlockMutationCreator.php
NightJar: OK sweet, there's a setting to make it sort alphabetically. But it does it by FQCN, not just classname... I hope that's OK :)

tests/Services/ElementTabProviderTest.php
NightJar: https://github.com/dnadesign/silverstripe-elemental/issues/431

NightJar:
How rude.
```
[vagrant@archlinux inlineaddelement]$ vendor/bin/phpunit -v -d memory_limit=512M vendor/dnadesign/silverstripe-elemental/tests/GraphQL/ '' flush=1
PHPUnit 5.7.27 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.2.10
Configuration: /srv/http/inlineaddelement/phpunit.xml.dist

....                                                                4 / 4 (100%)

Time: 19.31 seconds, Memory: 104.25MB

OK (4 tests, 11 assertions)
[vagrant@archlinux inlineaddelement]$ head -n3 .env
# DB credentials
SS_DATABASE_CLASS="PostgreSQLDatabase"
SS_DATABASE_SERVER="localhost"
```

tests/GraphQL/AddElementToAreaMutationTest.php
```diff
@@ -0,0 +1,86 @@
...
+        $args = [
+            'ClassName' => $className,
+            'ElementalAreaID' => $elementalAreaID,
+        ];
+
+        if (!is_null($afterElementId)) {
+            $args['AfterElementID'] = $afterElementId;
+        }
+
+        $mutation->resolve(null, $args, $context, $resolveInfo);
```
ScopeyNZ: You will probably have to have a way of returning IDs here so that you can compare against them in your continued tests. Alternatively you can seed with a fixture and specify IDs. You can't 100% rely on IDs being ascending order from 1 in functional tests.

NightJar: Sorry I don't understand what you're trying to tell me.

ScopeyNZ: You add a block with the mutation, then assume the ID of that block is `1`. It would be best practice to get the ID from the result of the `resolve` method and use that as the source of truth for the ID rather than assuming it's going to be `1`.

ScopeyNZ: FYI this is why the "SortBlockMutation" test fixture specified IDs.

robbieaverill: Yeah I think I agree. Using hard coded IDs in tests is bad and introduces unnecessary tech debt that we'll have to fix in future. FYI there's discussions in core at the moment about abstracting the primary key assignment for models so UUIDs would become the default - pieces of code like this are the main reason we're talking about not doing it in 4.x at the moment because it'll break people's tests (potentially)

ScopeyNZ: FYI we do set IDs in the fixtures for SortBlocks. I don't think seeding the IDs in that manor is particularly bad? Even if the IDs switch to UUID manually setting the IDs to int would probably still work?

robbieaverill: I guess the better option is to assert orders based on some piece of fixed data like a model's Title or whatever

NightJar: OK, I can do this.

However I'd like to also point out that while I can agree that one should code to an abstraction rather than an implementation, in order to switch the IDs out to e.g. UUID there would need to be much more work than simply altering these tests - all ID's are currently typed as Integers, and GraphQL is strongly typed. It is not simply tests that would break.

robbieaverill: I think this conversation needs to head over to the RFC https://github.com/silverstripe/silverstripe-framework/issues/8411

NightJar:
OK so I had to reluctantly loosen the assertions for now, but the tests now all pass.

src/Services/ReorderElements.php
ScopeyNZ: +1 this. Private makes it unnecessarily difficult to extend.

ScopeyNZ:
There's a few of Robbie's comments on `Services/ReorderElements.php` that look valid to me.

NightJar:
Weird, they don't show up in the review log for me :(
Cheers, I'll take a look

src/Services/ReorderElements.php
NightJar: Updated :)

src/Services/ReorderElements.php
NightJar: Added

src/Services/ReorderElements.php
NightJar: Done

PR #432, opened 2018-10-04
NEW Treat space key press as mouse click

### LINKED ISSUE #423: Element not keyboard accessible in IE and Edge
Related to the ongoing feature development of the ElementalEditor on `dev-master`:

In the CMS, individual elements within the list are not expandable via keyboard/tabbing. Whilst the element is clickable when in focus, in IE and Edge - via BrowserStack at least - ,this triggers a save action of the entire page instead of the expanding the block.

raissanorth:
Resolves #423 by treating a press of the space key like a button click. This causes the element to expand. Works on IE and Edge and therefore makes the element accessible. 🙂 

On press of the enter key the form gets 'submitted' and the page is saved.


sminnee:
@ScopeyNZ or @robbieaverill could we remove type/enhancement from this?

ScopeyNZ:
> @ScopeyNZ or @robbieaverill could we remove type/enhancement from this?

Done.

PR #433, opened 2018-10-05
Pulls/4.0/layout annoIEnces

NightJar:
fix layout issues in IE & Edge


robbieaverill:
Changes LGTM but I'd like to test this in other browsers before merging

NightJar:
Tests were run in chrome, safari, ie11, edge... and I think firefox (but not by me - @raissanorth I think you did though right?)

NightJar:
Here is a picture of:
```
Edge | Chrome
---------------
IE11 | Firefox
```

image

Previously the plus icon was below the bar on edge, and the line was higher than the edges of two elements, and overlays the top one (and the plus icon is a bit out):

image

ScopeyNZ:
Seems like you guys have the bases covered...

PR #437, opened 2018-10-09
WIP NEW GraphQL for elements is now per block and the client side scaffolds the query

### LINKED ISSUE #304: Implement a block registry so GraphQL queries can be dynamically structured
With #284 we added a "SummaryData" field to the block node in GraphQL. This exposes an arbitrary dataset defined by a `getSummaryData` method on `BaseElement`. We need to convert this to a structured GraphQL query utilising fragments:

```GraphQL
query readBlocks {
  readBlock {
    edges {
      node {
        ID
        Title
        ... on ContentBlock {
          Summary
        }
        ... on ImageBlock {
          Thumbnail
        }
      }
    }
  }
}
```

The tricky part is that this needs to be scaffolded correctly on the PHP side and the query needs to be dynamic in React.

Suggestions:
- Create a registry on the client side (and probably the PHP side - at least for consistency) for block authors to opt-in their custom blocks
- PHP can auto-scaffold fields given the opted in blocks
- Elements can indicate additional fields (or methods) to expose
- Generate a GraphQL query from registered blocks. This might use a [Query component](https://www.apollographql.com/docs/react/essentials/queries.html#basic). 
- Require blocks to define a react component that displays a summary - We could also define a default component for non-editable blocks as the design only relies on attributes which are present on all elements.

### LINKED ISSUE #387: Change "readBlocksForPage" to "readBlocksForArea"
Currently we have a GraphQL query which is `getBlocksForPage`. This is a bit awkward because blocks can be separated into multiple areas, and pages don't necessarily have elemental areas. A better solution would be to expose the actual relationship that exists - the elemental area.

We should convert our query to `getBlocksForArea`:

```gql
query ReadBlocksForArea($id:ID!) {
  readOneArea(ID: $id, Versioning: {
    Mode: DRAFT
  }){
    ID
    Elements {
      pageInfo {
        totalCount
      }
      edges {
        node {
          ID
          Title
          BlockSchema
          InlineEditable
          IsLiveVersion
          IsPublished
          Version
        }
      }
    }
  }
}
```

This also removes the awkward `ElementalAreaIfExists`.

After this, the components will need to be updated to deal with area IDs instead of page IDs.

ScopeyNZ:
Ok - Got a little carried away with this one.

This aims to fix #304 . 

It also addresses:
- fixes #387 - The query is now `readElementsForArea`
- Having a simple way for authors to implement a react component to generate a summary
- Renaming block -> element where relevant

I still need to write/fix tests and various clean-up which will I'll try to finish off this week at some point.


client/src/components/ElementSummary/Default.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
+export default () => (
+  <div className="element-editor-summary element-editor-summary--default">
+    <p className="element-editor-summary__default">
+      There is no preview available for this block
```
robbieaverill: This needs to be translatable =)

ScopeyNZ:
I'm closing this and splitting it into some more targetted PRs.

PR #439, opened 2018-10-11
NEW All unsaved blocks on a page will now save as the page is saved

### LINKED ISSUE #328: Save all blocks on page save
In-line edited blocks can be saved individual but should be saved (en masse) when a page save is triggered. This would involve either looping through all "dirty" blocks and triggering a save (asynchronously) or creating some form of aggregation tool for multiple `FormBuilderLoader` forms.

I'm not entirely sure on the best approach here.

A/Cs
- [ ] Saving a page with an elemental area will also save all edits to blocks
- [ ] Blocks that have been edited but collapsed without saving are still saved
- [ ] Other page elements save as expected 

Split from #295

ScopeyNZ:
Fixes #328 . This works by syncing redux state with a hidden field. The content of the hidden field is then used to write back into the various elements within an elemental area.

Relies on silverstripe/silverstripe-admin#682

Todo:
- [x] Use some API provided by @raissanorth's recent PR
- [x] Behat test
- [ ] See if I can do some optimistic cache updates and remove the nuclear option (https://github.com/creative-commoners/silverstripe-elemental/blob/6e73b50d22e01ddc0088ae5c4d160040a4976bd7/client/src/legacy/ElementEditor/entwine.js#L41)


ScopeyNZ:
Someone might be able to provide a good solution to optimistic cache updates. If I can get the `CMSMain::save` response to return some custom data I'm pretty confident I can do it, but it doesn't seem extensible enough. Open to ideas.

src/Forms/ElementalAreaField.php
```diff
@@ -208,4 +211,48 @@ public function performReadonlyTransformation()
             ->setName($this->getName())
             ->addExtraClass('elemental-area--read-only');
     }
+
+    public function setSubmittedValue($value, $data = null)
+    {
+        // Content comes through as a JSON encoded list through a hidden field.
+        $this->setValue(Convert::json2array($value));
```
robbieaverill: It'd be good to return `$this` for chaining, and also these methods could use doc blocks

src/Forms/ElementalAreaField.php
```diff
@@ -208,4 +211,48 @@ public function performReadonlyTransformation()
...
+    public function saveInto(DataObjectInterface $dataObject)
+    {
+        /** @var BlocksPage $dataObject */
+        parent::saveInto($dataObject);
+
+        $elementData = $this->Value();
+
+        foreach ($elementData as $form => $data) {
+            // Extract the ID
+            $elementId = (int) substr($form, 12);
```
robbieaverill: Magic numbers!

ScopeyNZ:
Okay the Behat tests _should_ pass once these two dependant PRs are merged:

- [ ] https://github.com/silverstripe/silverstripe-behat-extension/pull/184
- [ ] https://github.com/silverstripe/silverstripe-admin/pull/682

tests/Behat/Context/FeatureContext.php
```diff
@@ -282,4 +363,30 @@ protected function getCaretButton($block)
...
+    /**
+     * @param $block
+     * @param $name
+     * @return mixed
+     */
+    protected function findFieldInBlock($block, $name)
+    {
+        $label = $block->findAll('xpath', sprintf('//label[contains(text(), \'%s\')]', $name));
+
+        assertNotNull($label, sprintf('Could not find a label for a field with the content "%s"', $name));
```
raissanorth: In the case of `Title` I get the following issue with this:
`Found more than one label containing the phrase "Title".
      Failed asserting that actual size 0 matches expected size 1.`

The HTML looks as follows: 
`<label for="Form_ElementForm_11_PageElements_11_Title" class="form__field-label ">Title (displayed if checked)</label>`
and in the feature file I've got 
`And I fill in "Charlie's Block" for "Title" for block 2`

raissanorth: Okay, I found the mistake I made. This works as expected, but the message is wrong. Should assert the count instead.

client/src/legacy/ElementEditor/entwine.js
```diff
@@ -28,6 +31,15 @@ jQuery.entwine('ss', ($) => {
 
     onunmatch() {
       ReactDOM.unmountComponentAtNode(this[0]);
-    }
+    },
+
+    /**
+     * Invalidate cache after the form is submitted to force apollo to re-fetch.
+     */
+    'from .cms-edit-form': {
```
robbieaverill: Typo? `form`?

src/Forms/ElementalAreaField.php
```diff
@@ -208,4 +212,55 @@ public function performReadonlyTransformation()
             ->setName($this->getName())
             ->addExtraClass('elemental-area--read-only');
     }
+
+    /**
+     * {@inheritdoc}
+     */
```
robbieaverill: This is the same as not having a doc block at all, you may as well leave it out and save the lines

src/Forms/ElementalAreaField.php
```diff
@@ -208,4 +212,55 @@ public function performReadonlyTransformation()
...
+    public function setSubmittedValue($value, $data = null)
+    {
+        // Content comes through as a JSON encoded list through a hidden field.
+        return $this->setValue(Convert::json2array($value));
```
robbieaverill: FYI we've agree to deprecate these JSON handling methods, from now on we may as well just use regular PHP json methods. RFC: https://github.com/silverstripe/silverstripe-framework/issues/8424

src/Forms/ElementalAreaField.php
```diff
@@ -208,4 +212,55 @@ public function performReadonlyTransformation()
...
+    /**
+     * {@inheritdoc}
+     */
+    public function saveInto(DataObjectInterface $dataObject)
+    {
+        /** @var BlocksPage $dataObject */
+        parent::saveInto($dataObject);
+
+        $elementData = $this->Value();
+        $idPrefixLength = strlen(sprintf(ElementalAreaController::FORM_NAME_TEMPLATE, ''));
```
robbieaverill: When I read this I thought "I wonder if I'd prefer to read it as two variables; `$idPrefix` and `$idPrefixLength`", then I ignored myself until I saw the same pattern used on lines 250-251. The difference is that there both of those variables are used, whereas here they aren't. I'll just leave this here as food for thought (not a request necessarily)

client/src/legacy/ElementEditor/entwine.js
ScopeyNZ: I thought so too buuuut I copied from admin - it appears this is some entwine magic - as in if this event comes __from__ `.cms-edit-form`.

src/Forms/ElementalAreaField.php
ScopeyNZ: > It'd be good to return `$this` for chaining, and also these methods could use doc blocks

😆 I'll remove them again 🙂

src/Forms/ElementalAreaField.php
ScopeyNZ: Yeah. I usually try to limit variable declaration if it's not going to be used unless it's going to make the code super messy. In this case I opted for a variable because it's better to leave this stuff out of the loop.

client/src/legacy/ElementEditor/entwine.js
robbieaverill: 👍 TIL

src/Forms/ElementalAreaField.php
robbieaverill: Love me a transparent doc block

PR #443, opened 2018-10-15
Pulls/4.0/transmutation

### LINKED ISSUE #294: Add block to page (inline)

### LINKED ISSUE #340: Add a block when clicking a button in the "Add Block" popover

NightJar:
Resolves #340 
Resolves #294 
Allows a user to add an element without having to browse to a new form in order to fill in default information. This allows speedier content entry and reinforces inline editing as the standard way to edit elements.


client/src/components/ElementEditor/AddElementPopover.js
```diff
@@ -3,6 +3,7 @@ import { Button, Input, InputGroup, InputGroupAddon, Popover } from 'reactstrap'
 import classNames from 'classnames';
 import { elementTypeType } from 'types/elementTypeType';
 import i18n from 'i18n';
+import addElementMutation from 'state/editor/addElementMutation';
```
ScopeyNZ: We normally apply these HOC's with Injector. This allows people to sub out this query for their own if need be.

client/src/components/ElementEditor/AddNewButton.js
```diff
@@ -43,8 +43,8 @@ class AddNewButton extends Component {
           target={buttonAttributes.id}
           isOpen={this.state.popoverOpen}
           elementTypes={elementTypes}
-          baseAddHref={baseAddHref}
           toggle={this.toggle}
+          elementalAreaId={elementalAreaId}
```
ScopeyNZ: The `insertAfterElement` prop is required but not set here. Additionally I think we want the default location for new blocks to be at the top so it should be `'0'`?

client/src/components/ElementEditor/ElementEditor.js
```diff
@@ -22,7 +35,7 @@ class ElementEditor extends PureComponent {
 ElementEditor.propTypes = {
   elementTypes: PropTypes.arrayOf(elementTypeType).isRequired,
   pageId: PropTypes.number.isRequired,
-  baseAddHref: PropTypes.string.isRequired,
+  elementalAreaId: PropTypes.oneOfType([PropTypes.number, PropTypes.string]).isRequired,
```
ScopeyNZ: It'd be nice to have some better type safety on this prop.

client/src/components/ElementEditor/Header.scss
```diff
@@ -47,10 +47,6 @@
     width: unset;
   }
 
-  &__version-state {
-    box-shadow: $gallery-item-shadow;
-  }
-
```
ScopeyNZ: Wasn't this shadow part of the designs?

client/src/components/ElementEditor/HoverBar.js
```diff
@@ -97,7 +97,7 @@ class HoverBar extends Component {
       'element-editor__add-block-hover-bar',
       {
         'element-editor__add-block-hover-bar--delay': !calledInstantaneously,
-        'element-editor__add-block-hover-bar--inst': calledInstantaneously,
+        'element-editor__add-block-hover-bar--inst': calledInstantaneously
```
ScopeyNZ: This change isn't required - it's arguably nicer to leave trailing commas if they're supported - smaller diffs means less chances to conflict!

client/src/components/ElementEditor/HoverBar.js
```diff
@@ -117,52 +117,55 @@ class HoverBar extends Component {
...
+  elementalAreaId: PropTypes.oneOfType([PropTypes.number, PropTypes.string]).isRequired
 };
 export { HoverBar as Component };
 
 export default inject(
   ['AddElementPopover'],
-  (AddElementPopoverComponent) => ({
-    AddElementPopoverComponent,
+  AddElementPopoverComponent => ({
+    AddElementPopoverComponent
```
ScopeyNZ: Trailing comma should be left in.

client/src/state/editor/addElementMutation.js
```diff
@@ -0,0 +1,42 @@
...
+mutation AddBlock($className: String!, $elementalAreaID: ID!, $afterElementID: ID) {
+  addElementToArea(
+    ClassName: $className,
+    ElementalAreaID: $elementalAreaID,
+    AfterElementID: $afterElementID
+  ) {
+    ID,
+    Sort,
+    IsPublished,
+    IsLiveVersion,
```
ScopeyNZ: These commas aren't needed. I actually thought GraphQL kicks up a fuss but apparently not 🤔 . Additionally you're doing a refetch so you can technically omit all return values. 

Alternatively you can update the return values to match `readBlocksForPage` and then you can do an update on the store after the values are returned. See the sort mutation: https://github.com/dnadesign/silverstripe-elemental/blob/master/client/src/state/editor/sortBlockMutation.js#L28. The `optimisticResponse` is not relevant because you can't assume the response, but you should be able to splice your item into the known list of blocks.

Essentially here you're reading the cached response of a GraphQL query and updating that response. Apollo will automatically update consumers of that query.

This will drop a request to the server (the refetch of the query)

client/src/state/editor/addElementMutation.js
```diff
@@ -0,0 +1,42 @@
+import { graphql } from 'react-apollo';
+import gql from 'graphql-tag';
+
+// GraphQL query for deleting a specific block
+const mutation = gql`
+mutation AddBlock($className: String!, $elementalAreaID: ID!, $afterElementID: ID) {
```
ScopeyNZ: This mutation should be called `AddElement` or maybe `AddElementToArea`.

client/src/components/ElementEditor/AddNewButton.js
NightJar: Not required, but could have a default. @clarkepaul - can you please confirm whether new blocks should be added at first, or last?

client/src/components/ElementEditor/ElementEditor.js
NightJar: I'd be nice to have return types be consistent, yes. I may be confusing my JS and PHP here though, is it _always_ an Int in React? I've seen errors in both systems.

client/src/components/ElementEditor/Header.scss
NightJar: It's removed due to linter taking issue with the equivalent definition immediately below the removed block.

This block also has a `box-shadow`.

If one is more correctorer than the other, it would be great to know which is which. Is @newleeland able to tell from the definitions?

client/src/components/ElementEditor/HoverBar.js
NightJar: It'd be great if usage was consistent throughout the system too, and this change is trying to bring things in line with existing code (particularly where e.g. my linter formatter goes and makes edits that are apparently contrary to someone else's...). Is there any guidance on this I'm particularly unaware of?

client/src/state/editor/addElementMutation.js
NightJar: Yes, it should.

client/src/components/ElementEditor/AddNewButton.js
NightJar: Ideal ta, I'll set a defaultProps... err, property :)

client/src/components/ElementEditor/AddElementPopover.js
robbieaverill: To be fair, this pattern is how we've done it before, but we _should_ be aiming to make them injectable where possible. Related: https://github.com/silverstripe/silverstripe-admin/issues/611

client/src/components/ElementEditor/AddElementPopover.js
```diff
@@ -17,12 +18,38 @@ class AddElementPopover extends Component {
...
+    this.handleKeyDown = this.handleKeyDown.bind(this);
+    this.getElementButtonClickHandler = this.getElementButtonClickHandler.bind(this);
 
     this.state = {
       searchValue: ''
     };
   }
 
+  /**
+   * click handler that preserves the details of what was clicked
```
robbieaverill: Can you document `elementType` and the return please?

client/src/components/ElementEditor/ElementEditor.js
```diff
@@ -8,12 +8,25 @@ import { elementTypeType } from 'types/elementTypeType';
...
-        <ListComponent elementTypes={elementTypes} pageId={pageId} baseAddHref={baseAddHref} />
+        <ToolbarComponent
+          elementTypes={elementTypes}
+          elementalAreaId={elementalAreaId}
+        />
+        <ListComponent
+          elementTypes={elementTypes}
+          pageId={pageId}
+          elementalAreaId={elementalAreaId}
+        />
```
robbieaverill: I generally think that these kinds of changes are unnecessary and pollute the purpose of a PR - maybe in future we could keep them to separate PRs

client/src/components/ElementEditor/ElementEditor.js
robbieaverill: We can cast it before we inject the prop if you want to just make it `number`. IDs in SilverStripe can be strings or ints in different environments =D

Actually if this comes from a GraphQL query then we may not be able to do this gracefully. This is what the GraphQL schema says about the ID type:

>The ID type appears in a JSON response as a String; however, it is not intended to be human-readable. When expected as an input type, any string (such as "4") or integer (such as 4) input value will be accepted as an ID.

client/src/components/ElementEditor/Header.scss
robbieaverill: Might've been a merge conflict thing, makes sense. Presumably the change you're removing is not used anyway

client/src/components/ElementEditor/HoverBar.js
robbieaverill: Linters that automatically change existing code are very dangerous!! I've wasted a lot of time with peer reviews that have boiled down to automated changes. We should keep this kind of change to separate PRs (as mentioned above), where we can say "OK, it's automated and all LGTM so merge"

client/src/components/ElementEditor/HoverBar.js
```diff
@@ -117,52 +117,55 @@ class HoverBar extends Component {
    * @returns {DOMElement}
    */
   render() {
-    const { AddElementPopoverComponent, elementTypes, baseAddHref, elementId } = this.props;
+    const {
+      AddElementPopoverComponent,
+      elementTypes,
+      elementId,
+      elementalAreaId
```
robbieaverill: Here'd be good to add a trailing comma so this type of diff doesn't re-occur

client/src/components/ElementEditor/HoverBar.js
```diff
@@ -117,52 +117,55 @@ class HoverBar extends Component {
...
+    return (
+      <div className="element-editor__add-block-area-container" id={`AddBlockArea_${elementId}`}>
+        <div
+          className="element-editor__add-block-area"
+          onMouseEnter={this.handleMouseEnter}
+          onMouseLeave={this.handleMouseLeave}
+        >
+          {/* render the hover bar in the corresponding style */}
+          {(this.state.delayAreaActive && this.renderHoverBar(false)) ||
+            (this.state.instAreaActive && this.renderHoverBar(true))}
```
robbieaverill: I think I preferred it the way it was =/

src/Forms/ElementalAreaField.php
```diff
@@ -111,8 +111,11 @@ public function FieldHolder($properties = array())
     public function getSchemaDataDefaults()
     {
         $schemaData = parent::getSchemaDataDefaults();
-        $pageId = ($this->getArea() && ($page = $this->getArea()->getOwnerPage())) ? $page->ID : null;
+
+        $area = $this->getArea();
+        $pageId = ($area && ($page = $area->getOwnerPage())) ? $page->ID : null;
         $schemaData['page-id'] = $pageId;
+        $schemaData['elemental-area-id'] = $area ? $area->ID : null;
```
robbieaverill: Right so if ElementEditor gets the ID from here we can definitely cast as int

client/src/components/ElementEditor/ElementEditor.js
NightJar: Tests will fail because of line length if I don't ;)

client/src/components/ElementEditor/ElementEditor.js
NightJar: I'm not certain of the outcome on this comment :/

client/src/components/ElementEditor/ElementEditor.js
robbieaverill: I see, you've added a new prop to it - disregard

client/src/components/ElementEditor/ElementEditor.js
robbieaverill: If you can control the input (e.g. from entwine hydration) then cast it to a number and make this strict, if you can't control it then leave it like this =)

client/src/components/ElementEditor/HoverBar.js
NightJar: Hoverbar.js seems to be the one with all the auto-format crazyness - I'll tidy up the noise :)

client/src/components/ElementEditor/ElementEditor.js
NightJar: Yes, it is an annoyance of this type of change, I agree. It's hard to quickly see what has changed - but the linter be as the linter do! (except where its not defined about trailing commas... :/ )

client/src/components/ElementEditor/AddNewButton.js
NightJar: Went to passing in `0` explicitly instead, because setting a prop to `undefined` is difficult, and I'm unsure the impact `null` will have on the GraphQL query, since it's a valid type (but unexpected by the mutation).

client/src/state/editor/addElementMutation.js
NightJar: So turns out commas are OK.
So are line returns.
So are spaces.
So are combinations of all of the above 😑

robbieaverill: They might be OK but in lieu of a GraphQL query "coding standard" we should at least aim for uniformity among the queries in core and the wider ecosystem (no commas)

client/src/components/ElementEditor/AddNewButton.js
NightJar: Done, no auto resolve as it's the next line down

src/Forms/ElementalAreaField.php
NightJar: related proptypes are all `number.isRequired` now :)

client/src/components/ElementEditor/AddElementPopover.js
NightJar: Was implemented as directed, however have updated it anyway as I also think it's a good idea.

client/src/state/editor/addElementMutation.js
NightJar: OK, so we looked at this together, however as the `pageId` is required we (@ScopeyNZ and I) thought it better to wait for the refactor from `getElementsForPage` to become `getElementsForArea` - we already have the `ElementalAreaID`, and re-inserting the PageID to filter all the way through all areas seems... unnecessary (read: tech debt that we'd have to remove later).

NightJar:
Final behat test failure appears unrelated - working with @ScopeyNZ to resolve (over multiple development branches).

NightJar:
GREEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEN 💚 

💜 ?

robbieaverill:
@NightJar I merged another PR which it'd be good for you to resolve the conflicts on and re-test, I think it was #439 which also changes from pulling blocks for a page to pulling blocks for a page's area. I approve the PR as it stands, and am happy for you to self merge it in the morning once it's green.

NightJar:
Cool, it'll be needing a rebase since another merge - I'll do that then merge when green :)

NightJar:
Neat, thanks team :)

PR #447, opened 2018-10-16
Stop inheritance for disallowed elements

ohararyan:
In the config we can set a flag to stop element inheritance - this currently only works for allowed_elements but should also work for disallowed_elements. 

This allows the following example to work correctly:

```yml
Site\PageTypes\Page:
  extensions:
    - DNADesign\Elemental\Extensions\ElementalPageExtension
  disallowed_elements:
    - Site\Elements\ElementAllowedOnSpecialPage

Site\PageTypes\SpecialPage:
  stop_element_inheritance: true
```

### DIFF
--- a/src/Extensions/ElementalAreasExtension.php
+++ b/src/Extensions/ElementalAreasExtension.php
@@ -81,7 +81,11 @@ class ElementalAreasExtension extends DataExtension
             $availableClasses = ClassInfo::subclassesFor(BaseElement::class);
         }
 
-        $disallowedElements = (array) $config->get('disallowed_elements');
+        if ($config->get('stop_element_inheritance')) {
+            $disallowedElements = (array) $config->get('disallowed_elements', Config::UNINHERITED);
+        } else {
+            $disallowedElements = (array) $config->get('disallowed_elements');
+        }
         $list = array();
 
         foreach ($availableClasses as $availableClass) {


src/Extensions/ElementalAreasExtension.php
```diff
@@ -81,7 +81,11 @@ public function getElementalTypes()
             $availableClasses = ClassInfo::subclassesFor(BaseElement::class);
         }
 
-        $disallowedElements = (array) $config->get('disallowed_elements');
+        if ($config->get('stop_element_inheritance')) {
```
robbieaverill: I think it'd be nice to use a positive form of config rather than a negative, e.g. `inherit_allowed_elements: true` by default and disable it to prevent. Thoughts?

ohararyan: I've just kept with the existing config that's set so as not to cause any issues where that's already being used. To me it doesn't feel like it needs another config variable as these should do the same thing.

ohararyan:
Not sure what's happened here with this PR - some how have managed to try and merge it into the 3 branch and have now subsequently got all of those commits in here. Happy to open this as another PR and clean that up

robbieaverill:
@ohararyan it just needs a good rebase onto your target branch =)

src/Extensions/ElementalAreasExtension.php
robbieaverill: Sorry, I thought you were adding a new config prop - all good!

ohararyan:
@robbieaverill really odd it appears to be all good - if I compare here https://github.com/dnadesign/silverstripe-elemental/compare/master...ohararyan:pr/stop-inheritance-for-disallowed-elements this shows I'm one commit above master.

robbieaverill:
Yeah you're right. I wonder if it's a GitHub glitch...

robbieaverill:
I rebased onto 3 then back again onto master, seems ok now

PR #449, opened 2018-10-17
NEW Don't render InlineEditForm on loadingError

raissanorth:
Also make ElementalAreaController inherit from CMSMain

Change in this PR:

- Implement an onLoadingError method to be passed to FormBuilderLoader. This let's us know in InlineEditForm is an error has occurred and take necessary action, such as preventing the Element component from expanding the Content. When a loading error occurs the InlineLineEditForm cannot be rendered, this would cause the Summary component to be toggled erroneously.

- It makes more sense for ElementalAreaController to inherit from CMSMain rather than LeftAndMain, as this means that permissions are checked in relation to the Page Level rather than for the entire CMS.


client/src/components/ElementEditor/InlineEditForm.js
```diff
@@ -1,11 +1,45 @@
...
+
+    this.setState({
+      loadingError: true,
+    });
+
+    $.noticeAdd({
+      text: i18n.inject(
+        i18n._t(
+          'EditForm.ERROR_NOTIFICATION',
+          'Error rendering the edit form for this block'),
```
robbieaverill: I think that "rendering" is maybe a dev focussed term. Maybe we could use "loading" instead?

client/src/components/ElementEditor/InlineEditForm.js
```diff
@@ -1,11 +1,45 @@
...
+        ),
+        stay: false,
```
robbieaverill: I'm still not sold on using a toast message for this. It will disappear quickly and people might miss it, look back at their screen and not know why something is broken. Can you convince me otherwise? =D

client/src/components/ElementEditor/InlineEditForm.js
```diff
@@ -1,11 +1,45 @@
...
+import i18n from 'i18n';
 
 class InlineEditForm extends PureComponent {
+  constructor(props) {
+    super(props);
+
+    this.handleLoadingError = this.handleLoadingError.bind(this);
+
+    this.state = {
+      loadingError: null
```
robbieaverill: I'm curious about whether we can use one of the injected GraphQL props (loading or error) from the Apollo query rather than defining our own state for it: https://github.com/dnadesign/silverstripe-elemental/blob/master/client/src/state/editor/readBlocksForPageQuery.js#L46-L48

These would be available as props on the component that the query is bound to

PR #450, opened 2018-10-17
NEW Implement drag and drop for elements

### LINKED ISSUE #285: Implement React based drag and drop for reordering elements in the editor

ScopeyNZ:
This adds drag and drop functionality to the new elemental editor

Fixes #285


client/src/components/ElementEditor/Header.js
```diff
@@ -16,6 +17,19 @@ class Header extends Component {
     };
   }
 
+  componentDidUpdate() {
+    if (this.state.tooltipOpen && this.props.disableTooltip) {
```
raissanorth: Perhaps we could destructure these props and state attributes for consistency's sake?!

ScopeyNZ:
Updated with feedback & fixed Behat which were actually all valid failures 😥

client/src/components/ElementEditor/ElementEditor.js
```diff
@@ -1,15 +1,52 @@
...
+    this.state = {
+      dragTargetElementId: null,
+      dragSpot: null,
+    };
+
+    this.handleDragOver = this.handleDragOver.bind(this);
+    this.handleDragEnd = this.handleDragEnd.bind(this);
+  }
+
+  handleDragOver(element = null, isOverTop = null) {
```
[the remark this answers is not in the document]
ScopeyNZ: Done :)

raissanorth:
Tested it locally and it's awesome! Great job.

ScopeyNZ:
Done 🙂

client/src/components/ElementActions/PublishAction.js
```diff
@@ -9,21 +9,20 @@ import i18n from 'i18n';
  * Adds the elemental menu action to publish a draft/modified block
  */
 const PublishAction = (MenuComponent) => (props) => {
+  const { element, actions: { handlePublishBlock } } = props;
```
NightJar: I think `element` here could become `element: { id, Title: title, BlockSchema, Version, isLiveVersion }` and remove a lot of the other changes (effectively unnecessary repetition).
Same with the other `Actions` HoCs.

ScopeyNZ: I considered this but thought it was being sort of over-verbose:

`element: { ID: id, Title: title, BlockSchema: { type }, Version: version, IsLiveVersion: isLiveVersion }`

We've not de-structured `element` in some other places too.

I will change it though - it doesn't really bother me what we end up doing.

NightJar: Yeah, tbh I was of 2 minds about it. Having the prefix makes it obvious where the value comes from too.
But I think it's that we prefer destructuring where possible.

NightJar: On second thought, I don't care that strongly about this.

PR #456, opened 2018-10-18
FIX Two CMSField fixes (related to UploadField / FileBlock)

### LINKED ISSUE #452: Unable to upload files inline
Running the latest elemental `dev-master f7a4c73826` it is not possible to add a file to a file or banner block through the inline interface.

ScopeyNZ:
Fixes #452

- Fields that are also request handlers are now supported
- Fields that have submitted values different from saved values are now supported

### DIFF
--- a/src/Controllers/ElementalAreaController.php
+++ b/src/Controllers/ElementalAreaController.php
@@ -165,6 +165,14 @@ class ElementalAreaController extends CMSMain
         return HTTPResponse::create($body)->addHeader('Content-Type', 'application/json');
     }
 
+    /**
+     * Provides action control for form fields that are request handlers when they're used in an in-line edit form.
+     *
+     * Eg. UploadField
+     *
+     * @param HTTPRequest $request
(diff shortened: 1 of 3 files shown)


src/Controllers/ElementalAreaController.php
```diff
@@ -159,6 +161,19 @@ public function apiSaveForm(HTTPRequest $request)
         return HTTPResponse::create($body)->addHeader('Content-Type', 'application/json');
     }
 
+    public function formAction(HTTPRequest $request)
```
robbieaverill: Can we have a doc block explaining why this method exists please?

raissanorth: Thanks for adding the doc block!

raissanorth:
I have pulled the changes and tried it out locally. So far the file only gets saved when the entire page is saved, but not when an individual element is saved via the button in its action menu. 😞

NightJar:
This is still waiting for corresponding work in https://github.com/silverstripe/silverstripe-admin/pull/711
Upload field sort-of needs to work in order to have this be validated for merge :)

PR #457, opened 2018-10-18
NEW Details about element types now come from config

ScopeyNZ:
This replaces the current source of this information - the GraphQL query. Because this will not change after a block has been created and it is implementation detail it doesn't really seem appropriate to exist in the GraphQL query

This should enable some wizardry with optimistic updates on add :crossed_fingers:


ScopeyNZ:
I'm also not entirely convinced that I should extract the element types from config when it's needed... We could still pass this through as props and just extract from config during the "entwine". Keen to hear thoughts.

ScopeyNZ:
I'm just going to re-attempt this PR rather than trying to rebase it up to scratch.

PR #463, opened 2018-10-22
API Remove reference HistoryControllerFactory.

maxime-rainville:
Fix https://github.com/silverstripe/silverstripe-versioned-admin/issues/67 and https://github.com/silverstripe/silverstripe-cms/issues/2277

### DIFF
--- a/src/Extensions/HistoryControllerFactoryExtension.php
+++ b/src/Extensions/HistoryControllerFactoryExtension.php
@@ -7,11 +7,15 @@ use SilverStripe\Core\Extension;
 
 /**
  * Instructs the history viewer controller to be enabled for pages that have elemental
+ *
+ * @deprecated 3.0.0:4.0.0
  */
 class HistoryControllerFactoryExtension extends Extension
 {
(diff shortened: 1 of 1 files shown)


ScopeyNZ:
I'm not convinced that we should do this in the 3 branch. The 4 branch will be released soon(-ish) and we can remove it from that line (it has 4.3 as a minimum requirement), but the 3 line can still run with 4.2 that would benefit from this code.

robbieaverill:
It depends on the constraint in this module. 3.x will still work on SS 4.3

ScopeyNZ:
Yeah but people using the latest version of the 3 branch on 4.2 would no longer have the new page history viewer with this patch.

maxime-rainville:
Cool ... I brought back the YML config and tweaked the logic to return `void` instead of `false`. I think we're better off waiting until the versioned-admin PR is merged before merging this one. Removing `HistoryControllerFactoryExtension` from 4 can wait until the 4.3 release is out the door.

PR #464, opened 2018-10-22
FIX Bind handleLoadingError in Element constructor

raissanorth:
By binding `this` to the `handleLoadingError` function in the constructor we prevent a console warning, since we are calling `this.setState()`.


raissanorth:
Tests passed, see https://travis-ci.org/dnadesign/silverstripe-elemental/builds/444831002?utm_source=github_status&utm_medium=notification

PR #465, opened 2018-10-22
FIX Update prop type of elementId to string

raissanorth:
This PR changes the proptype of elementId provided to the `HoverBar` component to correctly reflect that a String is passed (and thus prevent a console warning).


client/src/components/ElementEditor/HoverBar.js
```diff
@@ -158,7 +158,7 @@ class HoverBar extends Component {
 
 HoverBar.propTypes = {
   elementTypes: PropTypes.arrayOf(elementTypeType).isRequired,
-  elementId: PropTypes.number.isRequired,
+  elementId: PropTypes.string.isRequired,
```
NightJar: Could it be `oneOfType(PropType.string, PropType.number)`?
I feel it would be a little more future proof (they _should_ be integers, in theory).

raissanorth: Okay, I can change it.

robbieaverill: Me too, we have been going back and forth constantly on defining prop types for IDs. At this stage they’re numbers, the only trick is whether SilverStripe provides them as ints or casts them as strings when returning from the ORM or from graphql queries. It’s not a biggy right now though, I agree there

NightJar: Yeah to be clear, I too would prefer it to be `number`, an only number. But the return type doesn't seem to be consistent enough, so for now this is a middle ground I'm content with.

PR #475, opened 2018-10-29
FIX Reinstate CSS selector to find popover in Behat tests

raissanorth:
This PR reverts changes made to the CSS selector to find the popover within a custom definition to allow the simulation of choosing block types. 

This also  affects inheriting block's Behat tests such as `Banner Block`.


ScopeyNZ:
I think this is caused by something else. We'll re-open if we need it 🙂

PR #494, opened 2018-11-09
Upgrade to React 16 and latest node.js version

lukereative:
As a part of https://github.com/silverstripe/silverstripe-admin/issues/648


robbieaverill:
@unclecheese can you suggest a good way of easily switching between node 6 and 10 on MacOS? I'm unable to build the master branch at the moment, but will need to keep using node 6 for all other module development

ScopeyNZ:
> @unclecheese can you suggest a good way of easily switching between node 6 and 10 on MacOS? I'm unable to build the master branch at the moment, but will need to keep using node 6 for all other module development

```
brew install nvm
nvm install 10
nvm install 6
nvm use [num]
```

PR #495, opened 2018-11-13
NEW Adding a plus icon to the Add Blocks button

ScopeyNZ:
Adds a missing plus icon from the button that was in the designs at some point

Goes from:

image

To:

image

/cc @sachajudd


ScopeyNZ:
Hopefully as it's tagged as a beta we can still squeeze this in? 🙏

PR #498, opened 2018-11-15
Check if elementdata exists before looping it

lexakami:
This allows the user to publish the page without having any blocks on it.


lexakami:
Just realised this only applies when there are no blocks on the whole site, but still good to have 😄

ScopeyNZ:
Hmmmn. I'll have a think about this and update the PR

PR #500, opened 2018-11-16
FIX Block types that are in-line editable will have links to their page on reports

ScopeyNZ:
Currently the link on reports that takes you to edit a block will take you to the `GridFieldDetailForm` view of the block. Given we now offer in-line editing on a block it makes more sense to take the user to the page where the block exists. 

I don't know why we haven't changed the `CMSEditLink` API for in-line editable blocks but I'm happy to do that instead if we agree that makes more sense.

### DIFF
--- a/src/Reports/ElementsInUseReport.php
+++ b/src/Reports/ElementsInUseReport.php
@@ -83,7 +83,7 @@ class ElementsInUseReport extends Report
      * @param BaseElement $item
      * @return string
      */
-    protected function getEditLink($value, $item)
+    protected function getEditLink($value, BaseElement $item)
     {
         return sprintf(
             '<a class="grid-field__link" href="%s" title="%s">%s</a>',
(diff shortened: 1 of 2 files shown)


ScopeyNZ:
I think I'll leave `CMSEditLink` for #506 to fix. Until then I'm not sure we can just switch `CMSEditLink` to just take you to the page containing the block - seems a bit lacking in terms of functionality.

ScopeyNZ:
I've changed my mind again - I think we should update CMSEditLink because we're running into the same issue on HistoryViewer now.

I'll update this PR...

src/Reports/ElementsInUseReport.php
```diff
@@ -83,7 +83,7 @@ public function columns()
      * @param BaseElement $item
      * @return string
      */
-    protected function getEditLink($value, $item)
+    protected function getEditLink($value, BaseElement $item)
```
robbieaverill: If we can removed this change then LGTM - this would be a semver violation for a patch change

robbieaverill: Oh but this module isn't stable in this new major release line yet, so it's OK

PR #501, opened 2018-11-16
NEW Using the breadcrumbs API of reports if available

### LINKED ISSUE #507: Breadcrumbs in "Content blocks in use" report should show parent report
Currently the "Content blocks in use" is a de facto "child" of the "Content block types" report. The breadcrumbs should reflect this so it's easier to navigate these two reports.

### PRs

- [x] #501 
- [x] silverstripe/silverstripe-reports#113

cc @sachajudd

ScopeyNZ:
Using the new feature added in silverstripe/silverstripe-reports#113 this will update the breadcrumbs on the "Elements in use" report so it is a child of the "Element type" report. This code is benign without the right version of reports

Fixes #507

### DIFF
--- a/src/Reports/ElementsInUseReport.php
+++ b/src/Reports/ElementsInUseReport.php
@@ -13,6 +13,8 @@ class ElementsInUseReport extends Report
 {
     /**
      * The string used in GET params to filter the records in this report by element type
+     *
+     * @var string
      */
     const CLASS_NAME_FILTER_KEY = 'ClassName';
 
@@ -116,7 +118,7 @@ class ElementsInUseReport extends Report
     }
 
     /**
-     * When used with newer versions of silverstripe-reports this method will automatically be added as breadcrumbs
+     * When used with silverstripe/reports >= 4.4, this method will automatically be added as breadcrumbs
      * leading up to this report.
      *
      * @return ArrayData[]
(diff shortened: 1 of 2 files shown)


src/Reports/ElementsInUseReport.php
```diff
@@ -6,10 +6,16 @@
 use SilverStripe\Forms\GridField\GridField;
 use SilverStripe\ORM\DataList;
 use SilverStripe\Reports\Report;
+use SilverStripe\View\ArrayData;
 use SilverStripe\View\Requirements;
 
 class ElementsInUseReport extends Report
 {
+    /**
+     * The string used in GET params to filter the records in this report by element type
```
robbieaverill:
```suggestion
     * The string used in GET params to filter the records in this report by element type
     *
     * @var string     
```

src/Reports/ElementsInUseReport.php
```diff
@@ -109,6 +115,30 @@ public function getReportField()
         return $field;
     }
 
+    /**
+     * When used with newer versions of silverstripe-reports this method will automatically be added as breadcrumbs
```
robbieaverill: Can you be more explicit with "newer"? This would have a tendency to confuse me in 2 years time 😆it'd be silverstripe/reports >=4.4 right?

robbieaverill:
I pushed a commit updating two bits of doc blocks, since there are no logical changes I assume you'll be fine with my updates, so I'll merge on green

PR #502, opened 2018-11-16
ENHANCEMENT Adjust getType to use singular_name

scott1702:
Currently developers need to override the `getType` function in order to get a custom block name showing in the dropdown. I think using singular_name as the default makes sense as this covers a common use case for what people may want from `getType`


robbieaverill:
I think we can make this a documentation change instead, sorry and thanks anyway!

PR #504, opened 2018-11-16
Add $Anchor as class in ElementHolder

thezenmonkey:
Best practice for CSS requires using classes for styling instead of the IDs. Adding the $Anchor as css class will allow developers to target individual elements with CSS in more standardized way


ScopeyNZ:
I'm not sure I agree with having anchor as a class. It's a unique identifier - which makes sense as an ID. Also it's "volatile" and could change easily as it adds numbers depending on the order of blocks on a page. 

What's the particular use case here?

ScopeyNZ:
I'm not convinced this is a good idea for the reasons stated above. I'm going to close this PR. Please feel free to comment if you feel strongly about this. Thanks for the contribution!

PR #515, opened 2018-11-20
FIX TextCheckboxGroupField should passthrough if it has only one child

ScopeyNZ:
Currently if a developer removes the checkbox it will render a small gap on the side of the text field. This will prevent that. 

cc @scott1702


ScopeyNZ:
First try!

scott1702:
Can confirm this works for me! Should be worth noting that this fix is only required because the only way we've been able to remove the isDisplayed checkbox is with:
```
$fields->removeByName('ShowTitle');
$fields->fieldByName('Root.Main.Title')->setTitle('Title');
```

ScopeyNZ:
Tests should be fixed by #516

robbieaverill:
So it becomes a text checkbox group without a checkbox? Wouldn’t it be better to remove the group and use a simple text field for the title?

ScopeyNZ:
Yeah I was noting the redundancy of this fix as I did it. Currently it's not so easy for reasons I haven't quite nailed down yet. I figure this at least handles this situation in case somebody does similar...

robbieaverill:
Probably because CompositeFields don't have names, so they're hard to target via `$fields->removeByName()` 😆

PR #520, opened 2018-11-23
NEW Fixing change tracking and adding an icon to show unsaved changes

ScopeyNZ:
As titled.

Fixes instances where the "unsaved changes" warning would appear when nothing has changed
Fixes the page save/pulish actions to correctly note the dirty/pristine state of the elemental area
Adds a new unsaved changes icon to blocks - like the designs had in the first place.

cc @clarkepaul @newleeland

I don't like the interaction between entwine & react :(


ScopeyNZ:
There's probably a bunch of "on save" stuff I have to fix. Thanks Behat... 😥

robbieaverill:
Slash @raissanorth

ScopeyNZ:
🐝 🎩 👑

NightJar:
We're in the RC phase of this now, should this `NEW` feature not be targeting 4.1?

ScopeyNZ:
Luckily we're still in beta 😉 . There's two bugfixes here too. If you feel strongly I can separate the "NEW" from the "FIX"es.

robbieaverill:
I raised #524, the indicator should be orange not blue

PR #521, opened 2018-11-25
NEW Element types are now exposed in config

ScopeyNZ:
- Types are registered through a registry
- The registry interpolates some configuration information (like icon, type name, in-line editability, etc...)
- Updates the React app to fetch the type from config instead

Take 3. Recreation of #457 

This paves the way for both optimistic mutation responses (hopefully), some work on GraphQL updates (scaffolding a query for each type of block), and a slightly more extensible way of doing block previews.

Also this removed dependency on `BlockSchema` which is a bit of an anti-pattern.


client/src/components/ElementEditor/tests/Header-test.js
```diff
@@ -10,15 +10,19 @@ Enzyme.configure({ adapter: new Adapter() });
...
-    InlineEditable: true,
     Title: 'Sample File Block',
-    BlockSchema: {
-      type: 'File',
-      iconClass: 'font-icon-block-file',
-    },
+  };
+  const type = {
+    inlineEditable: true,
+      title: 'File',
```
robbieaverill: Indentation is a bit out here

PR #522, opened 2018-11-27
FIX React element areas should be specific to areas not the page

### LINKED ISSUE #387: Change "readBlocksForPage" to "readBlocksForArea"

ScopeyNZ:
Fixes #387

Currently the new elemental area editor does not support multiple areas per page - where it apparently did previously. This PR updates the read blocks mutation to read per area rather than page. Addiitonally there's some more fixes to help multiple blocks per page:

- Updated the readBlocks mutation to fetch per area and not page
- Added some padding on the bottom of the field
- Renamed elementAreaId to just areaId in the React code - element is already in the name of all the components
- Updated sortBlock optimistic updates as needed
- Updated addElementToArea mutation to specify the variables for the refetchQuery to run

cc @clarkepaul


client/src/state/editor/readBlocksForAreaQuery.js
```diff
@@ -5,25 +5,22 @@ import gql from 'graphql-tag';
...
-        edges {
-          node {
-            ID
-            Title
-            BlockSchema
-            IsLiveVersion
-            IsPublished
-            Version
-          }
+    Elements {
```
robbieaverill: Does this mean we can remove the extension that provides `ElementalAreaIfExists` now?

ScopeyNZ: Ah yep I forgot to do that.

ScopeyNZ:
It appears the actions aren't working for Behat 😞 .

client/src/state/editor/readBlocksForAreaQuery.js
ScopeyNZ: Ok I've removed that extension as part of this PR 🙂

client/src/state/editor/sortBlockMutation.js
```diff
@@ -28,13 +28,13 @@ const config = {
...
       update: store => {
-        const variables = readBlocksConfig.options({ pageId }).variables;
+        const variables = readBlocksConfig.options({ areaId }).variables;
         const data = store.readQuery({ query: readBlocksQuery, variables });
 
         // Query returns a deeply nested object. Explicit reconstruction via spreads is too verbose.
         // This is an alternative, relatively efficient way to deep clone
         const newData = JSON.parse(JSON.stringify(data));
-        let { edges } = newData.readOnePage.ElementalAreaIfExists.Elements;
+        let { edges } = newData.readOnePage.Elements;
```
robbieaverill: Should this now be `readOneElementalArea`?

client/src/state/editor/sortBlockMutation.js
```diff
@@ -56,7 +56,7 @@ const config = {
         }
 
         // Add it back to the full result
-        newData.readOnePage.ElementalAreaIfExists.Elements.edges = edges;
+        newData.readOnePage.Elements.edges = edges;
```
robbieaverill: Also here, should this be `readOneElementalArea`? I'm looking at the changes in the GraphQL query

src/Extensions/SiteTreeExtension.php
```diff
@@ -1,26 +0,0 @@
...
-     * @return ElementalArea
-     */
-    public function ElementalAreaIfExists()
-    {
-        if ($this->owner->hasExtension(ElementalPageExtension::class)) {
-            return $this->owner->ElementalArea();
-        }
-        return ElementalArea::create();
-    }
-}
```
robbieaverill: Also needs the config removed: https://github.com/dnadesign/silverstripe-elemental/blob/ed298dc8c712761154c16760f61572b03c6ccfbb/_config/extensions.yml

client/src/state/editor/sortBlockMutation.js
ScopeyNZ: That's why sort wasn't working! I missed one 😅

src/Extensions/SiteTreeExtension.php
ScopeyNZ: I did remove that :) It's at the very top of the diff.

robbieaverill: Weird, I found it in your commit tree. I guess the dev/build would've failed anyway, ignore me

ScopeyNZ:
I've updated from feedback but I haven't had a chance to re-test sorting. I'd like to do that before any merge. Thanks for the feedback so far!

robbieaverill:
Drag and drop isn't working for me - console errors

robbieaverill:
It looks like this is happening because `ElementDragPreview` is rendered without providing an `elementTypeType` to it, which then renders a `Header` that requires an `elementTypeType` prop called `type` which doesn't get provided.

I had a quick go at patching this but I don't know enough about the implementation. @ScopeyNZ could you please have a look at fixing that up?

ScopeyNZ:
That's all fixed up now. Thanks for the debugging assistance @robbieaverill 🙂 . Refer to the last commit if you were interested in how I addressed the issue.

PR #532, opened 2018-11-30
FIX being able to swap tabs on inline blocks

scott1702:
A previous PR removed unpacking all props into the child component in favour of explicitly defining the properties. In this case the `handleTabClick` handler was not being passed down which was preventing the tab buttons in the actions menu from working.


client/src/components/ElementEditor/Header.js
```diff
@@ -148,6 +149,7 @@ class Header extends Component {
                 element={element}
                 activeTab={activeTab}
                 editTabs={type.editTabs}
+                handleEditTabsClick={handleEditTabsClick}
```
ScopeyNZ: Would you mind adding this to the PropTypes on the `ElementActions` component?

robbieaverill:
Build error:

```
  ● Console

    console.error node_modules/fbjs/lib/warning.js:33
      Warning: Failed prop type: Required prop `handleEditTabsClick` was not specified in `ElementActions`.
          in ElementActions


```

Will fix while rebasing #522

PR #533, opened 2018-11-30
MINOR Avoid fetching owner page when ElementalArea has not been saved

maxime-rainville:
This change does 2 things to speed up `ElementalArea::getOwnerPage()`:
* Start by checking if the ElementalArea has an ID and return null if it doesn't. This is quite beneficial when working with a new ElementalArea, because `null` responses aren't cache.
* Update the logic to find owner page when `OwnerClassName` is unknown. Instead of doing one query for each combination elemental relation of each elemental page type, it will do one query for each elemental page type. It's only going to make a difference if you have pages with multiple elemental areas ... so it's not going to make a difference for most people.

Related to our beloved Key Project.

### DIFF
--- a/src/Models/ElementalArea.php
+++ b/src/Models/ElementalArea.php
@@ -187,6 +187,11 @@ class ElementalArea extends DataObject
      */
     public function getOwnerPage()
     {
+        // You can't find the owner page of a area that hasn't been save yet
+        if ($this->ID == 0) {
+            return null;
+        }
+
         // Allow for repeated calls to read from cache
         if (isset($this->cacheData['owner_page'])) {
             return $this->cacheData['owner_page'];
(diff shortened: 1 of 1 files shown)


src/Models/ElementalArea.php
```diff
@@ -187,6 +187,11 @@ public function onBeforeWrite()
      */
     public function getOwnerPage()
     {
+        // You can't find the owner page of a area that hasn't been save yet
+        if ($this->ID == 0) {
```
ScopeyNZ: Any reason we're not using `isInDb()` here?

```suggestion
        if (!$this->isInDB()) {
```

maxime-rainville: Because ... I wanted to know if you were paying attention :roll_eyes:

PR #534, opened 2018-11-30
FIX Use standalone PropTypes library instead of that bundled with React 15

### LINKED ISSUE #529: Restrict v4.0 to SS4.3
From @robbieaverill :

When running elemental 4.0 (react 15) on SS 4.4 (react 16 backend) I'm getting an undefined on PropTypes from react, even though we import PropTypes from our own version of react in elemental... my initial suspicion is that webpack-config is loading react from admin instead, and that we'll need to restrict this version of elemental (4.0) to SS 4.3 only and start a new 4.1 for SS 4.4

robbieaverill:
Resolves #529


robbieaverill:
Drag and drop doesn't work but I think that's a pre-existing issue. I've left some notes on #522 about why this is happening.

ScopeyNZ:
I'll fix the drag and drop and do a full run through in the other PR 🙂

PR #537, opened 2018-12-03
FIX always build an ElementalArea for all newly extended classes

NightJar:
Previously only brand new pages would have ElementalAreas built for
them, as this process happens on write. If an existing page type was
updated to have ElementalAreas on it, all existing pages in the CMS
would error when trying to be viewed, as the ElementalArea would not
exist. This was particularly true for GraphQL queries, which are
abundant in the upcoming 4.0 release.

### DIFF
--- a/src/Extensions/ElementalAreasExtension.php
+++ b/src/Extensions/ElementalAreasExtension.php
@@ -13,6 +13,7 @@ use SilverStripe\Core\Config\Config;
 use SilverStripe\Forms\FieldList;
 use SilverStripe\Forms\LiteralField;
 use SilverStripe\ORM\DataExtension;
+use SilverStripe\ORM\DataObject;
 
 /**
  * This extension handles most of the relationships between pages and element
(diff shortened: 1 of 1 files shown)


NightJar:
Not sure how to write tests for this one :/

NightJar:
@robbieaverill 	I would hope that the default behaviour on `DataObject` would hold true, and detect that there have been no changes to any of the data fields, and skip the write.
However all the `onBeforeWrite` hooks run _before_ this point, so we would still end up with an ElementalArea related to the parent class (often a `Page`).

This would work only because the `ElementalArea`s are a `has_one` to a page, so there is no need to write the page itself.

NightJar:
oh, wait no I got that entirely backwards :( Derp.
Will have to do a manual DB update or something (like `gridfieldextensions` does with OrderableRows), as the `has_one` is on the _page_ (or whatever it's related to), not the area.

NightJar:
Urgh, no.
We're only writing the parent (page), which does not cascade through `owns`.
Only the page would become draft, and only if it did not have elemental areas previously.

Ran a dev build twice to test, can confirm. Updated selection query helps too I guess.

PR #539, opened 2018-12-03
DOCS Update userhelp to reflect current UI updates

sachajudd:


ScopeyNZ:
If Paul is happy then so am I 🙂

PR #571, opened 2019-01-21
Open virtual link in a new window

wilr:
As opening it in the same window - when going back, displays a broken page.

### DIFF
--- a/code/models/ElementVirtualLinked.php
+++ b/code/models/ElementVirtualLinked.php
@@ -46,7 +46,7 @@ class ElementVirtualLinked extends BaseElement
     public function getCMSFields()
     {
         $message = sprintf(
-            '<p>%s</p><p><a href="%2$s">%2$s</a></p>',
+            '<p>%s</p><p><a href="%2$s" target="_blank">%2$s</a></p>',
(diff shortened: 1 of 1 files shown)


code/models/ElementVirtualLinked.php
```diff
@@ -46,7 +46,7 @@ public function __construct($record = null, $isSingleton = false, $model = null)
     public function getCMSFields()
     {
         $message = sprintf(
-            '<p>%s</p><p><a href="%2$s">%2$s</a></p>',
+            '<p>%s</p><p><a href="%2$s" target="_blank">%2$s</a></p>',
```
ScopeyNZ: I doubt this'll ever be tabnab-able but do you want to add in `rel="noopener noreferrer"` just in case?

ScopeyNZ:
Something funky is happening with the tests :suspect: 

They're "passing" but it doesn't appear as though they actually ran any tests...

ScopeyNZ:
@wilr Did you want to make a PR to the SS4 version?:

https://github.com/dnadesign/silverstripe-elemental-virtual/blob/master/src/Model/ElementVirtual.php#L63

wilr:
@ScopeyNZ can do, this project is SS3 only at the moment but will be upgrading at some point in the year!

PR #578, opened 2019-01-23
Improve i18n

wernerkrauss:
Fix TextCheckboxGroupField translation
Translate settings tab
expose i18n Javascript

### DIFF
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -295,6 +295,10 @@ class BaseElement extends DataObject
                     )
             );
 
+            // Rename the "Settings" tab
+            $fields->fieldByName('Root.Settings')
+                ->setTitle(_t(__CLASS__ . '.SettingsTabLabel', 'Settings'));
+
             // Add a combined field for "Title" and "Displayed" checkbox in a Bootstrap input group
             $fields->removeByName('ShowTitle');
             $fields->replaceField(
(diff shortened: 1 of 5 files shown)


src/Models/BaseElement.php
```diff
@@ -295,6 +295,10 @@ public function getCMSFields()
                     )
             );
 
+            // Rename the "Settings" tab
+            $fields->fieldByName('Root.Settings')
+                ->setTitle(_t(__CLASS__ . '.SettingsTabLabel', 'Content'));
```
robbieaverill: You're renaming "Settings" to "Content"? Do you think that might be a little misleading?

wernerkrauss: oops, copy/paste error, thanks for the heads up!

src/Models/BaseElement.php
```diff
@@ -295,6 +295,10 @@ public function getCMSFields()
                     )
             );
 
+            // Rename the "Settings" tab
+            $fields->fieldByName('Root.Settings')
+                ->setTitle(_t(__CLASS__ . '.SettingsTabLabel', 'Settings'));
```
robbieaverill: Ok, what was the label originally? Is this just so we can make the label translatable with an explicit translation key?

wernerkrauss: The default "Settings"; I added the key to https://github.com/dnadesign/silverstripe-elemental/pull/578/files#diff-e0c0c7c5f72bb31c02647d33f6a0b563R17 according to the other label for "Content"

Is there a more elegant way for translating tab labels?

PR #579, opened 2019-01-24
Isolate allowed elements for sorting tests to prevent other modules breaking them

robbieaverill:
Issue: https://github.com/silverstripe/cwp-recipe-kitchen-sink/issues/20


robbieaverill:
Self merging since this is a low impact change to unit tests only

PR #580, opened 2019-01-25
FIX Ensure element orders are persisted to the Live table

### LINKED ISSUE #548: Dragging of non-line and inline editable blocks do not save their position
When you have a mix of in-page editable blocks and blocks like the Form Block which aren't in-page editable, and you re-order them the blocks do not save their position.

Tested in 4.3 (pre stable release)

cc @brynwhyman

ScopeyNZ:
Currently when elements are reordered it will ensure the other elements that were affected have their sort order updated - but only on the draft table (ie `Element`). This patch ensures that this change  is also persisted to the live table (ie `Element_Live`).

Fixes #548

### DIFF
--- a/src/Services/ReorderElements.php
+++ b/src/Services/ReorderElements.php
@@ -7,6 +7,7 @@ use InvalidArgumentException;
 use SilverStripe\Core\Convert;
 use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\Queries\SQLUpdate;
+use SilverStripe\Versioned\Versioned;
 
 class ReorderElements
 {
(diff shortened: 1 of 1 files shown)


src/Services/ReorderElements.php
```diff
@@ -92,24 +93,34 @@ public function reorder($elementToBeAfterID = 0)
...
-        } else {
-            $operator = '-';
-            $filter = "\"$tableName\".\"Sort\" <= $sortAfterPosition AND \"$tableName\".\"Sort\" > $currentPosition";
-            $newBlockPosition = $sortAfterPosition;
+        $baseTableName = Convert::raw2sql(DataObject::getSchema()->tableName(BaseElement::class));
+
+        // Update both the draft and live versions of the records
+        $suffixes = [''];
+        if (BaseElement::has_extension(Versioned::class)) {
+            $suffixes[] = '_Live';
```
Cheddam: For future proofing, it might be better to use `Versioned::stageTable()` to determine the table name rather than hardcoding the suffix.

ScopeyNZ:
Okay I've made an adjustment to use the Versioned API following @Cheddam's suggestion.

PR #581, opened 2019-01-28
FIX Incorrect field names in state flush after saving

### LINKED ISSUE #565: Publishing the page causes the content in a block to disappear until refresh
Version: `4.0.0` with SilverStripe 4.3.0

Steps to replicate: 

1. Create an inline editable block on a page
2. Expand the block and add text into the content field
3. Click "Publish" on the page
4. The updated content doesn't display inside the block

If you refresh or publish a second time, the content is displayed. So it is saving the value of the field, but rendering the previous version of the content (which in this case is none). If you add more content, and publish the page, it shows the previous version of content until you refresh or publish again. This only seems to happen when publishing the page, not publishing the block individually. 

I also get this error in the console: `Unhandled promise rejection Error: "Network error: Store reset while query was in flight(not completed in link chain)"`

Cheddam:
When saving / publishing changes to block content, the Redux Form state is intended to be flushed to force the newly fetched field data to be applied. This was not happening, due to incorrect references to the field names in the `destroy` call.

Fixes #565.


robbieaverill:
The behat failure looks valid

PR #583, opened 2019-01-29
FIX Applying the allowed elements fix to the other add element buttons

ScopeyNZ:
This is a follow-up to #570. The patch was applied to affect the "Add Block" button that appears in the header but not to the "hover-bar" buttons that appear between blocks.


Kethatril:
Tested this locally and looks like its working correctly.

PR #587, opened 2019-01-30
FIX Implementing custom read resolver to fix permission issue

### LINKED ISSUE #586: Content authors cannot edit blocks
Currently only users with full administrative rights can edit content blocks.

**Steps to reproduce**:
1. Login as a content author user with permission to edit any page
2. Navigate to a page with an ElementalArea

**Expected result:**
User can edit content blocks as per usual

**Actual result**
ElementalArea gets stuck in loading state, with a JavaScript error of `Cannot view ElementalArea`

ScopeyNZ:
This is required because the default  resolver in GraphQL checks permissions against fresh data objects. Elemental areas require that they are saved before permission checks can be accurate

Fixes #586

/cc @scott1702

### DIFF
--- a/tests/Behat/features/element-editor.feature
+++ b/tests/Behat/features/element-editor.feature
@@ -6,45 +6,62 @@ Feature: View types of elements in an area on a page
 
   Background:
     Given I add an extension "DNADesign\Elemental\Extensions\ElementalPageExtension" to the "Page" class
+      And a "group" "AUTHOR group" has permissions "Access to 'Pages' section"
       And a "page" "Blocks Page" with a "Alice's Block" content element with "Some content" content
       And the "page" "Blocks Page" has a "Bob's Block" content element with "Some content II" content
 
-    Given I am logged in with "ADMIN" permissions
-      When I go to "/admin/pages"
+  Scenario Outline: I can see the title and summary of each element
+    Given I am logged in with "<group>" permissions
+    When I go to "/admin/pages"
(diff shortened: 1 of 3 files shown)


ScopeyNZ:
I've also raised a fix on the GraphQL repo: https://github.com/silverstripe/silverstripe-graphql/pull/207

Considering the impact of this bug I think this still has value but I can probably make it deprecated if/when the other PR is merged?

ScopeyNZ:
Great idea. I'll try to get around to this soon.

ScopeyNZ:
Thanks for the suggestion @Cheddam . I have now (finally) added a Behat test to cover a less privileged user.

tests/Behat/features/element-editor.feature
```diff
@@ -6,45 +6,62 @@ Feature: View types of elements in an area on a page
...
+    Given I am logged in with "<group>" permissions
+    When I go to "/admin/pages"
       And I left click on "Blocks Page" in the tree
-    When I see a list of blocks
-      Then I should see "Alice's Block"
-      And I should see "Bob's Block"
-
-  Scenario: I can see the title and summary of each element
-    Given I see a list of blocks
-    Then I should see "Alice's Block" as the title for block 1
```
ScopeyNZ: This looks like a reasonably heavy refactor but I've just moved the login and some preliminary assertions out of the `Background`. It didn't really make sense to be here and trying to relogin as a different user wasn't working for me - I didn't really look much further at that.

PR #601, opened 2019-02-10
FIX Element type is now passed to actions components so they can use the "title" for toast messages

### LINKED ISSUE #600: Saved 'Untitled {type} block' successfully
**4.0.x-dev elemental with SS 4.3.x-dev**

If you save a block without a title, you get this toast message:

image

I guess it should say "Saved untitled block successfully" with no capitalisation or placeholder if there's no title set.

### LINKED ISSUE #603: Saving a block inline with no changes reports as a failure
**4.0.x-dev**

This happens because the API result says `updated: false`, so we detect that as a failure. In fact, nothing has changed, so nothing was updated.

Perhaps we need to pass an explicit `success` boolean as well to indicate whether the request was successful or not (false if there was an exception thrown for example).

robbieaverill:
~At some point this was removed in favour of "typeName". This adds it back so toast messages can use it on the CMS frontend.~

Fixes #600 and fixes #603


ScopeyNZ:
The reason this was removed is we moved type definition to client config. It's not something that's stored in the DB - it's configured with code or yml. The "other" solution would be to ensure that the `ElementActions` component gets the type passed to it, and you can extract the type name from that. See:

https://github.com/dnadesign/silverstripe-elemental/blob/master/client/src/components/ElementEditor/Header.js#L100-L103

And further down in that file:

https://github.com/dnadesign/silverstripe-elemental/blob/master/client/src/components/ElementEditor/Header.js#L150-L156

Up to you what you think is best. I feel like removing as much reliance on `BlockSchema` as possible is the good option here.

robbieaverill:
Yeah, that makes sense. I didn't realise the `type` was available in the Element component already. Updating...

PR #604, opened 2019-02-12
Hotfix/require default records

StephenMakrogianni:
- Added extra qurey condition to account for values equal to 0
- Removed unused variable. Added extra query parameters


robbieaverill:
Thanks!

PR #610, opened 2019-02-13
FIX Ensuring drag events do not start an element reorder drag when the target is an inline edit form

### LINKED ISSUE #575: Unable to highlight text areas / other fields in inline editing mode
When you expand an element in inline editing mode, and try to highlight text inside the field, it causes the element to re-render:

output

and you cannot highlight the text. 

The react-dnd library needs to disable drag-and-drop when an element is expanded.

ScopeyNZ:
Fixes #575

This change adds the "DragSource" to both the element and the header within the element. The element remains the drag source while the in-line form is not expanded. The header becomes the drag source when the in-line form is expanded.


client/src/components/ElementActions/SaveAction.js
```diff
@@ -64,8 +64,7 @@ const SaveAction = (MenuComponent) => (props) => {
           type: 'success'
         });
       })
-      .catch((something) => {
-        console.log(something);
```
robbieaverill: Didn't you play cricket as a kid?

client/src/components/ElementEditor/Header.js
```diff
@@ -174,6 +198,9 @@ Header.propTypes = {
   previewExpanded: PropTypes.bool,
   disableTooltip: PropTypes.bool,
   formDirty: PropTypes.bool,
+  connectDragSource: PropTypes.func.isRequired,
+  connectDragPreview: PropTypes.func.isRequired,
```
robbieaverill: Since these are required it constitutes a breaking change to this class. It's unlikely that anyone has extended this class in 4.0 since its release, but noting it anyway.

ScopeyNZ: Yeah that's a good point. It might be slightly different considering those two are provided with an HoC, but any extension will have to implement the DragSource HoC now anyway.

PR #616, opened 2019-02-18
Remove coupling between ElementalArea and SiteTree

kinglozzer:
As [briefly discussed on Slack](https://slackarchive.silverstripe.org/slack-archive/message/256649), this is the hotfix I’m currently using with deeply-nested `ElementalAreas`


robbieaverill:
Do you want to target 3.0? I'm not sure we'll be doing any more 3.x releases

kinglozzer:
Have switched to 3.0 :)

PR #618, opened 2019-02-21
FIX Cleaning the look of the "View Block" link in history viewer

### LINKED ISSUE #509: History styling is off: Caret on the history "View Block" link is incorrect & font sizing issues
### Affected Versions:

3.x, 4.x

#### Firstly the "View Block" link:

The font-icon is too small.

image

The underline on hover shouldn't go under the icon: 

image

Designs:

image

_(Note that this was originally intended to link to history of the individual block. Currently this only goes to edit the draft version of the block. The new wording is correct)_

#### The font size issue:

The titles of the various fields seem to have a non-standard smaller font-size:

image

### Steps to reproduce

- Open a blocks page
- Have at least one block
- View the history for the page and select a version that has your block
- Observe the weird icon on the "View block" link & the small font size.

cc @sachajudd 
Reference: https://github.com/dnadesign/silverstripe-elemental/issues/223

ScopeyNZ:
Resolves #509.

Now looks like:

image (Hover state)

/cc @sachajudd 

(Is changing CSS classes considered breaking? I can change them back if need be...)


client/src/components/HistoricElementView/HistoricElementView.js
```diff
@@ -27,9 +27,9 @@ const ElementalAreaHistoryFactory = (FieldGroup) =>
           {legend}
           <div className={'elemental-preview elemental-preview--historic'}>
             {data.ElementEditLink &&
-              <a className={'elemental-preview__link'} href={data.ElementEditLink}>
-                {i18n._t('HistoricElementView.BLOCK_HISTORY', 'View block')}
-                <i className={'font-icon-angle-right'} />
+              <a className="elemental-preview-link" href={data.ElementEditLink}>
```
robbieaverill: The CSS classname change could be interpreted as a breaking change. I don't see a problem with the way it was - perhaps you could revert that part?

ScopeyNZ: Fixed.

Issue #622, opened 2019-02-26
Error: Call to a member function canView() on null


emteknetnz:
dnadesign/silverstripe-elemental - 4.0.x-dev - 37d13f1ec4d8207ce860a76942a1c452ec7be360
SilverStripe 4.3.1

Error is from Raygun, not sure what triggered this. 

Reported url is /admin/graphql

```
#0 DNADesign\Elemental\GraphQL\ReadOneAreaResolver::resolve called at [:0]
#1 ::call_user_func_array called at [/var/www/<dirname>/vendor/silverstripe/graphql/src/Scaffolding/Scaffolders/OperationScaffolder.php:427]
#2 SilverStripe\GraphQL\Scaffolding\Scaffolders\OperationScaffolder::SilverStripe\GraphQL\Scaffolding\Scaffolders\{closure} called at [:0]
#3 ::call_user_func called at [/var/www/<dirname>/vendor/webonyx/graphql-php/src/Executor/Executor.php:488]
#4 GraphQL\Executor\Executor::resolveOrError called at [/var/www/<dirname>/vendor/webonyx/graphql-php/src/Executor/Executor.php:449]
#5 GraphQL\Executor\Executor::resolveField called at [/var/www/<dirname>/vendor/webonyx/graphql-php/src/Executor/Executor.php:227]
#6 GraphQL\Executor\Executor::executeFieldsSerially called at [/var/www/<dirname>/vendor/webonyx/graphql-php/src/Executor/Executor.php:249]
#7 GraphQL\Executor\Executor::executeFields called at [/var/www/<dirname>/vendor/webonyx/graphql-php/src/Executor/Executor.php:174]
#8 GraphQL\Executor\Executor::executeOperation called at [/var/www/<dirname>/vendor/webonyx/graphql-php/src/Executor/Executor.php:93]
#9 GraphQL\Executor\Executor::execute called at [/var/www/<dirname>/vendor/webonyx/graphql-php/src/GraphQL.php:56]
#10 GraphQL\GraphQL::executeAndReturnResult called at [/var/www/<dirname>/vendor/silverstripe/graphql/src/Manager.php:354]
#11 SilverStripe\GraphQL\Manager::SilverStripe\GraphQL\{closure} called at [/var/www/<dirname>/vendor/silverstripe/graphql/src/Middleware/HTTPMethodMiddleware.php:30]
[...]

PR #623, opened 2019-02-26
FIX Assert elemental areas exist before calling methods on them

ScopeyNZ:
Related: #622

### DIFF
--- a/src/GraphQL/ReadOneAreaResolver.php
+++ b/src/GraphQL/ReadOneAreaResolver.php
@@ -3,7 +3,9 @@
 namespace DNADesign\Elemental\GraphQL;
 
 use DNADesign\Elemental\Models\ElementalArea;
+use Exception;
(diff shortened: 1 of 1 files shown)


src/GraphQL/ReadOneAreaResolver.php
```diff
@@ -12,6 +12,10 @@ public function resolve($object, array $args, $context, ResolveInfo $info)
     {
         $area = ElementalArea::get()->byID($args['ID']);
 
+        if (!$area) {
+            throw new \Exception('Could not find elemental area matching ID ' . $args['ID']);
```
sminnee: `InvalidArgumentException` maybe?

ScopeyNZ: Using base exception is fairly rampant in GraphQL code. I agree with using specific exceptions though. I've updated it

PR #635, opened 2019-03-18
FIX Allow multi readonly trans in TextCheckboxGroupField

### LINKED ISSUE #630: Title field causes fatal error on read only transformation
Elemental `4.0.x-dev#1624abcfa413ceb1583fdf233cff280c61237a8f` has a non-recoverable fatal error when a user attempts to view a read only page with an `ElementalArea` on it.
By "non-recoverable" I mean that this error once seen persists even after enabling the user to edit that page again.

Reproduction steps:
- Create a user
- Create a group
- Give the group the ability to access the Pages section of the CMS (as the only permission) image
- Assign the new user to the new group
- Sign in as the new user and create a `BlocksPage` (in a browser/session)
- Edit that blocks page (add blocks, that have titles and have the displayed checked)
- Publish the blocks page
- Use the admin user on another browser/session to use "Site Settings: Access" tab and enable only Administrators to edit pages 
image
- On the browser/session with the new user logged in, refresh the page.
- Observe that the elemental edit view still loads
- Browse to the home page to ensure that it is read only (it should be).
- Browse back to the Blocks Page
- Observe an "Internal server error" toast message
- refresh the page (the URL should be correctly set to edit the Blocks Page)
- Observe a Alert level error `[Alert] Call to a member function Value() on null` 
image
- Interestingly also observe the lack of stack trace 
image

## Setup

- Error occurs on `vendor\dnadesign\silverstripe-elemental\src\Forms\TextCheckboxGroupField.php`
- using `cwp/cwp-recipe-kitchen-sink` as context for testing.
- Windows 10, PHP 7.1.10, Apache 2.4.23
- SilverStripe 4.3.x-dev, `framework#a8605b0`, `cms#543307a`, `admin#d1d0a73`, `versioned-admin#747c508`, `versioned#4ff6126` Elemental as per opening sentence.

chillu:
Fixes #630

In general, every field in SilverStripe should be able to reliably transform itself into a readonly version. That includes fields which are *already* readonly, for example `ReadonlyField->performReadonlyTransformation()` simply returns `$this`. Most fields don't need the `$this->readonly` check implemented in here, because they return a *different* field instead (most of the time a `ReadonlyField` or `LookupField`).

### DIFF
--- a/src/Forms/TextCheckboxGroupField.php
+++ b/src/Forms/TextCheckboxGroupField.php
@@ -39,23 +39,27 @@ class TextCheckboxGroupField extends CompositeField
      */
     public function performReadonlyTransformation()
     {
-        $field = parent::performReadonlyTransformation();
+        $field = $this;
 
-        $field->setTemplate(CompositeField::class);
-        $field->setTitle('Title');
+        if (!$this->readonly) {
+            $field = parent::performReadonlyTransformation();
 
-        $field->replaceField('Title', LiteralField::create(
-            'Title',
-            $field->fieldByName('Title')->Value()
-        ));
+            $field->setTemplate(CompositeField::class);
+            $field->setTitle('Title');
 
-        $displayedText = _t(__CLASS__ . '.DISPLAYED', 'Displayed');
-        $notDisplayedText = _t(__CLASS__ . '.NOT_DISPLAYED', 'Not displayed');
+            $field->replaceField('Title', LiteralField::create(
+                'Title',
+                $field->fieldByName('Title')->Value()
+            ));
 
-        $field->replaceField('ShowTitle', LiteralField::create(
-            'ShowTitle',
-            $field->fieldByName('ShowTitle')->Value() === 'Yes' ? $displayedText : $notDisplayedText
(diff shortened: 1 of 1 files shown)


robbieaverill:
Do we want this against the 4.0 branch?

chillu:
It's a bugfix, so technically I can rebase to a minor release branch - would you prefer that? Depends on when 4.1 is planned I guess.

robbieaverill:
Yeah I think so. 4.1 will be SilverStripe 4.4 minimum, so 4.0 will be the only version people can use until they upgrade to that core release

ScopeyNZ:
(And 4.1 is broken until we get `HtmlEditorField` fixed in `silverstripe/admin`)

chillu:
Here you go: https://github.com/dnadesign/silverstripe-elemental/pull/636. I thought you could rebase PRs in the UI now? Or did I dream that?

robbieaverill:
You can rebase on merge, and you can switch the target branch, but you can’t rebase in the UI otherwise. We usually switch the branch in GitHub and rebase manually.

Issue #617, opened 2019-02-18
MigrateContentToElement doesn't find any pages to migrate


quamsta:
After adding ElementalPageExtension to Page's extensions and running a dev/build?flush, ElementalAreas are created and the MigrateContentToElement task doesn't seem to work because of this filter:

`$pages = $pageType::get()->filter('ElementalAreaID', 0);`

Since every page now has an ElementalArea, the ElementalAreaID will never be 0. Not even sure when this filter could ever be true. I removed the filter, and this task does seem to run, but creates multiple Elements per ElementalArea, and I'm not really sure where to go from there.

ScopeyNZ:
Yeah that's an interesting problem. In the end it looks like it was easier to just to get the new editor handling missing areas rather than ensuring the pages always have an area. Not quite sure how to fix this one too be honest - hence the `effort/hard` label 😅

altwohill:
What about adjusting the filter to be 
` $pages = $pageType::get()->filter('Content:not', '');`

robbieaverill:
@ScopeyNZ does #652 resolve this?

ScopeyNZ:
Yes.

PR #638, opened 2019-03-19
Addresses #617

altwohill:
Update filter to select based off non-empty content.
Publish changes (in line with the SS3->4 Image Asset Migration policy)

### DIFF
--- a/src/Tasks/MigrateContentToElement.php
+++ b/src/Tasks/MigrateContentToElement.php
@@ -6,6 +6,7 @@ use DNADesign\Elemental\Models\ElementalArea;
 use DNADesign\Elemental\Models\ElementContent;
 
 use SilverStripe\Dev\BuildTask;
+use SilverStripe\Versioned\Versioned;
 
 class MigrateContentToElement extends BuildTask
 {
@@ -21,7 +22,7 @@ class MigrateContentToElement extends BuildTask
         $pageTypes = singleton(ElementalArea::class)->supportedPageTypes();
         $count = 0;
         foreach ($pageTypes as $pageType) {
-            $pages = $pageType::get()->filter('ElementalAreaID', 0);
+            $pages = $pageType::get()->filter('Content:not', ['', null]);
             foreach ($pages as $page) {
                 $content = $page->Content;
                 $page->Content = '';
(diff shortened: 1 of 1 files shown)


ScopeyNZ:
Looks like that should work well. Would you like to do two things that will help me get this merged?

- Retarget this PR at the `4.0` branch. It's a bugfix so it makes sense to go in the next `4.0.x` release which will be based of that branch. You can edit your PR and change the base branch but you will probably also have to `git rebase` your branch locally and `git push --force`.
- Write a test that verifies this change has the desired effect?

ScopeyNZ:
I've rebased your commit onto the `4.0` branch but it's significantly harder and more error prone when the PR branch called `master` (and also confusing 😂 ).

You've created this pull request from one of the "core branches" rather than a branch specifically intended for this pull request. In the future please create a new branch for each individual pull request. See our [guide about contributing PRs](https://docs.silverstripe.org/en/4/contributing/code/#step-by-step-from-forking-to-sending-the-pull-request).

I'll look at the test failures if those are still happening shortly.

src/Tasks/MigrateContentToElement.php
```diff
@@ -33,6 +34,9 @@ public function run($request)
                 $element->HTML = $content;
                 $element->ParentID = $area->ID;
                 $element->write();
+                if (class_exists(Versioned::class)) {
```
ScopeyNZ:
```suggestion
                if ($page->hasExtension(Versioned::class)) {
```

Versioned is a dependency so it will always exist 🙂

src/Tasks/MigrateContentToElement.php
```diff
@@ -33,6 +34,9 @@ public function run($request)
...
+                    $page->copyVersionToStage(Versioned::DRAFT, Versioned::LIVE);
```
ScopeyNZ:
```suggestion
                    $page->publishRecursive();
```

`publishRecursive` will ensure all owned DataObjects (most importantly the element you just added) are also published.

ScopeyNZ:
I think I'll quickly implement my own suggestions, do some clean-up and write some tests.

src/Tasks/MigrateContentToElement.php
ScopeyNZ: Versioned will always been an extension actually so this condition can just be removed.

ScopeyNZ:
Oh this can be closed now. Sorry @altwohill I hope you don't mind that I turned this into a "rewrite the task" in a separate PR 🙂

PR #641, opened 2019-03-21
FIX Reports linking to pages now use the correct CMS edit link

robbieaverill:
Fixes https://github.com/dnadesign/silverstripe-elemental/issues/629

### DIFF
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -661,20 +661,17 @@ class BaseElement extends DataObject
             return null;
         }
 
-        $editLinkPrefix = '';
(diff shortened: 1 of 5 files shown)


robbieaverill:
Maybe we should just sent both of those links straight to the page, since inline editing is likely to become the default behaviour going forward

robbieaverill:
Alright I've just updated it so it works

robbieaverill:
Technically it’s a breaking change, but I’m ok with it

ScopeyNZ:
`getEditLink()` probably should be `@internal` anyway 😉

PR #642, opened 2019-03-21
FIX Don't explicitly include Title when namespacing fields

### LINKED ISSUE #632: Can't remove or disable Title field
I want to have a fairly simple Photo block which just allows you to choose an image and set a couple of options. Setting a title on this is redundant and unnecessary work for the user because the title of the selected photo can be used automatically.

But trying to remove the Title throws an error, and disabling it doesn't work either :(

(BTW I'm just trying this module out for the first time - it looks pretty nice overall!)

```php
use SilverStripe\Assets\Image;

class PhotoElement extends BaseElement
{
    private static $has_one = [
        'Photo' => Image::class,
    ];

    {

        // This removes the text field nested within the TextCheckboxGroupField and throws Uncaught Error: Call to a member function getName() on null IN GET /admin/elemental-area/schema/elementForm/1 Line 60 in (...) vendor/dnadesign/silverstripe-elemental/src/Forms/EditFormFactory.php
        $fields->removeByName('Title');
        // This doesn't do anything
        $fields->dataFieldByName('Title')->performReadonlyTransformation();

        // Target the TextCheckboxGroupField combo field (it's name is Title but so is the text field within it)
        $titleComboField = $fields->fieldByName('Title'); // No result - doesn't work
        $titleComboField = $fields->dataFieldByName('Title'); // Wrong - returns the text field nested within the TextCheckboxGroupField
        $titleComboField = $fields->findTab('Root.Main')->Fields()->first(); // This does the trick but it's fragile and could break very easily

        // Try to delete the whole field (doesn't work)
        $fields->remove($titleComboField);

        // Try to make the whole field read only at least (doesn't work)
        $titleComboField->performReadonlyTransformation();

        return $fields;
    }

    public function getTitle()
    {
        return $this->Photo()->getTitle();
    }
}
```

ScopeyNZ:
Currently it's difficult to remove the Title field as it's referred to explicitly in the EditFormFactory during "namespacing". We shouldn't be so presumptuous when it comes to the fields that might make up a custom block.

Resolves #632


robbieaverill:
Won't this mean you end up having Title field conflicts between blocks on a page?

ScopeyNZ:
Looks like behat has indicated that I've broken something. 

> Won't this mean you end up having Title field conflicts between blocks on a page?

I don't think so. The field should be namespaced already because it should be a dataField.

I really don't think we should make assumptions on what fields a block has. It currently just breaks if you remove `Title`.

ScopeyNZ:
Ok Behat is pointing out a valid issue. The `for` attribute in the label for the "Title" field does not refer to the correct ID for the actual "Title" field.

ScopeyNZ:
Updated. The JS field now ensures the field ID matches the for attribute given to the label.

PR #643, opened 2019-03-21
FIX Use KeyboardEvent.key as 'code' is unsupported by React

### LINKED ISSUE #626: Keyboard accessibility on block editing
* Focus on an unexpanded block by tabbing into it. I'd expect the ENTER key to expand the block for editing (if it's inline editable), or to open the detail view if it isn't
 * Continue tabbing into the block. The "settings" icon works as expected, but the "expand" button doesn't. It has no outline indicating focus, and when hitting ENTER it loads the "metadata" tab, not the "content" tab.

dnadesign/silverstripe-elemental                  4.0.x-dev dd8e73b

ScopeyNZ:
See facebook/react#11306 for more detail.

React wraps events in "SyntheticEvent" wrappers. For KeyboardEvent the `code` property is unsupported. This PR converts to using the `key` property instead.

Fixes #626


robbieaverill:
Tested, works nice

PR #647, opened 2019-03-23
Add the Italian language

kenlog:
Hi, I'm Italian mother tongue. I took care of the Italian translation. Very useful module :+1:


ScopeyNZ:
Hi @kenlog and thanks for the contribution. Our translations are automatically generated when we release and they are compiled from transifex. Although I can accept your contribution they will be overwritten next time we release.

We would love help translating though which can be done here:

https://www.transifex.com/silverstripe/silverstripe-elemental/translate/#it

You should be able to sign up with your GitHub and request access to translate.

PR #650, opened 2019-04-01
Save repeated database queries for same ElementalArea (closes #648)

### LINKED ISSUE #648: [Performance] Use DataObject::get_by_id() to fetch elements’ ElementalArea
I’ve noticed that on a page with a lot of elements, there are a number of repeated database queries triggered by `BaseElement::getPage()`. They’re cached against each element, but that still means one call per element when the parent is always going to be the same.

Using the following instead of calling `$this->Parent()` directly saves those repeated lookups (and probably means we could remove `$this->cacheData['page']`):

```php
// Replace...
$area = $this->Parent();

// ...with
$class = DataObject::getSchema()->hasOneComponent($this, 'Parent');
$area = DataObject::get_by_id($class, $this->ParentID);
```

kinglozzer:
I left `$this->cacheData['page']` in there for 2 reasons:

1. Playing it safe for BC - it’s not part of the public API, but it’s possible someone could be relying on it
2. It saves repeated schema lookups (I think they’re cached anyway, but it’s still a few less method calls)

### DIFF
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -525,8 +525,9 @@ class BaseElement extends DataObject
             return $this->cacheData['page'];
         }
 
-        $area = $this->Parent();
 
+        $class = DataObject::getSchema()->hasOneComponent($this, 'Parent');
+        $area = ($this->ParentID) ? DataObject::get_by_id($class, $this->ParentID) : null;
         if ($area instanceof ElementalArea && $area->exists()) {
             $this->cacheData['page'] = $area->getOwnerPage();
             return $this->cacheData['page'];


robbieaverill:
The behat failures are pre-existing, but perhaps the Postgres test failures are valid =/

ScopeyNZ:
> perhaps the Postgres test failures are valid

I just reran travis on that job in both this PR and `master`. `master` is still green and this PR is still red 😞

kinglozzer:
Gah that sucks. Why on earth would PGSQL have a problem with such a simple query?! 😖

kinglozzer:
Okay I’ve solved it, by checking whether `$this->ParentID` is set before running the query. I have absolutely no idea why PGSQL threw that error instead of just returning no results, but 🤷‍♂️

Issue #651, opened 2019-04-04
Guidance for migrating page types to content blocks


chillu:
## Overview

As a developer upgrading a SilverStripe site, I want to get a better understanding of the steps to migrate from page types to elemental, in order to assess the risk of suggesting that move to my customer

## Notes

 * There won't be a fully generic task, since website structures and blocks are quite different, but similar to the upgrade guide, I think we can help devs to get started with some example code and a tutorial. For example, how to migrate a "page with banner" template (move over `has_one image` to an image content block, extract template code into a custom image content block template)
 * We have a [MigrateToContentElement](https://github.com/dnadesign/silverstripe-elemental/blob/master/src/Tasks/MigrateContentToElement.php) task, but that's only a stop-gap solution

PR #652, opened 2019-04-05
NEW Overhaul to content migration task

### LINKED ISSUE #617: MigrateContentToElement doesn't find any pages to migrate

ScopeyNZ:
- Fixes #617 - The task now checks for pages with no content instead of no elemental area
- Developers can configure a specific element type and attribute to migrate content too
- Content is now published during the migration
- Added a flag to control whether content is published
- Whether this task clears content from the page during migration is configurable. If it is not cleared out, content will not migrate if the page already has elements.
- Added some extension points and extracted some logic into methods to allow flexible subclassing / extensions

### DIFF
--- a/docs/en/content_migration.md
+++ b/docs/en/content_migration.md
@@ -10,11 +10,11 @@ content for you. It will:
 - Clear the existing page content
 - Save and publish the page and the block if the latest page version was previously published
 
-There are several configuration options and extension hooks to allow customising the functionality of this class
+There are several configuration options and extension hooks to allow customising the functionality of this class.
 
 ### Configuring the element that is created
 
-You may configure which element content is migrated to by using the following configuration
+You may configure which element content is migrated to by using the following configuration:
 
 ```yml
 DNADesign\Elemental\Tasks\MigrateContentToElement:
@@ -53,4 +53,4 @@ configuration option is enabled.
 
 While this task is built for pages that use the `ElementalPageExtension` it is possible to use elemental using only the 
 `ElementalAreaExtension`. In this case you can extend this task and overload the `isMigratable` and 
-`getAreaRelationFromPage` methods to support your use-case
+`getAreaRelationFromPage` methods to support your use-case.
--- a/docs/en/content_migration.md
+++ b/docs/en/content_migration.md
@@ -1,0 +1,56 @@
+# Migrating Content from Pages
+
+After installing elemental and applying it to certain page types - the standard content editor is replaced by a block 
+editor. This means existing content is lost. The `MigrateContentToElement` task is provided that will migrate existing
+content for you. It will:
+
+- Identify pages that have `ElementalPageExtension`
+- Filter out any that have empty content
+- Create a content block (`ElementContent`) and set the content to the existing page content
+- Clear the existing page content
+- Save and publish the page and the block if the latest page version was previously published
+
+There are several configuration options and extension hooks to allow customising the functionality of this class
+
+### Configuring the element that is created
+
+You may configure which element content is migrated to by using the following configuration
+
+```yml
+DNADesign\Elemental\Tasks\MigrateContentToElement:
+  target_element: My\App\Blocks\MyBlock
+  target_element_field: Content
+```
+
+`target_element` specifies the element that will be created and `target_element_field` specifies the data field that 
+will be populated with the content from the page. Additionally you can use the `updateMigratedElement` extension point 
+to further modify the element using the existing content and page.
+
+### Disabling auto publish
+
+You may disable the "auto-publish" functionality of this task with config
+
+```yml
+DNADesign\Elemental\Tasks\MigrateContentToElement:
+  publish_changes: false
+```
+
+### Keeping the content on the page
+
+In some cases you may wish to keep the existing content stored on the `Content` field on pages. Removing it is optional
+and can be turned off with config:
+
+```yml
+DNADesign\Elemental\Tasks\MigrateContentToElement:
+  clear_content: false
+```
+
+With this setting enabled the task can no longer filter out pages that have already been migrated by checking if the
+existing content is empty. Instead the task will not migrate pages that have at least one block already while this 
+configuration option is enabled.
+
+### Working without the `ElementalPageExtension`
+
+While this task is built for pages that use the `ElementalPageExtension` it is possible to use elemental using only the 
+`ElementalAreaExtension`. In this case you can extend this task and overload the `isMigratable` and 
+`getAreaRelationFromPage` methods to support your use-case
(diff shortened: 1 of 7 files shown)


ScopeyNZ:
I'll quickly add some docs updates to address #651

docs/en/content_migration.md
```diff
@@ -0,0 +1,56 @@
...
+editor. This means existing content is lost. The `MigrateContentToElement` task is provided that will migrate existing
+content for you. It will:
+
+- Identify pages that have `ElementalPageExtension`
+- Filter out any that have empty content
+- Create a content block (`ElementContent`) and set the content to the existing page content
+- Clear the existing page content
+- Save and publish the page and the block if the latest page version was previously published
+
+There are several configuration options and extension hooks to allow customising the functionality of this class
```
robbieaverill:
```suggestion
There are several configuration options and extension hooks to allow customising the functionality of this class.
```

docs/en/content_migration.md
```diff
@@ -0,0 +1,56 @@
...
+
+### Configuring the element that is created
+
+You may configure which element content is migrated to by using the following configuration
```
robbieaverill:
```suggestion
You may configure which element content is migrated to by using the following configuration:
```

docs/en/content_migration.md
```diff
@@ -0,0 +1,56 @@
...
+
+With this setting enabled the task can no longer filter out pages that have already been migrated by checking if the
+existing content is empty. Instead the task will not migrate pages that have at least one block already while this 
+configuration option is enabled.
+
+### Working without the `ElementalPageExtension`
+
+While this task is built for pages that use the `ElementalPageExtension` it is possible to use elemental using only the 
+`ElementalAreaExtension`. In this case you can extend this task and overload the `isMigratable` and 
+`getAreaRelationFromPage` methods to support your use-case
```
robbieaverill:
```suggestion
`getAreaRelationFromPage` methods to support your use-case.
```

src/Tasks/MigrateContentToElement.php
```diff
@@ -2,13 +2,50 @@
...
-
+use SilverStripe\CMS\Model\SiteTree;
+use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Dev\BuildTask;
+use SilverStripe\Versioned\Versioned;
 
 class MigrateContentToElement extends BuildTask
 {
+    /**
+     * Configures if the existing content should be left blank.
```
robbieaverill:
```suggestion
     * Configures if the existing content should be cleared once the migration task has completed.
```

I found the "should be left blank" part a little ambiguous

src/Tasks/MigrateContentToElement.php
```diff
@@ -2,13 +2,50 @@
...
+     *
+     * @config
+     * @var bool
+     */
+    private static $clear_content = true;
+
+    /**
+     * The FQN of an element that will be the target of the content
```
robbieaverill: Do you want to add something like "must extend BaseElement" ?

src/Tasks/MigrateContentToElement.php
```diff
@@ -2,13 +2,50 @@
...
+     * The name of the field on the `target_element` where the content should be placed
+     *
+     * @config
+     * @var string
+     */
+    private static $target_element_field = 'HTML';
+
+    /**
+     * Indicates that the updated page and elements should be immediately published (provided the Versioned extension
+     * is present)
```
robbieaverill:
```suggestion
     * is present, and the page was previously in a published state)
```

src/Tasks/MigrateContentToElement.php
```diff
@@ -17,26 +54,113 @@ class MigrateContentToElement extends BuildTask
...
+                    $page->Content = '';
+                }
+
+                // Get the area
+                $area = $this->getAreaRelationFromPage($page);
+
+                // Write the page if we're clearing content or if the area doesn't exist - we write to trigger a
+                // relationship update
+                if ($clearContent || !$area->exists()) {
+                    $page->write();
```
robbieaverill: I think you should do this within a try/catch, incase creating the new element with this content fails and you end up in a state where you have no content at all

robbieaverill:
Further comment: the issue this fixes is triaged as a bug, and I don't see a reason why this couldn't be considered as a patch change. Fancy re-labelling the commit and targeting an existing release line?

src/Tasks/MigrateContentToElement.php
ScopeyNZ: That's already a requirement of elements as `ElementTypeRegistry` looks for subclasses of `BaseElement`. I can add something if you feel strongly though

src/Extensions/ElementalPageExtension.php
```diff
@@ -9,6 +9,10 @@
 use SilverStripe\Core\Config\Config;
 use SilverStripe\View\SSViewer;
 
+/**
+ * @method ElementalArea ElementalArea
+ * @property ElementalArea ElementalArea
```
robbieaverill: I don't think it's a property, is it?

ScopeyNZ: Unary components are also accessible as properties on `DataObject`s 💥

robbieaverill: TIL 👍

ScopeyNZ:
Closed/reopened to bump travis to the correct base branch.

src/Tasks/MigrateContentToElement.php
```diff
@@ -17,26 +55,121 @@ class MigrateContentToElement extends BuildTask
...
+                }
+
+                // Get the area
+                $area = $this->getAreaRelationFromPage($page);
+
+                // Write the page if we're clearing content or if the area doesn't exist - we write to trigger a
+                // relationship update
+                if ($clearContent || !$area->exists()) {
+                    try {
+                        $page->write();
```
robbieaverill: I guess I was more meaning a kind of transaction style operation where migrating the content into an element fails, you roll it back so you don't clear the page content

ScopeyNZ: Oh yep I understand - I'll follow that up later

robbieaverill: I think I'll merge this for now, if you want to do another patch change to add transactional style rollbacks that'd be cool but not blocking merge

PR #665, opened 2019-05-08
FIX Catch exceptions while gathering element summary in ElementsInUseReport

robbieaverill:
This can happen if you have invalid width or height attribute values in image shortcodes in your content. It no longer breaks the report, but skips and logs a debug message.

Fixes https://github.com/dnadesign/silverstripe-elemental/issues/664

### DIFF
--- a/src/Reports/ElementsInUseReport.php
+++ b/src/Reports/ElementsInUseReport.php
@@ -3,6 +3,9 @@
 namespace DNADesign\Elemental\Reports;
 
 use DNADesign\Elemental\Models\BaseElement;
+use InvalidArgumentException;
+use Psr\Log\LoggerInterface;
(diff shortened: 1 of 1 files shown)


src/Reports/ElementsInUseReport.php
```diff
@@ -59,11 +62,21 @@ public function columns()
                     return '<span class="element__note">' . _t(__CLASS__ . '.None', 'None') . '</span>';
                 },
             ],
-            'Summary' => [
+            'ElementSummary' => [
```
robbieaverill: When this was `Summary` it would actually call getSummary() early, then call it again during the try/catch below. This is a minor performance optimisation as well as essential for the bug fix.

PR #671, opened 2019-05-23
Fix incorrect ElementalArea.OwnerClassName on save

bcairns:
OwnerClassName in the ElementalArea table can become incorrect in certain scenarios*, and there is currently no way to fix this short of manually editing the database.  An incorrect OwnerClassName breaks the behaviour of BaseElement->getPage(), which breaks all kinds of other things in BaseElement that depend on it.

For example BaseElement->canView() will always return false.  I suspect a lot of people are working around this issue by putting canView{ return true } in all their block types.  That's what we were doing until I decided to really dig into the root cause of why BaseElement->canView() was failing.

This change modifies ElementalAreasExtension->ensureElementalAreasExist(), to verify that the OwnerClassName is correct, and re-saves it with the correct value if it is wrong.

*For example, apply ElementalPageExtension to Page, create a Page, then switch ElementalPageExtension to OtherPage, and switch your created page type to OtherPage.  The OwnerClassName in the DB will still be "Page" and there is no way to fix this currently.

### DIFF
--- a/src/Extensions/ElementalAreasExtension.php
+++ b/src/Extensions/ElementalAreasExtension.php
@@ -257,6 +257,13 @@ class ElementalAreasExtension extends DataExtension
                 $area->OwnerClassName = get_class($this->owner);
                 $area->write();
                 $this->owner->$areaID = $area->ID;
+            }else{
+                // ensure correct OwnerClassName
+                $area = ElementalArea::get_by_id($this->owner->$areaID);
+                if( $area->OwnerClassName !== get_class($this->owner) ){
+                    $area->OwnerClassName = get_class($this->owner);
+                    $area->write();
+                }
             }
         }
         return $this->owner;


ScopeyNZ:
Thanks for the patch @bcairns! There'll be a bunch of linting issues - you can probably run `phpcbf` across your code to clear that up.

src/Extensions/ElementalAreasExtension.php
```diff
@@ -257,6 +257,13 @@ public function ensureElementalAreasExist($elementalAreaRelations)
                 $area->OwnerClassName = get_class($this->owner);
                 $area->write();
                 $this->owner->$areaID = $area->ID;
+            }else{
+                // ensure correct OwnerClassName
+                $area = ElementalArea::get_by_id($this->owner->$areaID);
+                if( $area->OwnerClassName !== get_class($this->owner) ){
+                    $area->OwnerClassName = get_class($this->owner);
+                    $area->write();
+                }
```
ScopeyNZ: I think this doesn't really make sense in the context of `ensureElementalAreasExist`. We can't rename the method (which i'd prefer) but we can make a new method. Would you mind doing that?

michalkleiner:
@bcairns would you have time to finalise this? It's nearly there!

ScopeyNZ:
Thanks for the reminder @michalkleiner . We spoke on Slack and I offered to finalise this PR. I haven't had a chance to get around to it yet though.

lbradley-mm:
I've created a new PR from the latest **4** branch
I've done the code in outside of `::ensureElementalAreasExist()` as suggested

Cheddam:
#816 has been merged, so closing this one out.

PR #680, opened 2019-06-14
FIX Upgrade vulnerable dev dependencies

Cheddam:
Ensures `handlebars`, `fstream` and `js-yaml` are up to date.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,7 @@ before_script:
   # Init PHP
   - phpenv rehash
   - phpenv config-rm xdebug.ini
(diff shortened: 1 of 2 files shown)


robbieaverill:
Fancy merging the branches up, then rebasing this? It will have conflicts with #679. You may also want to add something like `  - echo 'memory_limit = 2G' >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini` to the `before_script` section of .travis.yml to fix the builds, somewhere before `composer install` is run

Cheddam:
@robbieaverill All merged up! Looks like Yarn's lockfile structure is working its magic - there's no conflicts 👍

Issue #347, opened 2018-08-23
RFC: Restructure data models to use more default relationship handling


robbieaverill:
I've had a quick look through the data model relationships in this module.

ElementalArea defines `OwnerClassName` and handles its value internally. This should really be at polymorphic has one relationship.

This is presumably so that things other than pages can have Elemental applied to them, for example if a dev was using DataObjects as pages and not using the CMS.

The owner (page etc) should also have the ability to define extra has_one ElementalAreas to add new areas.

The way I see it is:

```
- Page (or other)
  - has_one "ElementalArea" (default)
    - has_one "Owner" (inverse relationship - polymorphic)
    - has_many "Elements"

- BaseElement
  - has_one "Parent" ElementalArea
```

The API should support us accessing:

* the page from the element (via the elemental area or parent elements) e.g. `$page = $element->getPage()` (pre-existing API)
* the page from the elemental area e.g. `$page = $elementalArea->getPage()` (currently is called getOwnerPage, polymorphic relationship would make it `Owner()`)
* the elemental area from the page e.g. `$page->ElementalArea()`
* the elements fro the page (via elemental area) e.g. `$elements = $page->ElementalArea()->Elements()`

Also the terminology - in some places we refer to "Owner" and others we refer to "Page". These should be standardised to "Owner" (wary of potential Extension conflicts with extending this code though). Another option `Container()`.

I guess my RFC is to remove the custom Owner handling in ElementalArea and make it a polymorphic has one. This should be reasonably backwards compatible, but we also have the opportunity right now to make breaking changes in master for elemental 4.x.

robbieaverill:
Basically, there is a fair amount of technical debt and responsibility in methods such as `ElementalArea::getOwnerPage()` that could be alleviated by using out of the box ORM relationships in this instance

ScopeyNZ:
Huge 👍 from me. I think this is a great time to essentially _fix_ this.

NightJar:
More core; more default; more betterer 👍

chillu:
Yep, we should use core features where available. The `owner` terminology is also used for `DataExtension`, so if you're using an extension on a block, you'll end up with `$this->getOwner()->getOwner()`, which is a bit weird. But that can be make clearer through `$block = $this->getOwner(); $page = $block->getOwner();` so I'm not too concerned.

Does this have performance implications? If so, I'd consider this `impact/high` in terms of prioritisation.

robbieaverill:
[...]

PR #685, opened 2019-06-28
FIX Allow direct access to set page cache for performance fixes

ScopeyNZ:
Pushing the boundaries of "patch" for this "fix". At the moment without #347 there's some relatively inefficient relation querying going on. In custom code that creates lists of blocks we can often set the page ahead of time as we know it. This type of code is a performance fix that this patch enables. Hopefully that's enough :sweat_smile:

### DIFF
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -491,6 +491,16 @@ class BaseElement extends DataObject
     }
 
     /**
+     * Set the page in this elements internal cache. This should be used with caution.
+     *
+     * @param DataObject $page
+     */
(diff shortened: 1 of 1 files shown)


src/Models/BaseElement.php
```diff
@@ -490,6 +490,16 @@ public function getPage()
         return null;
     }
 
+    /**
+     * Set the page in this elements internal cache. This should be used with caution.
+     *
+     * @param SiteTree $page
+     */
+    public function setCachePage(SiteTree $page)
```
robbieaverill: Does it have to be a SiteTree?

PR #693, opened 2019-07-17
FIX Element modifications are no longer persisted between page changes with PJAX

### LINKED ISSUE #605: Changes persist between page browsing and breaks page level change tracking
"Changes persist between page browsing" probably doesn't sound like a bug but here's the current process:

- Open a blocks page, choose to edit an in-line editable block, and make some changes
- Note that the page level change tracker (the save button) indicates changes have been made
- Without saving or publishing, browse to a different page
- Note that you get a "You have unsaved changes... They'll be discarded" message
- Accept the prompt and continue to the other page
- Browse back to your original page
- Note that the unsaved changes remain on the block you edited.
- Now the "Save" button implies there are no changes
- The page level change tracker never updates (and will actually have a flipped state where it will only show changes when you manually revert your changes 😅 )

Having the changes persist is pretty awesome. Redux stores the form state (and the page never actually changed because we use pjax for page navigation) 

This isn't very good UX though.

I think we should revert changes when they browse away from the page - this should fix the page level tracking problem too.

I considered the possibility of not showing the unsaved changes message, but considering the other edit forms in SS this would be unusual. Additionally fields outside of the elemental area would still have their state lost.

robbieaverill:
This ensures that the form state and schemas are cleared when navigating away from a page with elements on it, meaning that the modified state will not persist when you navigate back to the original page again.

"But you shouldn't be manipulating the state directly" I hear you say - unfortunately the reducers that initiate the state we need to clear belong to admin, and don't provide an action to clear them. I've raised an issue to add that: https://github.com/silverstripe/silverstripe-admin/issues/909

This is a bit hacky but it's all we can do as a patch change at the moment.

Fixes #605


ScopeyNZ:
Can't we use `redux-form` actions for this? See:

https://github.com/dnadesign/silverstripe-elemental/blob/707cd95b01bf88cf1ca22aa49c144d9445b076eb/client/src/legacy/ElementEditor/entwine.js#L61-L63

robbieaverill:
I had done it directly because I figured we wouldn't be able to clear the schemas in the store as well, but I suppose it's not possible for the UI to modify the form schema so it doesn't matter. I'll see if I can adjust this PR to use the redux form action instead.

robbieaverill:
Cool, PR updated

Issue #655, opened 2019-04-18
TinyMCE Inline Toolbar placement


christopherdarling:
SilverStripe 4.3.3

When using a `HTMLEditorField` inside a Elemental block, the placement of the inline toolbar (https://github.com/silverstripe/silverstripe-admin/blob/0f0a0a064bf4a10f4122795fb56ecf69a1da65cd/client/src/components/TinymceInlineToolbar/TinymceInlineToolbar.js) breaks when the link is not the first paragraph. This only appears to be an issue when using the inline editing feature of Elemental (issue not present when using the field directly on Page or via a typical GridField). For example;

**Correct placement** - when link is first paragraph
![image](https://user-images.githubusercontent.com/178039/56349835-51c5a300-61c1-11e9-8110-bf155c0c7ba3.png)

**Incorrect placement** - when link is on a second paragraph (see the bottom of the popup near the cursor at the top of the screenshot)
![image](https://user-images.githubusercontent.com/178039/56349820-48d4d180-61c1-11e9-9d0e-061497347b48.png)

robbieaverill:
Reproduced, I think you need to be in the mobile viewport to experience this

christopherdarling:
Thanks Robbie, I can confirm that this was also on desktop

robbieaverill:
I can also confirm that this is an elemental specific problem, not happening on regular page types

robbieaverill:
PR at https://github.com/silverstripe/silverstripe-admin/pull/910. I've tested the popup on each possible line until you start scrolling, and it isn't consistent, but IMO that's much lower value than the fact that it was off the screen originally.

See the PR for screenshot examples.

@christopherdarling do you feel strongly enough that we should raise a new issue to tweak those positioning rules?

christopherdarling:
@robbieaverill I haven't had chance to test the PR but the screenshots look like a definite improvement.

Am I right in my understanding from your description on the PR that the issue (_that I raised here, and the positioning issue you mentioned that's still present on the PR_) only occurs in HtmlEditorField's with < 6 rows? If that's the case, perhaps change https://github.com/dnadesign/silverstripe-elemental/blob/6e04835688f3d45af21fa381e49c9cae0c647fc7/src/Forms/EditFormFactory.php#L40 to 6 or greater?

robbieaverill:
[...]

PR #695, opened 2019-07-17
FIX Drop the TinyMCE toolbar but increase the row count

ScopeyNZ:
This fixes an issue where the edit link popover didn't really have room to position itself and would end up appearing off-screen or covering the link you were editing

### DIFF
--- a/src/Forms/EditFormFactory.php
+++ b/src/Forms/EditFormFactory.php
@@ -36,16 +36,18 @@ class EditFormFactory extends DefaultFormFactory
         $fields = parent::getFormFields($controller, $name, $context);
 
         // Configure a slimmed down HTML editor for use with blocks
-        /** @var HTMLEditorField $editorField */
+        /** @var HTMLEditorField|null $editorField */
         $editorField = $fields->fieldByName('Root.Main.HTML');
-        $editorField->setRows(7);
+        if ($editorField) {
+            $editorField->setRows(7);
 
-        $editorConfig = $editorField->getEditorConfig();
+            $editorConfig = $editorField->getEditorConfig();
 
-        // Only configure if the editor is TinyMCE
-        if ($editorConfig instanceof TinyMCEConfig) {
-            $editorConfig->setOption('statusbar', false);
-            $editorField->setEditorConfig($editorConfig);
+            // Only configure if the editor is TinyMCE
+            if ($editorConfig instanceof TinyMCEConfig) {
+                $editorConfig->setOption('statusbar', false);
+                $editorField->setEditorConfig($editorConfig);
+            }
         }
 
         return $fields;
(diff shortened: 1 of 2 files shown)


src/Forms/EditFormFactory.php
```diff
@@ -30,19 +30,6 @@ public function getForm(RequestHandler $controller = null, $name = self::DEFAULT
...
-
-        /** @var HTMLEditorField $contentField */
-        $contentField = $fields->fieldByName('Root.Main.HTML');
-        if ($contentField) {
-            $contentField->setRows(5);
-        }
-
-        return $fields;
-    }
-
```
ScopeyNZ: This isn't breaking because it's overriding a parent method.

ScopeyNZ:
This probably shouldn't be merged unless this is actually decided it's needed (see #655 )

src/Models/ElementContent.php
```diff
@@ -29,9 +31,20 @@ class ElementContent extends BaseElement
     public function getCMSFields()
     {
         $this->beforeUpdateCMSFields(function (FieldList $fields) {
-            $fields
-                ->fieldByName('Root.Main.HTML')
-                ->setTitle(_t(__CLASS__ . '.ContentLabel', 'Content'));
+            // Configure a slimmed down HTML editor for use with blocks
+            /** @var HTMLEditorField $editorField */
+            $editorField = $fields->fieldByName('Root.Main.HTML');
+            $editorField->setRows(7);
```
robbieaverill: The reason it was in the original place is because it then only affects inline editing. This would also affect non inline editable elements, which naturally have a whole window to fill so a default size would be acceptable

ScopeyNZ: Ah yep of course. I even did consider adding a `if ($this->inlineEditable())` statement in there! Ok I'll restore this back to it's original spot and I might as well move the status-bar removal there.

Cheddam: Adjusted 👍

ScopeyNZ: 🙏

src/Forms/EditFormFactory.php
```diff
@@ -34,10 +35,17 @@ protected function getFormFields(RequestHandler $controller = null, $name, $cont
...
         $fields = parent::getFormFields($controller, $name, $context);
 
-        /** @var HTMLEditorField $contentField */
-        $contentField = $fields->fieldByName('Root.Main.HTML');
-        if ($contentField) {
-            $contentField->setRows(5);
+        // Configure a slimmed down HTML editor for use with blocks
+        /** @var HTMLEditorField $editorField */
+        $editorField = $fields->fieldByName('Root.Main.HTML');
+        $editorField->setRows(7);
```
robbieaverill: I think you need to put the if statement back - this form factory applies to all blocks, not just content blocks - you may get a null pointer exception

robbieaverill: I've updated this

PR #701, opened 2019-07-25
NEW Page level history now links directly into an elements history view

robbieaverill:
This is a partial revert of https://github.com/dnadesign/silverstripe-elemental/pull/436

It contains a dirty hack, which works around https://github.com/silverstripe/silverstripe-admin/issues/911

Resolves https://github.com/dnadesign/silverstripe-elemental/issues/223


src/Models/BaseElement.php
```diff
@@ -1005,4 +1040,22 @@ public function EvenOdd()
...
+        $this->disableInlineEditing = (bool) $disable;
+        return $this;
+    }
+
+    /**
+     * @return bool
+     */
+    public function getDisableInlineEditing()
+    {
+        return (bool) $this->disableInlineEditing;
```
ScopeyNZ: Is there a way we can avoid having this temporary state property on `BaseElement`. I can see bugs in the future where strange things could happen if the in-line edit setting isn't toggle back on. Can we add `$directLinkOnly` or something as a parameter to `CMSEditLink`?

ScopeyNZ:
I've pushed a commit that resolves my comment. I've tested the auto-switching to history tab "hack" and it seems to work well. If you're happy with my changes then I think we're good to merge 🙂

Cheddam:
Ultimately, if we're deciding between...

- Adding a temporary layer of config state so we can adjust it before and after calling a method to push it into providing the right answer without context, or
- Passing a flag to that method so it _knows_ to output the right answer

...I personally feel the latter is tidier, particularly if we want to use the method in this way more widely in the future.

However, if we want to avoid flag-based methods, perhaps a third option is adding a second method that _explicitly_ outputs the non-inline-editable URL?

ScopeyNZ:
`CMSEditLink` is a method on `CMSPreviewable` but `BaseElement` does not implement it. Maybe that's the problem? In which case @Cheddam's suggestion is probably best - we have an alternate method - then we don't break the method signature if we decide to implement the `CMSPreviewalbe` interface in the future.

robbieaverill:
Yeah it's also a silently accessed API in a bunch of places in the CMS that don't type check CMSPreviewable... we should tighten it up in SS5

Issue #485, opened 2018-10-31
Loading indicator should be specific to block


brynwhyman:
The loading indicator within the ElementalArea is shown in the middle. For loads specific to one block, it would be nice to see a specific loading indicator contained within the block instead.

Scenarios that I can think of are:
- Expanding a block
- Adding a new block

robbieaverill:
I think we need to bump priority on this. If you have a large site, you click to add a new content block, it can do nothing for 2-3 seconds at least before you see the new block.

Even if we added a "Please wait" row or something, that'd be better than nothing.

PR #703, opened 2019-07-26
FIX Removing loading indicators by introducing optimistic deletes

ScopeyNZ:
Partially addresses #485 - ensuring loading indicators don't appear for the whole elemental area. In the case of an error (such as 403) it will re-show the block.


emteknetnz:
@ScopeyNZ This pull request hasn't had any activity for a while.  Looks like travis is red.  Also it's on a pretty old version of elemental so possibly consider rebase/retarget to 4.4. Are you going to be doing further work on this PR, or would you prefer to close this PR now?

robbieaverill:
It's a real shame we let things like this go stale 😞

emteknetnz:
It seems like there's going to be no further activity on this pull request, so we've closed it for now. Please open a new pull-request if you want to re-approach this work, or for anything else you think could be fixed or improved.

michalkleiner:
> It's a real shame we let things like this go stale 😞

Totally agree!

PR #710, opened 2019-08-02
DOCS Searching blocks

chillu:
Similar info to https://github.com/silverstripe/cwp/blob/master/docs/en/02_Features/01_Solr_search/03_Basic_usage.md#using-multiple-indexes

Related to https://github.com/dnadesign/silverstripe-elemental/issues/709

### DIFF
--- a/docs/en/searching-blocks.md
+++ b/docs/en/searching-blocks.md
@@ -1,0 +1,32 @@
+# Searching Blocks
+
+## Overview
+
+Content created through SilverStripe can be searched in various ways.
+A popular option is the [silverstripe/fulltextsearch](https://github.com/silverstripe/silverstripe-fulltextsearch)
+module, which integrates with search services such as Solr.
+
+## Usage
+
+When you install the fulltextsearch module, it will auto-discover
+an index provided by elemental: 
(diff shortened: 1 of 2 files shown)


docs/en/searching-blocks.md
```diff
@@ -0,0 +1,32 @@
...
+In many cases, you'll already have configured your own search index.
+Since the `ElementalSolrIndex` is auto-discovered, it duplicates
+search indexing effort even when it is unused.
+
+You can disable it via YAML config in favour of your own index definition:
+
+```yml
+SilverStripe\FullTextSearch\Search\FullTextSearch:
+  indexes:
+    - MyCustomIndex
```
robbieaverill: Wouldn't this just merge MyCustomIndex into whatever was already configured?

chillu: It would if there's other explicit definitions of this config array, but usually there aren't. Note that this advice is modeled off the existing docs in https://github.com/silverstripe/cwp/blob/master/docs/en/02_Features/01_Solr_search/03_Basic_usage.md#using-multiple-indexes. See https://github.com/silverstripe/silverstripe-fulltextsearch/blob/3/src/Search/FullTextSearch.php#L46.

PR #712, opened 2019-08-02
API Remove ElementalSolrIndex

chillu:
Marks master as 5.x, since this is the first API change.
See https://github.com/dnadesign/silverstripe-elemental/issues/709

Incorporates another PR against 4.x in order to modify what it contains ;)
https://github.com/dnadesign/silverstripe-elemental/pull/710

We need to create a `4` branch after merging this PR.


composer.json
```diff
@@ -34,7 +34,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "4.x-dev"
+            "dev-master": "5.x-dev"
```
robbieaverill: Before merging, we should create a `4` branch

src/Search/ElementalSolrIndex.php
```diff
@@ -1,25 +0,0 @@
...
-
-if (!class_exists(SolrIndex::class)) {
-    return;
-}
-
-/**
- * Provides ability to index Elemental content for a page, so it can be returned in the context of the page
- * that the elements belong to
- */
-class ElementalSolrIndex extends SolrIndex
```
robbieaverill: You could also just deprecate this without deleting it here or changing the master alias

robbieaverill: Done in #711

PR #717, opened 2019-08-14
FIX Reorder block actions

maxime-rainville:
This simply reorders the option in the action menu.

I need some design feedback from @clarkepaul about whatever we want to call the archive action "Archive" or "Unpublish and archive".

# Parent issue
* https://github.com/dnadesign/silverstripe-elemental/issues/707


robbieaverill:
The test/functional change for unpublish and archive is separate: https://github.com/dnadesign/silverstripe-elemental/issues/707#issuecomment-517549656

PR #724, opened 2019-08-22
Make the HoverBar more visible

maxime-rainville:
I've re factored the HoverBar to remove all the timeout logic. I've added a HoverBar at the start of the Elemental area as well. I made the HoverBar focusable and I added an aria label for accessibility.

# Parent issues
* https://github.com/dnadesign/silverstripe-elemental/issues/704
* https://github.com/dnadesign/silverstripe-elemental/issues/627


maxime-rainville:
Need to fix some IE issue.

maxime-rainville:
IE is happy. Ready for review.

client/src/components/ElementEditor/HoverBar.js
```diff
@@ -4,6 +4,45 @@ import PropTypes from 'prop-types';
...
+  elementId,
+  areaId,
+  popoverOpen,
+  onToggle }) {
+  const hoverBarClassNames = classNames('font-icon-plus-circled', 'hover-bar__line');
+  const areaClasses = classNames(
+    'hover-bar__area',
+    { 'hover-bar__area--focus': popoverOpen }
+  );
+  const label = i18n._t('ElementAddNewButton.ADD_BLOCK', 'Add block');
```
maxime-rainville: I use the same label as the AddNewButton. We might want to add something more contextual like "Insert block".

maxime-rainville:
https://youtu.be/hBRMkuMR0cM

client/src/components/ElementEditor/HoverBar.js
```diff
@@ -12,151 +51,29 @@ import { elementTypeType } from 'types/elementTypeType';
...
   }
 
-  /**
-   * Handle mouse entering the hover area between two elements. Calculates if mouse entered via
-   * 'delayed' or 'instantaneous' hover area and 'activates' these areas with a time out in the
-   * former case and instantaneously in the latter.
-   *
-   * @param event
-   */
-  handleMouseEnter(event) {
```
robbieaverill: I assume this is a breaking change, can you deprecate it instead?

client/src/components/ElementEditor/HoverBar.scss
```diff
@@ -1,65 +1,82 @@
-.element-editor__add-block-area-container {
+$transition: all ease .2s;
+
+.hover-bar {
```
robbieaverill: Can you make this more unique? Maybe something like `.element-editor__hover-bar`

maxime-rainville: Presumably, this is to avoid having the CSS bleed into other part of the CMS?

robbieaverill: Yeah, but it also follows the existing naming convention for other parts of elemental

client/src/components/ElementEditor/HoverBar.js
maxime-rainville: This method was meant to track the internal state of the HoverBar. I don't think this information was meant to be communicated back to any outside component. So if ES6 supported `private/protected/public` keywords, this method would have been `private`. In this context, I think it's safe to remove it altogether.

If I may jump on my favourite hobby horse, I would point out that TypeScript does support function access modifiers. Adopting TS would take the ambiguity out of this kind of front end work and allow us to make it clear what we consider to be part of our public API and what we consider to be internal logic.

robbieaverill: Alright - we still haven't formally introduced React into our semver commitments yet anyway, so kill it...

maxime-rainville: Die Die Die `handleMouseEnter` method!!! :smiling_imp:

client/src/components/ElementEditor/HoverBar.scss
maxime-rainville: Will do ... wasn't aware of this convention in elemental.

client/src/lib/prefixClassNames.js
```diff
@@ -0,0 +1,33 @@
+import classNames from 'classnames';
+
+/**
+ * Utility method wrap around classeNames to prefix the return class names
+ * @param cssPrefix
```
robbieaverill:
```suggestion
 * @param {string} cssPrefix
```

client/src/lib/prefixClassNames.js
```diff
@@ -0,0 +1,33 @@
...
+ * @returns {Function}
+ */
+const prefixClassNames = (cssPrefix) => (...args) => {
```
maxime-rainville: This is inspired by https://www.npmjs.com/package/classnames-prefix.

Felt a bit overkill to require an extra package for such a simple method. If we feel this is not the right place for this kind of method, I'm happy to remove it and hard code `element-editor__hover-bar` into the class names.

client/src/components/ElementEditor/HoverBar.scss
maxime-rainville: Done.

maxime-rainville:
I've centred the (+) icon.

client/src/lib/prefixClassNames.js
robbieaverill: Yeah I thought the same... I don't mind too much but I probably would've leaned towards writing them out before you'd added - now that it's there it can probably stay

maxime-rainville:
> LGTM, but I do think the fade in should be 100ms instead of 200ms. For someone who moves their mouse around quickly, they may never notice the hover bar at all.

The original version had a much longer delay. I think the point of the delay is to avoid having the hover bar pop up everywhere has you move your mouse across the elemental area.

It's a compromise between making the feature easy to discover and having this bar jumping at you all the time. If we make the delay so small that it's absolutely impossible to miss it, then we'll probably end up having this aggressive blue bar irritatingly popping everywhere.

client/src/components/ElementEditor/HoverBar.scss
```diff
@@ -1,65 +1,80 @@
...
-  width: 100%;
-  -webkit-appearance: button;
-  border: 0;
-  position: absolute;
-  display: flex;
-  justify-content: flex-end;
-  padding: 0;
-  top: 50%;
-  transform: translateY(-66%);
+    &__area {
```
robbieaverill: I think this level of nesting violates BEM a little... you should be able to list these separately, e.g.

```scss
.element-editor {
  &__hover-bar { ... }
  &__area { ... }
}
```

We try and keep nesting levels to a minimum wherever possible (extreme example is the code you've deleted in this file). The next block could also be removed, e.g. `&__area { &-inner { ... } }` becomes `&__area { ... } &__area-inner { ... }`.

robbieaverill:
If we're being picky, there's probably a pixel or so still needing to be moved in the alignment:

image

client/src/components/ElementEditor/HoverBar.scss
maxime-rainville: I'm happy to fix it as suggested, but isn't the underlying problem here that all those components are prefixed with element-editor?

It's probably too late to reconsider this, but I think it would make more sense to say that the root of the component should be the top block and parts of the component becomes elements and get the `__` seperator.

robbieaverill: Yeah I think that's a conversation for another issue - we can't change these class names any more without breaking things for people (potentially)

maxime-rainville:
I've added one pixel. It seems to make things a little bit better in chrome and a little bit worst in Firefox.

robbieaverill:
One quirk in IE11, when you tab into it and press enter it opens the new block dialog and flashes briefly (might be a reload of something), then the focus is lost and when you close the dialog (esc key) you continue to tab from the top of the screen again. Might be a separate issue, but worth mentioning.

maxime-rainville:
So when you escape the popover in IE11, you would want the focus to be on the same hover bar?

robbieaverill:
Yeah correct, this is how Chrome behaves for me

maxime-rainville:
Firefox acts more or less the same as a Chrome. 

Interestingly, the popover on the green "add block" button seems to return the focus to the start of the page in all browsers. We might need to add an `onEscape` handler a to the `AddElementPopoverComponent` that controls where the focus goes when you hit the escape key.

robbieaverill:
Ok I'll raise another issue to track this - you should be able to spam enter then esc to toggle repeatedly

PR #730, opened 2019-09-01
set $link to null if $page is not present, don't return

jules0x:
Our site needed to allow some elements (elemental-list) to be created in a ModelAdmin, outside of a page, then be virtualised using silverstripe-elemental-virtual

In order to do this, we needed an extension to be able to use the hook on the BaseElement in the CMSEditLink() function to rewrite the URLS.
The reason for this is that child elements of these lists attempt to find an owning page and navigate to it in order to be edited, but in our case, there is no owning page. Instead, we use the hook to change the URL pattern so the child elements can be edited in the context of their parent element, which is in the ModelAdmin.

The problem is that if a page was not detected, then null was returned, so the hook was never reached.

If the $link is set to null and the following logic takes it into account before trying to use $page, then we can override the CMSEditLink() functionality.

Unsure which branch to target

### DIFF
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -682,17 +682,17 @@ JS
         $page = $this->getPage();
 
         if (!$page) {
-            return null;
+            $link = null;
         }
 
         if (!$page instanceof SiteTree && method_exists($page, 'CMSEditLink')) {
             $link = Controller::join_links($page->CMSEditLink(), 'ItemEditForm');
-        } else {
+        } elseif ($page) {
             $link = $page->CMSEditLink();
         }
 
         // In-line editable blocks should just take you to the page. Editable ones should add the suffix for detail form
-        if (!$this->inlineEditable() || $directLink) {
+        if ($page && !$this->inlineEditable() || $directLink) {
             $link = Controller::join_links(
                 singleton(CMSPageEditController::class)->Link('EditForm'),
                 $page->ID,


ScopeyNZ:
I've noticed that you've created this pull request from one of the "core branches" rather than a branch specifically intended for this pull request. This can cause problems if we ask you to target this PR at a different branch. No action is needed at the moment, this is for your information.

In the future please create a new branch for each individual pull request. See our [guide about contributing PRs](https://docs.silverstripe.org/en/4/contributing/code/#step-by-step-from-forking-to-sending-the-pull-request).

---

I'd prefer that we just copy the `->extend` line to where we do the return. I think it reads a lot nicer rather than having conditional logic on page throughout the rest of the method.

This would mean just changing that one statement to

```php
if (!$page) {
    $this->extend('updateCMSEditLink', $link);
	return $link;
}
```

I think you should target this at the `4` branch. `master` is for the next major release of elemental.

ScopeyNZ:
Given the name of the branch matches a core branch you might find it easier to raise another pull request based off of the `4.0` branch rather than rebase this one.

jules0x:
Slight variation to your suggestion as $link was not yet available...

https://github.com/dnadesign/silverstripe-elemental/pull/731

ScopeyNZ:
I ommitted it because you can create variables if the arguments are accepted by reference with default values 😉 . It probably reads nicer with it being declared first though.

Thanks for the contribution. I've merged the other PR so I'll close this and merge up the fix to newer branches.

PR #732, opened 2019-09-03
Add config to replace content field

### LINKED ISSUE #410: Stop replacing the default Content field
https://github.com/dnadesign/silverstripe-elemental/blob/265b9481af13c2eb34f817bb3308ba937eeb634d/src/Extensions/ElementalAreasExtension.php#L151

This line is no longer necessary, and produces unexpected behaviour when someone wishes to have both a Content field (which may not necessarily be an `HTMLEditorField`) on a Page (or DataObject) that has Elemental applied.

This seems to me to be a hang-over from the way things used to be... but now Elemental is more versatile and can be applied in many more situations than to pre-existing pages... I think this can be done away with - or at least moved to into another extension.

ichaber:
Add configuration to allow not replacing the 'Content' field to enable backwards compatibility.
This config would allow to use the default content field added in SiteTree and therefore enable a migration from older content blocks modules.

Resolves #410

### DIFF
--- a/src/Extensions/ElementalAreasExtension.php
+++ b/src/Extensions/ElementalAreasExtension.php
@@ -72,6 +72,14 @@ class ElementalAreasExtension extends DataExtension
     private static $sort_types_alphabetically = true;
 
     /**
+     * Whether or not to replace the default SiteTree content field
+     *
+     * @var boolean
+     * @config
+     */
+    private static $replace_content_field = true;
+
+    /**
      * Get the available element types for this page type,
      *
      * Uses allowed_elements, stop_element_inheritance, disallowed_elements in
@@ -170,7 +178,9 @@ class ElementalAreasExtension extends DataExtension
 
         // add an empty holder for content as some module explicitly use insert
         // after content.
-        $fields->replaceField('Content', new LiteralField('Content', ''));
+        if (Config::inst()->get(ElementalAreasExtension::class, 'replace_content_field')) {
+            $fields->replaceField('Content', new LiteralField('Content', ''));
+        }
         $elementalAreaRelations = $this->owner->getElementalRelations();
 
         foreach ($elementalAreaRelations as $eaRelationship) {
--- a/src/Extensions/ElementalAreasExtension.php
+++ b/src/Extensions/ElementalAreasExtension.php
@@ -73,11 +73,13 @@ class ElementalAreasExtension extends DataExtension
 
     /**
      * Whether or not to replace the default SiteTree content field
+     * Applies globally, across all page types; unless a page type overrides this with its own config setting of
+     * `elemental_keep_content_field`
      *
      * @var boolean
      * @config
      */
-    private static $replace_content_field = true;
+    private static $keep_content_fields = false;
 
     /**
      * Get the available element types for this page type,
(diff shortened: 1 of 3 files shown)


ScopeyNZ:
Do you want to add a note to the content migration docs?

https://github.com/dnadesign/silverstripe-elemental/blob/4/docs/en/content_migration.md

Additionally, `master` is for the next major release of elemental. Do you want to target this at `4`?

ichaber:
Good point @ScopeyNZ. That could probably go into the 4 branch as well, because I tried to keep it backwards compatible.
I'll also update the documentation and commit that onto the same branch.

NightJar:
Added options for both global settings and per-application (of the extension - so class level, not instance level).
Tests included.
Docs updated.

src/Extensions/ElementalAreasExtension.php
```diff
@@ -168,9 +178,12 @@ public function updateCMSFields(FieldList $fields)
             return;
         }
 
-        // add an empty holder for content as some module explicitly use insert
-        // after content.
-        $fields->replaceField('Content', new LiteralField('Content', ''));
+        // add an empty holder for content as some module explicitly use insert after content
+        $globalReplace = !Config::inst()->get(self::class, 'keep_content_fields');
```
Firesphere: I'd not use the `!` here, but in the if statement below. It makes more sense to assign a variable to its actual value and check in the if instead of the other way around

src/Extensions/ElementalAreasExtension.php
```diff
@@ -168,9 +178,12 @@ public function updateCMSFields(FieldList $fields)
...
+        $classOverride = Config::inst()->get(get_class($this->owner), 'elemental_keep_content_field');
```
ScopeyNZ: The only minor feedback I have is that I'd call this just `keep_content_field`. You could actually probably just call both config settings `keep_content_field` - the plural doesn't really make much sense unless it would affect multiple content fields.

NightJar: The global one would affect multiple fields, was my thought there. But not on the same model object, yeah.

I used the prefix for the singular option to indicate it's meaning when assigned to an otherwise arbitrary model - a sort of namespace I guess. It has irked me in the past when I've looked up a model on a project to see a whole bunch of config values that I have no idea where they're coming from (particularly where multiple `DataExtension`s are involved), so I'd like to give a developer at least some indication in what it is for (rather than having to search a code base to see where the config value is referenced).

I also like the distinction between "this affects everything" and "this affects _this_ thing" when using the config key in conversation. I think it could reduce confusion and cognitive load.
But your counter argument works equally as well in regards to adding cognitive load with different names, I guess.

Do you feel strongly about this?

src/Extensions/ElementalAreasExtension.php
NightJar: I did this due to the name of the variable and the name of the config settings basically being opposites in intention. I made it this way to be easier to process (in a human brain).

I termed the config value `keep` in order to reflect the default better - the absence of the config value (i.e. `null`) should be in line with the default behaviour I thought. This makes sense on individual models (page types) where there won't be a setting, unless a developer specifically sets one. I then transferred this to the global value as well (which was previously 'replace') in order to reduce cognitive load with the two config values (for global and individual models) being opposite to each other. I thought it easier to align one's brain around the logic if both were termed with the same intention.

I found the variable name (which is opposite behaviour to the config value) to make more sense in the if statement, in a form of self documenting code. I found it difficult to name things in a way that made sense when it came to the if statement referring to 'keep' settings guarding the _remove_ statement, coupled trying to keep the boolean logic of the if statement in line with sensible human language (and not being too verbose in itself).

src/Extensions/ElementalAreasExtension.php
ScopeyNZ: > Do you feel strongly about this?

No.

NightJar: Cool, what are your thoughts on my response? :)
Does it make sense, should I add comments? Sounds good, YOLO?

NightJar:
Thanks for the reviews @Firesphere @ScopeyNZ 
Much appreciated :)

PR #734, opened 2019-09-11
Fix Used On tab showing element area

bergice:
Resolves https://github.com/silverstripe/silverstripe-asset-admin/issues/903


emteknetnz:
Pretty sure we don't want this particular PR
- We want to keep showing 'Content Blocks' in the used-on/campaigns https://github.com/silverstripeltd/product-issues/issues/222#issuecomment-663253169
- ElementalArea's are already hidden from campaigns https://github.com/dnadesign/silverstripe-elemental/blob/4.4/src/Models/ElementalArea.php#L63

I've got a new PR raised to hide ElementalArea's from the UsedOnTable https://github.com/dnadesign/silverstripe-elemental/pull/817

PR #735, opened 2019-09-19
Update sortablegridfield to ~1.0

PeideAmonNg:


emteknetnz:
This is targeting a very old branch of elemental and as a dependency change, it would require at least a minor release.  I don't believe there's going to be any more minor releases on the 1.x line so I'll just close this one.

PR #739, opened 2019-10-09
NEW Introduce error boundary to Element for softer crashes

Cheddam:
Most exceptions in the ElementalEditor are caused by a malfunction in an element's form components. Prior to this change, React would unmount the entire editor when an exception like this was thrown, resulting in confusion for users. Now, we capture these exceptions at the Element component level, and render a warning, instructing the user to attempt recovery by saving and reloading.

You can read more about error boundaries [here](https://reactjs.org/docs/error-boundaries.html).


client/src/components/ElementEditor/Element.js
```diff
@@ -21,6 +21,10 @@ import { elementDragSource, isOverTop } from 'lib/dragHelpers';
  * of an element's details when used in the CMS, including ID, Title and Summary.
  */
 class Element extends Component {
+  static getDerivedStateFromError() {
+    return { editingError: true };
```
dnsl48: `editingError` feels like it's caused by users editing the stuff in the element however, error boundaries don't catch errors in event handlers. Should we rename it to something like `childrenRenderingError`?

Cheddam: Tweaked :)

brynwhyman:
I came here looking for a screenshot of the new error - so I'll add one for next time I'm looking!
image

PR #742, opened 2019-10-25
Make sure ignored_classes are ignored

RVXD:
The method requireDefaultRecords does not check on config setting for ignored_classes, nor the extension method 'includeElemental'.

Because all classes are processed each as $ownerClass, subClasses will always also be processed, even when set as ignored.
So ElementalAreas will always be created, even on classes that do not need them.

This can be really problematic on large sites with many pages (ie. shops with 100K products), causing slow or failing dev/build.

This change adds checks to prevent the creation of these unwanted ElementalAreas.


ScopeyNZ:
Nice. Thanks

PR #743, opened 2019-10-29
FIX Resolve issues with drag/drop behaviour on Elements

### LINKED ISSUE #737: Drag 'n drop drop indicator looks broken
### Overview

When dragging a block the line indicating where the block is no longer accurate (regression).

### Versions affected
- `dnadesign/silverstripe-elemental 4.x-dev`
- NOTE: I can't recreate in `dnadesign/silverstripe-elemental 4.2.1`

### Steps to recreate
1. Attempt to drag a block to reorder it
2. Identify that the block will be re-positioned correctly based on the position of the cursor; but the indicator is in the wrong place. +

This makes it really unclear to know where the block will end up.

### Expected result

The blue line should appear above or below the block the cursor (thats dragging a block) is currently over. If the cursor is in the top half of a block, it goes above; if the cursor is in the bottom half of a block, it goes below.

Cheddam:
This PR fixes two drag/drop issues in the ElementalEditor:

- The top-most HoverBarComponent was being rendered during drag operations, resulting in broken drop target behaviour.
- Move operations were triggering when a dragged item was dropped back in its original place. There's a bug in that operation that moves the item to the top of the list in this situation. I opted to disable the operation entirely in this situation to minimise the work being done.

Fixes #737


ScopeyNZ:
LGTM code wise, didn't test is physically.

Cheddam:
I just added an additional fix to address the root cause of the issue with items shifting when they're put back in the same place. That bug won't get hit anyway, but if we happen to back out the no-op fix later, it'll avoid this behaviour breaking again.

dnsl48:
Sorry, there seems to be a bug. (console is empty)

peek-elemental-743

Cheddam:
@dnsl48 My fix for one of the existing movement bugs exposed an _additional_ movement bug that was stopping items from being shifted to the top.

I solved that, but was still able to reproduce the bug you trigger in your screencast when shifting a block from second position to third. In this case, it looks like the react-dnd library is handing back a move operation with no context. I couldn't find a root cause for this behaviour, but I've added logic to fully cancel the move operation when this occurs (otherwise it makes it through to the GraphQL operation and causes an error).

Give the new fixes a go and see if you can reproduce any further issues. Hopefully the move failures are minimal enough to avoid frustration - otherwise I can look at updating the library to see if that provides any fixes.

Cheddam:
I've looked into the Internet Explorer 11 issue, and raised a seperate issue: #750

I tried rolling back to before the change that caused the initial breakage this PR aims to resolve, and this was still a problem - so I think it'd be best to merge this PR to get DnD back up and running for all browsers, and resolve the additional IE issue separately.

dnsl48:
fair enough

PR #763, opened 2019-12-16
Don't redirect to home page on empty CMSEditLink

tractorcow:
CMSEditLink can sometimes be empty. In this case, clicking on this will take the user out of the CMS and to their site's home page (`<a href="" />`). This took a lot of tracking down. =(


michalkleiner:
Does it make sense to keep the `<a>` without the `href` attribute? Should the logic be around the whole link?

NightJar:
I think it does. An anchor isn't necessarily a link, although that is the most common use case.
E.g. removing it (or changing the tag type) could have wider impact with CSS or bound JS for example. Things that are otherwise invisible in this PR.

michalkleiner:
Ok, it makes sense, didn't know it was valid, but TIL https://stackoverflow.com/a/19167175

PR #773, opened 2020-02-06
MINOR: Add elements saved extension hook

unclecheese:


src/Forms/ElementalAreaField.php
```diff
@@ -247,6 +248,9 @@ public function saveInto(DataObjectInterface $dataObject)
 
             $element->updateFromFormData($data);
             $element->write();
+            $elements[] = $element;
         }
+
+        $this->extend('onElementsSaved', $elements);
```
emteknetnz: does this follow convention for extension hook naming? should it be `extend('updateSaveInto')` ?

michalkleiner:
How's the performance for a couple of hundreds of blocks? What's the expected use case, could it be e.g. just the IDs of the blocks?

unclecheese:
@michalkleiner It's only blocks you've written to, so if you've got 100s of inline edits in a single save, I would expect performance problems would be rearing their ugly heads elsewhere, too.

src/Forms/ElementalAreaField.php
unclecheese: Have updated to `onSaveInto` with the `$elements`, `$dataObject` and `$elementData` parameters.

src/Forms/ElementalAreaField.php
```diff
@@ -247,6 +248,9 @@ public function saveInto(DataObjectInterface $dataObject)
 
             $element->updateFromFormData($data);
             $element->write();
+            $elements[] = $element;
         }
+
+        $this->extend('onSaveInto', $elements, $dataObject, $elementData);
```
emteknetnz: just wondering it it makes sense to change the variable name `$elementData`, because there's a plural `$elements`, and a singlular `$elementData`

Would it make sense to rename it to `$elementalAreaFieldValue` to make it obvious what it is?

unclecheese: Maybe, but then the loop above looks weird as you're iterating through `elementalAreaFieldValue` which doesn't seem like it would be an array unless you know the internals.

How about `$elementsChanged` and `$allElementData`?

emteknetnz: Ah yeah course, my bad, "data" is essentially a plural - calling it elementArrays instead of elementData is would be equivalent, though data sounds nicer

Change `$elements` to `$updatedElements` and I'm happy naming wise

unclecheese:
@emteknetnz Integration tests added.

tests/Forms/ElementalAreaFieldTest.php
```diff
@@ -46,4 +52,78 @@ public function testFieldReturnsCompositeFieldOnReadonlyTransformation()
...
+        $element = $this->objFromFixture(TestElement::class, 'element1');
+        $page = $element->getPage();
+        $formID = sprintf(ElementalAreaController::FORM_NAME_TEMPLATE, $element->ID);
+        $fieldName = sprintf(EditFormFactory::FIELD_NAMESPACE_TEMPLATE, $element->ID, 'TestValue');
+
+        // Apply the extension
+        $extensionMock = $this->getMockBuilder(ElementalAreaFieldExtension::class)
+            ->setMethods(['onSaveInto'])
+            ->getMock();
+        $extensionMock->expects($this->once())
```
emteknetnz: could you add a comment about what this black magic is?  It's testing something though I'm not actually sure what?

tests/Extensions/ElementalAreaFieldExtension.php
```diff
@@ -0,0 +1,13 @@
...
+
+
+namespace DNADesign\Elemental\Tests\Extensions;
+
+use SilverStripe\ORM\DataExtension;
+
+class ElementalAreaFieldExtension extends DataExtension
+{
+    public function onSaveInto($updatedElements, $page, $elementData)
+    {
```
emteknetnz: A little confusing having an empty function here?  Might be easier to just change one of the values in $updatedElements and do an assertion the bottom of ElementalAreaFieldTest to test that the onSaveInto extension method was called?

tests/Forms/ElementalAreaFieldTest.php
unclecheese: Test doubles are black magic? Seems like a bit of a stretch given how much we use them. https://github.com/silverstripe/silverstripe-framework/search?q=in%3Afile+-%3EgetMock&unscoped_q=in%3Afile+-%3EgetMock

tests/Extensions/ElementalAreaFieldExtension.php
unclecheese: A mock that spies on the method call is the correct way to test this. The implementation I have tests all three parameters and doesn't rely on side effects. Agree the empty method call is weird, but `->extend()` does a `method_exists()` call before invoking anything, so there has to be something there. https://github.com/silverstripe/silverstripe-framework/blob/4/src/Core/Extensible.php#L465

tests/Forms/ElementalAreaFieldTest.php
ScopeyNZ: Yeah I have to agree with @unclecheese here. This is normal usage of PHPUnit mocks, and the API is pretty declarative.

emteknetnz:
@unclecheese got a failing test on this, would you be able to have a quick look and fix up?  Once green I'm happy to merge after you squash

emteknetnz:
This pull request hasn't had any activity for a while. Are you going to be doing further work on it, or would you prefer to close it now?

emteknetnz:
It seems like there's going to be no further activity on this pull request, so we've closed it for now. Please open a new pull-request if you want to re-approach this work, or for anything else you think could be fixed or improved.

PR #774, opened 2020-02-06
Top page feature (performance enhancement)

mfendeksilverstripe:
# Top page feature (performance enhancement)

Top page feature is a performance enhancement designed especially for projects with deeper nesting of objects. The deeper the nesting the higher performance gain.

This feature should be `Virtual Blocks` compatible.

## Performance issues

Currently, the bottom up data lookups (from block to page) are slow or almost impossible in some cases (when using Fluent and locale us unknown). These are the most common cases of such lookups:

* `CMS fields` - block level CMS fields depend on page data
* `templates` - block level templates logic depends on page data
* `dev tasks` - dev tasks needs to match block to pages (starting from block context)

Deeply nested data structures may not be that uncommon either. Consider this example:

```
Page
  ElementalArea
    RowBlock (represents grid row on frontend)
     ElementalArea
       AccordionBlock (block which can contain other content blocks)
         ElementalArea
           ContentBlock
```

This module supports following bottom up data lookups:

* `ElementalArea::getOwnerPage()` - this method returns most immediate owner object not necessarily the top page, the method is cached in-memory but needs to be called several times on different objects when traversing bottom up. In our example above this method would return `AccordionBlock` when calling from deepest located elemental area

* `BaseElement::getPage()` - this method is just a cached (in-memory) proxy to the `ElementalArea::getOwnerPage()` so it doesn't necessarily return the top page

## Solution

This feature stores the page reference on every block and every elemental area and it provides a simple API to lookup the top page. This feature is duplication safe. `Fluent module` integration support is available as well.

## Other changes

* minimum PHP version bumped up to `7.1`
* minor linting changes

## Related issues

https://github.com/dnadesign/silverstripe-elemental/issues/775

### DIFF
--- a/docs/en/advanced_setup.md
+++ b/docs/en/advanced_setup.md
@@ -275,3 +275,45 @@ FluentState::singleton()->withState(function (FluentState $state) {
 
 This is very important as global state is reverted back after the callback is executed so it's safe to be used.
 Unit tests benefit mostly from this as this makes sure that there are no dependencies between unit tests as the global state is always changed only locally in one test.
+
+## Top page reference feature
+
+In some cases your project setup may have deeply nested blocks, for example:
+
+```
+Page
+  ElementalArea
+    RowBlock (represents grid row on frontend)
+     ElementalArea
+       AccordionBlock (block which can contain other content blocks)
+         ElementalArea
+           ContentBlock
+```
+
+It's quite common to use top page lookups from block context, i.e. a block is querying data from the page that the block belongs to.
+
+Most common cases are:
+
+* `CMS fields` - block level conditional logic depends on page data
+* `templates` - block level render logic depends on page data
+
+This module uses some in-memory caching but this isn't good enough for such deeply nested data structures by default.
+
+In such cases it is recommended to use this feature which stores the top page reference on individual blocks and elemental areas.
+This speeds up data lookup significantly.
+
+If your project makes use of the Fluent module, it is recommended to use the following extensions to replace the ones used by default:
+
+```
+DNADesign\Elemental\Models\BaseElement:
+  extensions:
+    topPageDataExtension: DNADesign\Elemental\TopPage\FluentExtension
+
+DNADesign\Elemental\Models\ElementalArea:
+  extensions:
+    topPageDataExtension: DNADesign\Elemental\TopPage\FluentExtension
+```
+
+This will store the locale of the top page on blocks which simplifies top page lookup in case the locale is unknown at the time of page lookup from block context.
+
+The page reference data on the blocks can also be used for maintenance dev tasks as it's easy to identify which blocks belong to which pages in which locale.
--- a/docs/en/advanced_setup.md
+++ b/docs/en/advanced_setup.md
@@ -92,13 +92,13 @@ class BlockPage extends Page
     private static $field_include = [
         'ElementalAreaID',
     ];
-    
+
     // ...
 }
 ```
 
 This configuration allows us to have different `ElementalArea` for different locales of the page.
-We also need to create a copy of the `ElementalArea` when content is being localised. 
+We also need to create a copy of the `ElementalArea` when content is being localised.
 
 ```php
 class BlockPage extends Page
@@ -106,17 +106,17 @@ class BlockPage extends Page
     public function onBeforeWrite()
     {
         parent::onBeforeWrite();
-    
+
         if (!$this->isDraftedInLocale() && $this->isInDB()) {
             $elementalArea = $this->ElementalArea();
-    
+
             $elementalAreaNew = $elementalArea->duplicate();
             $this->ElementalAreaID = $elementalAreaNew->ID;
         }
-    
+
         return;
     }
-    
+
     // ...
 }
 ```
@@ -159,7 +159,7 @@ class BlockPage extends Page
         ElementalPageExtension::class,
         BlockPageFluentExtension::class,
     ];
-    
+
     // ...
 }
 ```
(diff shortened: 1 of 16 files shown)


Cheddam:
Hey @mfendeksilverstripe, thanks so much for raising this PR 😄 I'll try to get it reviewed soon, but in the meantime:

I've marked this as `change/minor`, since it seems to be opt-in new API surface, rather than a breaking change to existing behaviour. If that's correct, I'd recommend rebasing the PR onto the `4` branch, otherwise I'm keen to get a bit more context on what it breaks and why 👍

mfendeksilverstripe:
@Cheddam Thanks for the feedback, rebased on `4`

docs/en/advanced_setup.md
```diff
@@ -275,3 +275,75 @@ FluentState::singleton()->withState(function (FluentState $state) {
...
+  extensions:
+    topPage: DNADesign\Elemental\TopPage\SiteTreeExtension
+```
+
+If your project setup uses Fluent module it is recommended to use following configuration instead:
+
+```
+DNADesign\Elemental\Models\BaseElement:
+  extensions:
+    topPage: DNADesign\Elemental\TopPage\FluentExtension
```
phptek: Suggestion: Do we need to use the associative array syntax? This will mess with almost _every_ extension YML config I've seen which use indexed arrays for their extension syntax. If developers try to mix the two, the Symfony YML parser will barf. Also note Silverstripe's order of precedence when it comes to applying extensions. AFAIK, if developers attempt to declare >1 extension on a given class, only the latter will take and the former are clobbered.

docs/en/advanced_setup.md
```diff
@@ -275,3 +275,75 @@ FluentState::singleton()->withState(function (FluentState $state) {
...
+
+The page reference data on the blocks can also be used for maintenance dev tasks as it's easy to identify which blocks belong to which pages in which locale.
+
+### Top page reference data during object duplication
+
+Duplicating data object will not duplicate the top page reference data.
+Instead, the newly created data objects will have a new top page data assigned based on the context.
+
+For example, duplicating a page with all of its blocks will create a new page and new blocks.
+All the new blocks will have the new page stored as their top page reference. This works even if the deplication tree contains other pages as tree nodes.
```
phptek: Typo: "duplication tree" or possibly what was meant was "duplicated tree"?

src/TopPage/DataExtension.php
```diff
@@ -0,0 +1,216 @@
...
+     * the relations are not yet written so it's impossible to do a parent lookup at that time
+     */
+    protected function updateTopPage(): void
+    {
+        /** @var SiteTreeExtension $extension */
+        $extension = singleton(SiteTreeExtension::class);
+        $extension->addDuplicatedObject($this->owner);
+    }
+
+    protected function assignTopPage(Page $page): void
```
Cheddam: Needs a docblock

src/TopPage/FluentExtension.php
```diff
@@ -0,0 +1,43 @@
...
+class FluentExtension extends DataExtension
+{
+    /**
+     * @var array
+     */
+    private static $db = [
+        'TopPageLocale' => 'Varchar',
+    ];
+
+    protected function assignTopPage(Page $page): void
```
Cheddam: Needs a docblock

src/TopPage/SiteTreeExtension.php
```diff
@@ -0,0 +1,248 @@
...
+     *
+     * @param Page $original
+     * @param bool $doWrite
+     */
+    public function onAfterDuplicate(Page $original, $doWrite): void
+    {
+        $this->processDuplication($original, (bool) $doWrite);
+    }
+
+    public function getDuplicationKey(): ?string
```
Cheddam: Needs a docblock

src/TopPage/SiteTreeExtension.php
```diff
@@ -0,0 +1,248 @@
...
+
+        if (in_array($key, $this->duplicatedPages)) {
+            // this should never happen as it would indicate a duplication loop
+            return;
+        }
+
+        array_unshift($this->duplicatedPages, $key);
+    }
+
+    protected function processDuplication(Page $original, bool $written): void
```
Cheddam: Needs a docblock

src/TopPage/DataExtension.php
```diff
@@ -0,0 +1,216 @@
...
+
+/**
+ * Class DataExtension
+ *
+ * Top page data cache for improved performance
+ * intended owners of this extension are @see BaseElement and @see ElementalArea
+ * applying this extension to just one of these owners will not hinder top page functionality
+ * but the performance gain will be smaller
+ * it is recommended to apply this extension to BaseElement and for setups with deeper block nesting
+ * it is recommended to cover ElementalArea as well
```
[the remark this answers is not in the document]
mfendeksilverstripe: The downside is that you have to store more data in the DB and it may not be necessary if you have a flat block page (i.e. only one elemental area).

mfendeksilverstripe:
@Cheddam Thank you for your feedback, I pushed up some documentation changes.

I had a look at the `Virtual Blocks` module and I think that this feature will work with it just fine.

I do recommend to look into deprecating the shared blocks approach and recommend the use of `Virtual Blocks`.

I'm not very keen on moving this to a separate module as this is just a performance fix. The performance is painfully slow in a deeply nested blocks setup and I see this as a severe issue of this module which should be fixed in this module.

I will look into updating the feature to make it enabled by default and not break when a project is using shared blocks.

mfendeksilverstripe:
@Cheddam As far as I can see the shared blocks feature is no longer there as Elemental area has many blocks, but block only has one parent. I enabled this performance improvement by default and rebased on `4`.

mfendeksilverstripe:
@Cheddam All tests are now green :)

Cheddam:
@mfendeksilverstripe After discussion with community members on Slack, I've made an adjustment to disable the extensions by default, and reintroduced the documentation describing how to enable them. If you're happy with the change, we can merge and then raise a follow-up PR to `master` in order to enable it there by default.

mfendeksilverstripe:
@Cheddam Thanks I'll have a look at it today and I'll try to fix the unit tests ;-)

mfendeksilverstripe:
@Cheddam Your changes look fine but I'm not sure why the Behat tests are now failing. Note that the tests passed before your change. Please look into it.

Cheddam:
The behat failure was caused by an upstream issue, which has since been resolved, so CI is now green 👍

PR #777, opened 2020-02-18
FIX: getOwnerPage() returning VirtualPage instead of source page (fixes #776)

### LINKED ISSUE #776: Virtual pages break non-inline editing
# How to reproduce

Easiest repro steps are:

1. Set `DNADesign\Elemental\Models\BaseElement.inline_editable: false` in your config
2. Create a virtual page, point it to your home page, Save & Publish
3. Create a new page, add a content block, Save & Publish
4. Edit the virtual page, change it to point to your new page, Save & Publish
5. Edit your new page again, now you can’t edit any content blocks - clicking on one results in an error: `I can't handle sub-URLs on this handler.`

# What’s happening

The problem is (maybe) this query: https://github.com/dnadesign/silverstripe-elemental/blob/4.3.1/src/Models/ElementalArea.php#L211. When there’s a virtual page added which points to a page with content blocks, there are two records with the same `ElementalAreaID` added to the `Page`/`Page_Live` table:

```
ID	ElementalAreaID
2	123
3	123
```

The above query just returns the first row, so if the VirtualPage is added _last_ everything is fine. However, if the VirtualPage is added _first_ and then changed to point to a page that’s added after it, the first record returned will actually be the VirtualPage and not the source page. That results in jumping from a URL like `/admin/pages/edit/show/2` to a URL like `/admin/pages/edit/EditForm/3/field/ElementalArea/item/123/edit`. Notice the ID has changed from `2` (the normal, source page’s ID) to `3` (the VirtualPage’s ID).

# Solution

Haven’t had time to look in detail. I’m not sure why the VirtualPage is storing an `ElementalAreaID` - when you first create one without a source page assigned, it writes a record to the `Page` table with `ElementalAreaID = 0`. That changes as soon as you pick a source page. Not sure if that’s core VirtualPage behaviour, or a bug in that core behaviour, or something in this module...

Changing the query in `getOwnerPage()` to include a `ClassName` filter also seems to work:

```php
$table = DataObject::getSchema()->tableForField($class, $areaID);
$baseTable = DataObject::getSchema()->baseDataTable($class);
$page = DataObject::get_one($class, [
    "\"{$table}\".\"$areaID\" = ?" => $this->ID,
    "\"{$baseTable}\".\"ClassName\" = ?" => $class
]);
```

kinglozzer:
Fixes #776

### DIFF
--- a/src/Models/ElementalArea.php
+++ b/src/Models/ElementalArea.php
@@ -208,7 +208,11 @@ class ElementalArea extends DataObject
                 $areaID = $eaRelationship . 'ID';
 
                 $table = DataObject::getSchema()->tableForField($class, $areaID);
-                $page = DataObject::get_one($class, ["\"{$table}\".\"$areaID\" = ?" => $this->ID]);
+                $baseTable = DataObject::getSchema()->baseDataTable($class);
+                $page = DataObject::get_one($class, [
+                    "\"{$table}\".\"{$areaID}\" = ?" => $this->ID,
+                    "\"{$baseTable}\".\"ClassName\" = ?" => $class
+                ]);
 
                 if ($page) {
                     $this->setOwnerPageCached($page);


phptek:
Awesome work folks - what are the chances of a release-tag? Last one was produced in the mists of 2019..

PR #782, opened 2020-02-27
BUGFIX: Ensure fields exist before replacing them in readonly transform

unclecheese:
While the constructor composes the fields in its own specific way, that same composition cannot be relied upon in subsequent method calls. Use cases include removing the checkbox because Title is required and must display.

This change just puts guards around the replace calls to ensure the fields exist.

### DIFF
--- a/src/Forms/TextCheckboxGroupField.php
+++ b/src/Forms/TextCheckboxGroupField.php
@@ -46,7 +46,7 @@ class TextCheckboxGroupField extends CompositeField
 
             $field->setTemplate(CompositeField::class);
             $field->setTitle('Title');
-            
+
             $titleField = $field->fieldByName('Title');
             if ($titleField) {
                 $field->replaceField('Title', LiteralField::create(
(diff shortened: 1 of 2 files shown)


src/Forms/TextCheckboxGroupField.php
```diff
@@ -46,19 +46,24 @@ public function performReadonlyTransformation()
...
             $field->setTemplate(CompositeField::class);
             $field->setTitle('Title');
-
-            $field->replaceField('Title', LiteralField::create(
-                'Title',
-                $field->fieldByName('Title')->Value()
-            ));
+            
+            $titleField = $field->fieldByName('Title');
+            if ($titleField) {
```
emteknetnz: seems like it would make more sense to change this PR so that we test for the existence of the field within the replaceField function instead?

unclecheese: `replaceField` actually fails gracefully. It's the unguarded of `$titleField->Value()` that blows up.

emteknetnz: ah ok, so what's happening is it's going null->Value() ?

unclecheese: Yup. 💥

emteknetnz:
Could you please add a simple happy-day unit test to TextCheckboxGroupFieldTest.php

unclecheese:
Done

PR #792, opened 2020-04-22
Try using a regular publish rather than copyToStage

maxime-rainville:
This tries to update the publishing logic for blocks to do an actual publish and fire an onBeforePublish/onAfterPublish hook.

Potential fix for https://github.com/dnadesign/silverstripe-elemental/issues/779


maxime-rainville:
Oups ... this is a duplicate of https://github.com/dnadesign/silverstripe-elemental/pull/787

composer.json
```diff
@@ -18,6 +18,7 @@
     "require": {
         "silverstripe/cms": "^4.4@dev",
         "silverstripe/admin": "^1.4@dev",
+        "silverstripe/versioned": "^1.3@dev",
```
maxime-rainville, line 21: I decided to add an explicit requirements for Versioned. Versioned is implicitly installed because of versioned-admin, but since we're directly referencing Versioned APIs, I think we should explicitly state that the module requires Versioned.

maxime-rainville:
Updated the test and remove superfluous parameters. This is ready for review.

maxime-rainville:
FYI I've got that targeted to the next minor. Arguably, it's a bugfix and could/should be shipped in patch release. But I'm not sure how people will react if hooks are suddenly being picked up.

I'm kind of on the fence.

emteknetnz:
Have tested locally, can confirm that extension method is called

```
BaseElementExtension.php
<?php

namespace {

    class BaseElementExtension extends DataExtension
    {

        public function onAfterPublish()
        {
            // this code is run, file is created
            file_put_contents('/var/www/mysite/www/_c.txt', 'onAfterPublish');
        }
    }
}

extensions.yml
DNADesign\Elemental\Models\BaseElement:
  extensions:
    - BaseElementExtension
```

PR #797, opened 2020-05-12
NEW Cache helper methods

dnsl48:
A potential implementation of the cache helper methods, accordingly to the discussion in https://github.com/dnadesign/silverstripe-elemental/issues/795

Usage examples

#### Template partial cache
```html
<% cached $ElementalArea.getCacheKey if $ElementalArea.isCacheable %>
    $ElementalArea
<% end_cached %>
```
#### HTTP cache
```php
if (!$elementalArea->isCacheable()) {
    HTTPCacheControlMiddleware::singleton()
        ->disableCache();
}
```

### DIFF
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -1033,4 +1033,38 @@ JS
 
         return  ($odd) ? 'odd' : 'even';
     }
+
+    /**
+     * Returns true if the current block state can be cached
+     *
+     * This describes the current block instance state cacheability.
+     *
+     * E.g. if the block is currently rendering a half-filled form with
+     * private user data, it's not cacheable. If the block state is going to
+     * render a simple HTML content, perhaps it may be cached.
+     *
+     * @return bool False by default
+     */
+    public function isCacheable()
+    {
+        return false;
+    }
+
+    /**
+     * Returns the block cache key
+     *
+     * The key should be unique for every instance of the block (per BaseElement->ID).
+     *
+     * It does not have to be unique across different blocks
+     * nor even different instances of the same block.
+     *
+     * If the block is not cacheable {@see self::isCacheable}, this method may
+     * return anything (e.g. non-unique values).
+     *
+     * @return string
+     */
+    public function getCacheKey()
+    {
+        return '';
+    }
 }
--- a/src/Models/ElementalArea.php
+++ b/src/Models/ElementalArea.php
@@ -297,4 +297,42 @@ class ElementalArea extends DataObject
 
         return false;
     }
+
+    /**
+     * Return cacheability of the current elemental area instance
+     *
+     * This usually means every block in the current area is cacheable.
+     * If a single block is not cacheable, the whole area becomes not cacheable.
+     *
+     * @return bool
+     */
+    public function isCacheable()
+    {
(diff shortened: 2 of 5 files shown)


src/Models/ElementalArea.php
```diff
@@ -297,4 +297,42 @@ public function canView($member = null)
...
+
+    /**
+     * Returns a unique cache key of the current elemental area instance
+     *
+     * @return string
+     */
+    public function getCacheKey()
+    {
+        $hash = hash_init('sha256');
+        hash_update($hash, $this->ID.$this->LastEdited);
```
mfendeksilverstripe, line 328: This cache key doesn't cover cases like `Versioned` and `Fluent` (where ID is not enough). I recommend using [getUniqueKey](https://github.com/silverstripe/silverstripe-framework/blob/4/src/ORM/DataObject.php#L4203) or similar.

src/Models/BaseElement.php
```diff
@@ -1033,4 +1033,38 @@ public function EvenOdd()
...
+     * nor even different instances of the same block.
+     *
+     * If the block is not cacheable {@see self::isCacheable}, this method may
+     * return anything (e.g. non-unique values).
+     *
+     * @return string
+     */
+    public function getCacheKey()
+    {
+        return '';
```
mfendeksilverstripe, line 1068: This default is quite dangerous. This should either provide a good default key or throw.

src/Models/ElementalArea.php
dnsl48: good catch, thank you!

src/Models/BaseElement.php
dnsl48: Yes, I considered making it random by default and this could be dangerous if we needed it to be unique.
On the other hand, I decided it's better in this way because:
 - it's only bad if `isCacheable` returns `true`
 - if we return random here and users implement `isCacheable` but forget to override this method (dev mistake)
   - it's harder to notice the mistake in `DEV` stage
   - when it hits production, the app cache storage will be overwhelmed randomly, which is really hard to debug and catch

dnsl48: Throwing an exception might be a good alternative though, I didn't think about that one.
That's a good one. I think we should do that.

mfendeksilverstripe: I'm just thinking about how would I use this is another project where we implemented a very similar solution. In that project most blocks work with a very simple unique key, only few need to override, vast majority is cachable, but some are not. Given your feedback my recommendation is:

* provide a good default for `getCacheKey` for example getUniqueKey() which I mentioned above
* add an extension point inside the `getCacheKey` to override this in one place, overriding this on all blocks in a project is not the best option
* provide static configuration for `isCacheable` which defaults to `false` instead of hard coding the value in the code

dnsl48:
Thank you for good feedback @mfendeksilverstripe!
I'll address every point after we vote on this feature in https://github.com/dnadesign/silverstripe-elemental/issues/795
I'm currently preparing a vote about whether we want/need this API in the module at all.

dnsl48:
The feature didn't make it past the vote, so we'll rather be trying to provide better guidelines for users in the documentation.
Thanks for the review anyways! :)

PR #800, opened 2020-06-02
Return null if OwnerClassName doesn't exist

ec8or:
Fixes an issue where elements won't load if they belong to a page type that has been removed, for example in a model admin.

### DIFF
--- a/src/Models/ElementalArea.php
+++ b/src/Models/ElementalArea.php
@@ -198,6 +198,9 @@ class ElementalArea extends DataObject
 
         if ($this->OwnerClassName) {
             $class = $this->OwnerClassName;
+            if (!ClassInfo::exists($class)) {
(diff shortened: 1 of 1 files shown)


src/Models/ElementalArea.php
```diff
@@ -198,6 +198,9 @@ public function getOwnerPage()
 
         if ($this->OwnerClassName) {
             $class = $this->OwnerClassName;
+            if (!ClassInfo::exists($class)) {
+                return null;
+            }
```
ScopeyNZ, line 203:
```suggestion
        if ($this->OwnerClassName && ClassInfo::exists($this->OwnerClassName)) {
            $class = $this->OwnerClassName;
```

This way it still falls back to the alternate logic below.

PR #806, opened 2020-06-15
Update for 4.4

emteknetnz:


.travis.yml
```diff
@@ -26,22 +26,22 @@ env:
 matrix:
   include:
     - php: 7.1
-      env: DB=MYSQL RECIPE_VERSION=4.4.x-dev PHPCS_TEST=1 PHPUNIT_TEST=1
+      env: DB=MYSQL RECIPE_VERSION=4.5.x-dev PHPCS_TEST=1 PHPUNIT_TEST=1
```
maxime-rainville, line 29: Those are meant to target the minimum supported version of various module. You shouldn't bump them.

composer.json
```diff
@@ -17,11 +17,11 @@
     "prefer-stable": true,
     "require": {
         "php": "^7.1",
-        "silverstripe/cms": "^4.4@dev",
-        "silverstripe/admin": "^1.4@dev",
-        "silverstripe/versioned": "^1.3@dev",
-        "silverstripe/versioned-admin": "^1.2@dev",
-        "symbiote/silverstripe-gridfieldextensions": "^3.1",
+        "silverstripe/cms": "^4.6@dev",
```
maxime-rainville, line 20: You should have left those alone. They target the minimum supported version,

composer.json
```diff
@@ -36,7 +36,7 @@
     },
     "extra": {
         "branch-alias": {
-            "4.x-dev": "4.2.x-dev"
+            "4.x-dev": "4.4.x-dev"
```
maxime-rainville, line 39: Not sure why we still have branch aliases here. I thought we got rid of all of those things.

.travis.yml
```diff
@@ -16,7 +16,7 @@ addons:
 
 env:
   global:
-    - COMPOSER_ROOT_VERSION=4.3.x-dev
+    - COMPOSER_ROOT_VERSION=4.4.x-dev
```
maxime-rainville, line 19: That's really the only thing that needed updating here.

composer.json
emteknetnz: seems like we got rid of the a.b.x-dev branch-aliases, on a bunch of modules we still have the "dev-master": "5.x-dev", idea being that if (for whatever reason) you try to install the next major version you get the master branch.  I'm not entirely sure why we even bother with this, though I've gotten feedback from other core committers that they're sometimes useful

seems like this should be changed to "dev-master": "5.x-dev" though

composer.json
```diff
@@ -17,11 +17,11 @@
...
-        "silverstripe/cms": "^4.4@dev",
-        "silverstripe/admin": "^1.4@dev",
-        "silverstripe/versioned": "^1.3@dev",
-        "silverstripe/versioned-admin": "^1.2@dev",
-        "symbiote/silverstripe-gridfieldextensions": "^3.1",
+        "silverstripe/cms": "^4.6@dev",
+        "silverstripe/admin": "^1.6@dev",
+        "silverstripe/versioned": "^1.6@dev",
+        "silverstripe/versioned-admin": "^1.6@dev",
+        "symbiote/silverstripe-gridfieldextensions": "^3.2",
```
maxime-rainville, line 24:
```suggestion
        "silverstripe/cms": "^4.4@dev",
        "silverstripe/admin": "^1.4@dev",
        "silverstripe/versioned": "^1.3@dev",
        "silverstripe/versioned-admin": "^1.2@dev",
        "symbiote/silverstripe-gridfieldextensions": "^3.1",
```

maxime-rainville: Build was failing because some matrix couldn't resolve to an installable set of packages.

.travis.yml
```diff
@@ -26,22 +26,22 @@ env:
...
+      env: DB=MYSQL RECIPE_VERSION=4.5.x-dev BEHAT_TEST=1
     - php: 7.2
       env: DB=MYSQL RECIPE_VERSION=4.x-dev PHPUNIT_TEST=1
     - php: 7.3
-      env: DB=MYSQL RECIPE_VERSION=4.5.x-dev PHPCS_TEST=1
+      env: DB=MYSQL RECIPE_VERSION=4.6.x-dev NPM_TEST=1
     - php: 7.3
+      env: DB=MYSQL RECIPE_VERSION=4.x-dev PHPUNIT_TEST=1
+    - php: 7.4
+      env: DB=MYSQL RECIPE_VERSION=4.6.x-dev PHPCS_TEST=1
```
ScopeyNZ, line 43: You don't need two PHPCS tests.

ScopeyNZ: Correction, there's three now, there were two originally.

.travis.yml
maxime-rainville:
```suggestion
      env: DB=MYSQL RECIPE_VERSION=4.4.x-dev PHPUNIT_TEST=1
```

.travis.yml
maxime-rainville:
```suggestion
      env: DB=MYSQL RECIPE_VERSION=4.6.x-dev
```

maxime-rainville:
Looks like there's a legit failure on the PostgreSQL build. I'll merge as-is so we can do the CWP RC. But I'll make a note to resolve the failure after the RC is out.

PR #807, opened 2020-06-21
Update branch-alias to next major

emteknetnz:
Currently cannot install 4.x-dev properly due to misconfigured branch-alias

Failing travis test is an unrelated existing issue with postgres


emteknetnz:
Should probably just get rid of the branch-alias

PR #813, opened 2020-07-10
FIX add classes to read-only transformation

emteknetnz:
Fixes https://github.com/dnadesign/silverstripe-elemental/issues/716

View in archive admin now has _italic_ for display status

image

Still looks correct in page history view

image


emteknetnz:
If we use a `ReadonlyField` class in `TextCheckboxGroupField::performReadonlyTransformation()` then we then it'll look like this

image

However we cannot style the contents on the readonly field, because any html tags will get escaped for security reasons, it looks like this if we try

image

Most annoyingly, whatever we do within this function is also reflected in the Page History view, it now looks like this

image

I haven't dug too far into what happens on the page history side, though it seems like there's some javascript that modifies whatever is outputted by PHP

Unfortunately, because of this coupling, I'm not sure quite how to resolve this one without a significant amount of effort spent refactoring things. 

So it seems that we are very constrained with the options we available to make the styling look better

It does seem like LiteralField is useful here because it allows us to use whatever HTML tags we like, because they are not escaped

clarkepaul:
I think we need to just maintain the structure used within Block history for now which looks like this
image

At some stage (not now) all readonly fields need to be updated to use the proper field type for accessibility reasons (issue exists for this already). The "Displayed" text would not be in the field but outside of it so style wouldn't be an issue, the HTML of block titles would already allow this to be done easily with CSS because it separate from the field.

clarkepaul:
When we do start to use a proper readonly field we'd need to update the structure of the template used so that there is still an outer div, the field, displayed title, description etc.

emteknetnz:
@clarkepaul I'm not sure if you're asking me to make any further changes to this PR or not

> I think we need to just maintain the structure used within Block history for now which looks like this

As I previously mentioned, due the coupling between the archive history and the page/block history, this isn't possible right now without major refactoring, which is outside of the scope of this particular issue

The best I can do at this stage is to italicize the words 'Displayed/Not Displayed', which is at least an improvement for how things are currently

image

Would you like this work merged in as it is now, and we can look to do the refactoring in the future?

clarkepaul:
Sorry didn't think it would be difficult to have the same structure (couple extra divs with classes). The italics is slightly better but not really enough to distinguish it as not being part of the title unfortunately. 

Okay see if this is doable instead... can we just hide the "displayed/not displayed" text then. At this stage it would probably be better not showing it at all. Swap out the .show-title class for .d-none.

emteknetnz:
@clarkepaul 
I tried changing the class to `.d-none`, though unfortunately due to the coupling it also vanishes from the page/block history page

What I've done instead is add the following css:

```
// hide .show-title in the Archives ModelAdmin only
.cms-edit-form[action*='admin/archive/'] .show-title {
  display: none;
}

```

It's not the prettiest but it gets the job done

Are you happy to progress getting this peer reviewed and merged now?

client/src/components/TextCheckboxGroupField/TextCheckboxGroupField.scss
```diff
@@ -19,3 +19,8 @@
     }
   }
 }
+
+// hide .show-title in the Archives ModelAdmin only
+.cms-edit-form[action*='admin/archive/'] .show-title {
+  display: none;
```
ScopeyNZ, line 25: ...? Can you not just change the readonly transformation to remove this? This feels really hacky.

ScopeyNZ:
Here's the PR that updated the React rendered readonly version of this input: https://github.com/dnadesign/silverstripe-elemental/commit/e41aa6e5abdfc56157e95de9a35f5d63299ed1a7#diff-a0334774c6b450a2a2af5639d36dea18

Isn't the archive context rendered by React? Is it just missing the SCSS that is provided in that commit?

emteknetnz:
@ScopeyNZ I don't believe the archive model admin uses react

The edit form url is `/archive/DNADesign-Elemental-Models-BaseElement/EditForm/field/Blocks/item/3/view` and the it renders the edit form as as `<form id="Form_ItemEditForm"`

As mentioned previously, there seems to be coupling, meaning that whatever I change in the `performReadonlyTransformation()` function will affect both the acrhive editform as well as the react based page/block history view

I realise the solution I've provided is hacky, though it works.  I don't know of another solution that we can do without investing a significant amount of time.  If you know of another way this can be solved then please share.  Otherwise I suggest we just merge as is.

ScopeyNZ:
I personally feel like the hack is not worth the UX improvement, so I'd prefer to just close until we can provide a better solution.

emteknetnz:
All good, chatted with Garion, he agrees that it's a bad solution.  Will close now.

PR #816, opened 2020-07-22
Fix for when the "OwnerClassName" needs updating in the EA

oilee80:
Otherwise when you change `PageType` it can cause issues with retrieving
the Elements for the page

This is an alternative fix for #671 based on feedback comments

### DIFF
--- a/src/Extensions/ElementalAreasExtension.php
+++ b/src/Extensions/ElementalAreasExtension.php
@@ -227,6 +227,17 @@ class ElementalAreasExtension extends DataExtension
 
         $this->ensureElementalAreasExist($elementalAreaRelations);
 
+        $ownerClassName = get_class($this->owner);
+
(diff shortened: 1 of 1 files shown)


src/Extensions/ElementalAreasExtension.php
```diff
@@ -227,6 +227,17 @@ public function onBeforeWrite()
 
         $this->ensureElementalAreasExist($elementalAreaRelations);
 
+        $ownerClassName = get_class($this->owner);
+
+        // Update the OwnerClassName on EA if the class has changed
+        foreach($elementalAreaRelations As $eaRelation) {
```
Cheddam, line 233:
```suggestion
        foreach ($elementalAreaRelations as $eaRelation) {
```

lbradley-mm: Thanks, updated

PR #817, opened 2020-07-29
FIX Exclude ElementalAreas from file used-on-table

emteknetnz:
Tweak for ~https://github.com/silverstripeltd/product-issues/issues/222~ https://github.com/silverstripe/silverstripe-asset-admin/issues/903

Prevent ElementalArea's from showing on the asset-admin file used-on-table

**Before**

image

**After**

image

### DIFF
--- a/_config/config.yml
+++ b/_config/config.yml
@@ -10,6 +10,10 @@ SilverStripe\Admin\LeftAndMain:
   extensions:
     - DNADesign\Elemental\Extensions\ElementalLeftAndMainExtension
 
+SilverStripe\Admin\Forms\UsedOnTable:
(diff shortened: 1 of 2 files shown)


michalkleiner:
What if there were multiple elemental areas in use and it would be good to know which one the item was used through?

bergice (approved):

clarkepaul:
@michalkleiner That could possibly fit into the breadcrumb. I didn't consider that with the design but we want to keep the path as minimal as possible and I think `page name` and `block name` are more important. I'm suggesting keeping the same design pattern but maybe elemental area could be a potential addition at a later date?

PR #819, opened 2020-08-04
Remove redundant "(displayed if checked)" text from Title field

purplespider:
The visible checkbox with the text "Displayed" next to it portrays the same message, so I'd argue adding this wording to the title doesn't provide any value. In fact, I feel it adds clutter to the interface, especially as it maintains the same prominence as the field label, due to it being set in the field title, rather than as help/description text.

**Before:**

image

**After:**

image

### DIFF
--- a/lang/en.yml
+++ b/lang/en.yml
@@ -5,7 +5,7 @@ en:
     DISPLAYED: Displayed
     NOT_DISPLAYED: 'Not displayed'
     ShowTitleLabel: Displayed
-    TitleLabel: 'Title (displayed if checked)'
+    TitleLabel: 'Title'
   DNADesign\Elemental\Models\BaseElement:
     BlockType: Block
     CUSTOM_STYLES: 'Select a style..'
--- a/lang/en.yml
+++ b/lang/en.yml
@@ -5,7 +5,7 @@ en:
     DISPLAYED: Displayed
     NOT_DISPLAYED: 'Not displayed'
     ShowTitleLabel: Displayed
-    TitleLabel: 'Title'
+    Title: 'Title'
   DNADesign\Elemental\Models\BaseElement:
     BlockType: Block
     CUSTOM_STYLES: 'Select a style..'
--- a/src/Forms/TextCheckboxGroupField.php
+++ b/src/Forms/TextCheckboxGroupField.php
@@ -19,7 +19,7 @@ class TextCheckboxGroupField extends CompositeField
     public function __construct($title = null)
     {
         if (!$title) {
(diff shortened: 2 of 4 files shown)


ScopeyNZ:
cc @silverstripeux

This LGTM

lang/en.yml
```diff
@@ -5,7 +5,7 @@ en:
     DISPLAYED: Displayed
     NOT_DISPLAYED: 'Not displayed'
     ShowTitleLabel: Displayed
-    TitleLabel: 'Title (displayed if checked)'
+    TitleLabel: 'Title'
```
robbieaverill, line 8: Non-English languages will still have more context than this does. When we change English strings we usually must use a new translation key so it doesn't pollute non-English language packs. Can you please rename this? Something like `Title: 'Title'` would do

clarkepaul:
Heads up that this was an MVP and not where it was supposed to have stayed for as long. The intention was to allow users to adjust the heading size so that it can be made semantic to the other headers being used (not all content blocks use TinyMCE so they don't have the flexibility to use hierarchy).

image

image

[As per designs here](https://projects.invisionapp.com/dsm/silver-stripe/silver-stripe/asset/components/5c9c207ede91234535501a35) 

Beyond this, the next part was to move the title field to the very top when the block is open, reducing the overall height of the blocks. I wasn't going to push that part until we had the heading dropdown incorporated (small steps).

clarkepaul:
I'm thinking we used that additional text when we didn't have a label on the checkbox itself. I think its redundant now too.

emteknetnz:
@purplespider This pull request hasn't had any activity for a while. Are you going to be doing further work on it to implement the feedback from robbieaverill, or would you prefer to close it now?

purplespider:
@emteknetnz Yep, I'll update the PR to use a new translation key.

src/Forms/TextCheckboxGroupField.php
```diff
@@ -19,7 +19,7 @@ class TextCheckboxGroupField extends CompositeField
     public function __construct($title = null)
     {
         if (!$title) {
-            $title = _t(__CLASS__ . '.TitleLabel', 'Title (displayed if checked)');
+            $title = _t(__CLASS__ . '.TitleLabel', 'Title');
```
robbieaverill, line 22:
```suggestion
            $title = _t(__CLASS__ . '.Title', 'Title');
```

robbieaverill (changes requested):

robbieaverill (approved):

PR #820, opened 2020-08-04
Remove redundant "Untitled" text from blocks with no title

purplespider:
Instead of "Untitled Content block", it now displays "Content block" in the block summary. It's still clear that a Title hasn't been set as it already displays the summary title in a different style for untitled blocks.

The extra "Untitled" text increased the cognitive load when visually scanning over a mixture of titled and untitled blocks as the useful part of labels (block type) didn't line up with titled blocks.

Block titles are very much optional, and this change makes this clearer.

**Before:**

image

**After:**

image


ScopeyNZ:
cc @silverstripeux

We did this on purpose. The different styling for the missing title probably can't be considered accessible.

clarkepaul:
Some of the thinking that wen into that decision...

1. We want to encourage people to us different block headings even if they aren't being used in the frontend. With "Content block" being a super popular block you end up with 10 blocks in a row without and differentiator. People are more likely to keep the heading as "Content Block" than they would "Untitled content block".
2. The colour difference isn't super obvious as we want to keep it accessible, so we would be relying mainly on the italicised look.
3. The block heading gets pulled into lots of other areas which don't have either italicised or a lighter appearance. For this reason we thought it was best to keep the "Untitled" part so that people for example know it wasn't intentionally called "Banner block". e.g. in Campaign, Used on tab in files area, Block admins...

If the styling was more consistent throughout the CMS (point 3.) then I think we wouldn't have as much of an issue bringing in your recommendations as they are shorter, more concise. 

Side note: I can't recall whether they were to be structured like "Untitled content block" or "Untitled Content block". Looking at it now I prefer the former.

michalkleiner:
Our (a bit unrelated) 2 cents — we repurpose the heading field to be the CMS name of the block, not the visible heading that gets output on the front-end, as we need more flexibility around a block title/heading. We also conditionally (based on project requirements) remove the "Display on page" flag when the heading is mandatory, decoupling the composed heading field back to two normal fields.

clarkepaul:
@michalkleiner do you think if you had the following functionality available for clients you would remove the "display" feature? Designs https://projects.invisionapp.com/dsm/silver-stripe/silver-stripe/asset/components/5c9c207ede91234535501a35

michalkleiner:
@clarkepaul it would certainly help to have a wider set of options than just a visibility toggle.

Assuming everything would be configurable per block — if the options in the dropdown could be controlled, the "Do not display" option optional (allowing to make the title mandatory) — that would help us. Also being able to specify the default value for the dropdown.

We'd still add our CMS name field, which we see as the 'title' that should be used in all lists of blocks, not the title itself, though the title could still act as a fallback.

emteknetnz:
@clarkepaul is this PR something we want to pursue further or should we now close this one?

clarkepaul:
I think we can close. One day we need to do a more holistic view of all untitled items and how they are presented in different views. At the moment things are too inconsistent CMS wide to be making one0-off changes here alone.

PR #822, opened 2020-08-06
NEW Track broken links and files within ElementContent

emteknetnz:
Fix for https://github.com/dnadesign/silverstripe-elemental/issues/689

Also relates to https://github.com/silverstripeltd/product-issues/issues/224 (repository is not publically viewable)

The goal of this issue is to get the [BrokenLinksReport](https://github.com/silverstripe/silverstripe-cms/blob/4/code/Reports/BrokenLinksReport.php) to also read content from Elemental Blocks

SiteTreeLinkTracking and FileLinkTracking are on every DataObject by default, so there's not a whole lot we need to do to enable this functionality on ElementContent

https://github.com/silverstripe/silverstripe-cms/blob/4/_config/linktracking.yml

Updating the tracking only on SiteTree->write() and not ElementContent->write() to prevent unintended consequences, for instance, when I save the Element inline in the CMS I do not expect the Page to save too

### DIFF
--- a/src/Extensions/ElementalAreasExtension.php
+++ b/src/Extensions/ElementalAreasExtension.php
@@ -229,12 +229,22 @@ class ElementalAreasExtension extends DataExtension
 
         $ownerClassName = get_class($this->owner);
 
-        // Update the OwnerClassName on EA if the class has changed
         foreach ($elementalAreaRelations as $eaRelation) {
+            // Update the OwnerClassName on EA if the class has changed
+            /** @var ElementalArea $ea */
             $ea = $this->owner->$eaRelation();
             if ($ea->OwnerClassName !== $ownerClassName) {
                 $ea->OwnerClassName = $ownerClassName;
                 $ea->write();
+            }
+
+            // Owner page will have set its own HasBrokenLink / HasBrokenFile in SiteTreeLinkTracking / FileLinkTracking
+            // on any $page->Content BEFORE we get to this point
+            if (count(array_filter($ea->Elements()->column('HasBrokenLink')))) {
+                $this->owner->HasBrokenLink = true;
+            }
+            if (count(array_filter($ea->Elements()->column('HasBrokenFile')))) {
+                $this->owner->HasBrokenFile = true;
             }
         }
 
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -70,7 +70,9 @@ class BaseElement extends DataObject
         'ShowTitle' => 'Boolean',
         'Sort' => 'Int',
         'ExtraClass' => 'Varchar(255)',
-        'Style' => 'Varchar(255)'
+        'Style' => 'Varchar(255)',
+        'HasBrokenLink' => 'Boolean',
+        'HasBrokenFile' => 'Boolean',
     ];
 
     private static $has_one = [
@@ -291,6 +293,9 @@ class BaseElement extends DataObject
 
             // Remove link and file tracking tabs
             $fields->removeByName(['LinkTracking', 'FileTracking']);
+
+            // Remove SiteTreeLinkTracking and FileLinkTracking fields
+            $fields->removeByName(['HasBrokenLink', 'HasBrokenFile']);
 
             $fields->addFieldToTab(
                 'Root.Settings',
--- a/src/Models/ElementContent.php
+++ b/src/Models/ElementContent.php
@@ -12,7 +12,7 @@ class ElementContent extends BaseElement
     private static $icon = 'font-icon-block-content';
 
     private static $db = [
-        'HTML' => 'HTMLText'
+        'HTML' => 'HTMLText',
     ];
 
     private static $table_name = 'ElementContent';
--- a/tests/Extensions/ElementalAreasExtensionTest.php
+++ b/tests/Extensions/ElementalAreasExtensionTest.php
@@ -3,9 +3,11 @@
 namespace DNADesign\Elemental\Tests\Extensions;
 
 use DNADesign\Elemental\Extensions\ElementalAreasExtension;
+use DNADesign\Elemental\Extensions\ElementalPageExtension;
 use DNADesign\Elemental\Models\ElementContent;
 use DNADesign\Elemental\Tests\Src\TestElement;
 use DNADesign\Elemental\Tests\Src\TestUnusedElement;
+use SilverStripe\Assets\File;
 use SilverStripe\Core\Config\Config;
 use SilverStripe\CMS\Model\SiteTree;
 use SilverStripe\Dev\SapphireTest;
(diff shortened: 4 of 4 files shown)


michalkleiner:
Should this rather be on the `BaseElement` class? If folks have custom elements with WYSIWYG fields, they'd need to replicate adding the fields to their classes etc.

emteknetnz:
@michalkleiner yes good point.  I've updated.

src/Extensions/ElementalAreasExtension.php
```diff
@@ -229,13 +229,26 @@ public function onBeforeWrite()
...
+            // on any $page->Content BEFORE we get to this point
+            /** @var BaseElement $element */
+            foreach ($ea->Elements() as $element) {
+                if ($element->hasField('HasBrokenLink') && $element->HasBrokenLink) {
+                    $this->owner->HasBrokenLink = true;
+                }
+                if ($element->hasField('HasBrokenFile') && $element->HasBrokenFile) {
+                    $this->owner->HasBrokenFile = true;
+                }
+            }
```
Cheddam, line 251: Did some rough profiling on this with Xdebug enabled, with 100 Elements on a Page it didn't seem make too big a difference in response time (less than .5s), but could scale poorly in more complex situations (e.g. when Layout Blocks are used.) To minimise the overhead you could remove the loop and use `in_array()` with `column()` instead:

```php
            if (in_array(1, $ea->Elements()->column('HasBrokenLink'))) {
                $this->owner->HasBrokenLink = true;
            }

            if (in_array(1, $ea->Elements()->column('HasBrokenFile'))) {
                $this->owner->HasBrokenFile = true;
            }
```

This seemed to work out to about a 75% reduction in processing time in my profiling. (`Elements()` caches after the first fetch.)

Cheddam (commented):
Happy with the overall approach, but have suggested an optimisation.

src/Extensions/ElementalAreasExtension.php
Cheddam: Actually, I just realised there's no provision here for removing the HasBrokenLink state from the Page, so the following might make more sense (though it implies that Elements are the single source of truth for these properties):

```php
            $this->owner->HasBrokenLink = in_array(1, $ea->Elements()->column('HasBrokenLink'));
            $this->owner->HasBrokenFile = in_array(1, $ea->Elements()->column('HasBrokenFile'));
```

emteknetnz: Has to be the first suggestion, otherwise if HasBrokenLink/File was `true` before this extension, such as a broken like on $page->Content, the extension will incorrectly override that.

We don't need to worry about setting it to `false` because SiteTree will default it to `false` [here](https://github.com/silverstripe/silverstripe-cms/blob/4/code/Model/SiteTreeLinkTracking.php#L118) before looking for anything that's broken then setting it to `true`

Cheddam:
The following behaviours that differ from standard Page-level broken link tracking remain:

- If a link breaks in a block, the Page it belongs to will not appear in the Broken Links report until the page is saved.
- If a block is individually saved / published after removing a broken link from it, this will not result in removal of the Page from the Broken Links report - the Page itself must be saved to trigger this.

Effectively, we're depending on authors to regularly hit Save on affected pages to update their state, even when the Save button is 'clean'. This is an improvement on "it doesn't work in any situation", but we'll definitely need to follow this up with additional remedial work if we don't address it in this PR.

michalkleiner:
@Cheddam so without explicitly saving the page, I can get false positives or false negatives in the report without knowing why, is that correct? If it is, that should be addressed here if at all possible.

emteknetnz:
I had an honest go at solving this, though I'm not sure this is something that is easily solved unfortunately

I made an attempt at it [here](https://github.com/dnadesign/silverstripe-elemental/pull/861/commits/7c034a4ac8334ab98a0fa07f245c63b4eca642e3) 

This will update the $page->HasBrokenLink (and HasBrokenFile) on $element->write(), however, it only works properly in the instance where the $page->HasBrokenLink = false and the $element->HasBrokenLink = true

I stopped working on this approach when I realized that it can't work when $page->HasBrokenLink = true and $element->HasBrokenLink = false AND the user just fixed a broken link in $element.  At this point we cannot confidently update $page->HasBrokenLink = false because it may be broken there because of a broken link in:
- $page->Content, or
- Something that happened in a different extension

The only alternative solution I can think of here would be a queued job type of thing, though I really don't want to do that because that will introduce a lag between a user doing something (fixing a link or breaking a link) and seeing the report update

I think we may just have to live with broken links report only update when the user click the page save / publish button.

bergice (changes requested):
This doesn't seem to work after running the `CheckExternalLinksTask` task.

After reaching `vendor/silverstripe/externallinks/src/Tasks/CheckExternalLinksTask.php:211` I can see that `$page->HasBrokenLink == true` but `$pageTrack->BrokenLinks()->count() == 0` which is what it uses to report on the number of broken links.

src/Extensions/ElementalAreasExtension.php
```diff
@@ -229,13 +229,23 @@ public function onBeforeWrite()
...
+            /** @var ElementalArea $ea */
             $ea = $this->owner->$eaRelation();
             if ($ea->OwnerClassName !== $ownerClassName) {
                 $ea->OwnerClassName = $ownerClassName;
                 $ea->write();
             }
+
+            // Owner page will have set its own HasBrokenLink / HasBrokenFile in SiteTreeLinkTracking / FileLinkTracking
+            // on any $page->Content BEFORE we get to this point
+            if (in_array(1, $ea->Elements()->column('HasBrokenLink'))) {
```
michalkleiner, line 243: May I suggest using `if (array_filter($ea->Elements()->column('HasBrokenLink'))` to support any truthy value rather than check for integer value `1` in the list?

emteknetnz: I've updated this to check for truthy values in arrays

emteknetnz:
> After reaching `vendor/silverstripe/externallinks/src/Tasks/CheckExternalLinksTask.php:211` I can see that `$page->HasBrokenLink == true` but `$pageTrack->BrokenLinks()->count() == 0` which is what it uses to report on the number of broken links.

@bergice I was unable to replicate this.  That said, seem like this would be an issue with the silverstripe/silverstripe-externallinks module rather that this pull-request on elemental.  As long as `$page->HasBrokenLink = true` then this pull-request is doing its part of the job

emteknetnz:
I don't think this approach will ever fully work in practice.  There's issues with SiteTree:onBeforeWrite ... $fieldsIgnoredByVersioning containing 'HasBrokenLink' and 'HasBrokenFile' - what this means is the a SiteTree_Versions record isn't created even when setting $this->owner->HasBrokenLink from the elemental extension.  The internal broken link report reads off the SiteTree_Live table, which will grab the latest SiteTree_Versions record. You need to something like change the page title and save to get this broken links report to work.

We could potentially work around this, though it's honestly hackish and I'd rather just not

PR #827, opened 2020-08-25
ENH Link BaseElements to parent Page for file used on table

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1098

Blocked by https://github.com/silverstripe/silverstripe-admin/pull/1139

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,23 +26,15 @@ env:
 matrix:
   include:
     - php: 7.1
-      env: DB=MYSQL RECIPE_VERSION=4.4.x-dev PHPUNIT_TEST=1
+      env: DB=PGSQL RECIPE_VERSION=4.7.x-dev PHPCS_TEST=1 PHPUNIT_TEST=1
     - php: 7.1
-      env: DB=PGSQL RECIPE_VERSION=4.6.x-dev PHPUNIT_TEST=1
-    - php: 7.1
-      env: DB=MYSQL RECIPE_VERSION=4.5.x-dev PHPUNIT_COVERAGE_TEST=1
+      env: DB=MYSQL RECIPE_VERSION=4.7.x-dev PHPUNIT_COVERAGE_TEST=1 PDO=1
     - php: 7.2
-      env: DB=MYSQL RECIPE_VERSION=4.5.x-dev BEHAT_TEST=1
-    - php: 7.2
-      env: DB=MYSQL RECIPE_VERSION=4.x-dev PHPUNIT_TEST=1
+      env: DB=MYSQL RECIPE_VERSION=4.7.x-dev BEHAT_TEST=1
     - php: 7.3
-      env: DB=MYSQL RECIPE_VERSION=4.6.x-dev NPM_TEST=1
-    - php: 7.3
+      env: DB=MYSQL RECIPE_VERSION=4.7.x-dev NPM_TEST=1
+    - php: 7.4
       env: DB=MYSQL RECIPE_VERSION=4.x-dev PHPUNIT_TEST=1
-    - php: 7.4
-      env: DB=MYSQL RECIPE_VERSION=4.6.x-dev
-    - php: 7.4
-      env: DB=MYSQL RECIPE_VERSION=4.x-dev PHPCS_TEST=1
 
 before_script:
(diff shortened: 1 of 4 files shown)


Cheddam:
@emteknetnz Might need to tweak the build matrix to get this green, as it's trying to reference a dev branch that has been removed.

src/Extensions/ElementalAreaUsedOnTableExtension.php
```diff
@@ -3,24 +3,80 @@
...
+    public function updateUsageDataObject(DataObject $dataObject)
+    {
+        if (!($dataObject instanceof BaseElement)) {
+            return;
+        }
+        try {
+            if (!$dataObject->getPage()) {
+                $dataObject = null;
+            }
+        } catch (ValidationException $e) {
```
Cheddam, line 56: I _think_ this error boundary is unnecessary - `BaseElement::getPage()` shouldn't throw a `ValidationException`, despite its docblock stating otherwise.

src/Extensions/ElementalAreaUsedOnTableExtension.php
```diff
@@ -3,24 +3,80 @@
...
+            return;
+        }
+        try {
+            // BaseElement::getPage() caches results so there's no performance decrease from
+            // also calling it in updateUsageExcludeDataObject()
+            /** @var SiteTree $page */
+            if ($page = $dataObject->getPage()) {
+                $ancestorDataObjects[] = $page;
+            }
+        } catch (ValidationException $e) {
```
Cheddam, line 79: As above.

.travis.yml
```diff
@@ -26,23 +26,15 @@ env:
 matrix:
   include:
     - php: 7.1
-      env: DB=MYSQL RECIPE_VERSION=4.4.x-dev PHPUNIT_TEST=1
+      env: DB=PGSQL RECIPE_VERSION=4.5.x-dev PHPCS_TEST=1 PHPUNIT_TEST=1 PDO=1
```
Cheddam, line 29: Should we bump the minimum core version requirement if we're going to stop testing `4.4.x`?

Also - will setting `PDO=1` on a `PGSQL` still result in it connecting via PostgreSQL? It doesn't seem to be referenced anywhere in the scripts, so not sure if it actually has any effect?

src/Extensions/ElementalAreaUsedOnTableExtension.php
```diff
@@ -3,24 +3,80 @@
...
+     * Legacy function kept for semver, replaced with updateUsageExcludedClasses above
+     *
      * @return void
-     * @var ArrayList $dataObjects
+     * @var ArrayList $usage
      * @var DataObject $record
      * @see UsedOnTable::updateUsage
+     * @deprecated 4.5.0 Use self::updateUsageExcludedClasses instead
+     */
+    public function updateUsage(ArrayList &$usage, DataObject &$record)
```
Cheddam, line 36: Is this truly semver? What I mean is, if someone had extended this class and implemented this method, would it still do what they expected after updating to CMS 4.7 and Elemental 4.5?

Cheddam (commented):
Pretty happy with this as-is, just a couple of questions.

src/Extensions/ElementalAreaUsedOnTableExtension.php
Cheddam: Also - won't dropping the contents of this method change the behaviour of Used On entries for Elemental when run against CMS 4.6 and earlier?

src/Extensions/ElementalAreaUsedOnTableExtension.php
emteknetnz: It might be, though that doesn't stop phpstorm from complaining

src/Extensions/ElementalAreaUsedOnTableExtension.php
emteknetnz: Ditto

src/Extensions/ElementalAreaUsedOnTableExtension.php
Cheddam: Resolved by raising the minimum core requirement to 4.7.

Cheddam (approved):

PR #830, opened 2020-08-31
NEW: Better type default.

mfendeksilverstripe:
# Better type default for BaseElement

* better default for the `Type` on `BaseElement`
* simplified configuration (configured by `singular_name`)
* localisation is retained

## Related issues

https://github.com/dnadesign/silverstripe-elemental/issues/831

### DIFF
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -394,7 +394,7 @@ JS
     {
         $default = $this->i18n_singular_name() ?: 'Block';
 
-        return _t(__CLASS__ . '.BlockType' , $default);
+        return _t(__CLASS__ . '.BlockType', $default);
     }
 
     /**
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -392,7 +392,9 @@ JS
      */
     public function getType()
     {
-        return _t(__CLASS__ . '.BlockType', 'Block');
+        $default = $this->i18n_singular_name() ?: 'Block';
+
+        return _t(__CLASS__ . '.BlockType' , $default);
     }
 
     /**


ScopeyNZ:
Thanks for the PR @mfendeksilverstripe. You'll need to target `4` for this change as it's an enhancement.

I _believe_ that we didn't use `i18n_singular_name` here on purpose, but I can't remember why... Maybe Robbie will be able to shed some light.

mfendeksilverstripe:
@ScopeyNZ Rebased on `4`.

brynwhyman:
> I believe that we didn't use i18n_singular_name here on purpose, but I can't remember why... Maybe Robbie will be able to shed some light.

People might avoid this PR due to this comment. Let's see if @robbieaverill remembers ;P

src/Models/BaseElement.php
```diff
@@ -368,7 +368,7 @@ public function getCMSFields()
      */
     public function getType()
     {
-        return _t(__CLASS__ . '.BlockType', 'Block');
+        return $this->i18n_singular_name();
```
emteknetnz, line 371:
```suggestion
        return _t(__CLASS__ . '.BlockType' , $this->i18n_singular_name());
```
Seems like it should be this so that if anyone has defined a specific translation in yml for the block then it should use this before defaulting to i18n_singular_name(), otherwise we'll create a translation regression

emteknetnz (changes requested):
I've made a recommended change - would this work for you?

src/Models/BaseElement.php
mfendeksilverstripe: Yes, this is a great suggestion, thanks @emteknetnz .

emteknetnz:
Looking through previous comments + code I think the reason not to use `i18n_singular_name` was because if no translation yml or config defines singlular name, it will back to `ClassInfo::shortName($this)` which is the classname minus the namespace.  Depending on the name of the class, that may be worse reading than simply "Block"

However `i18n_singular_name` is used a lot throughout Silverstripe, so I'm happy as is

@mfendeksilverstripe there's a PHPCS issue to fix before merging https://app.travis-ci.com/github/silverstripe/silverstripe-elemental/jobs/544323711#L1298

mfendeksilverstripe:
@emteknetnz Thanks for feedback, I've pushed up the listing fix.

emteknetnz (approved):
Travis failure in PHP8 behat is existing

PR #832, opened 2020-09-03
NEW: Top Page enhancements.

mfendeksilverstripe:
# Top Page enhancements

WIP


chrispenny:
Just putting down some thoughts, I'm not sure if they're totally relevant here or not.

I have a module which attempts to generate yml fixtures from DataObjects (and this includes adding relationships and those related DataObjects).

One of the limitations that we have with fixtures is that you cannot include looping relationships.

EG:
- `Page` `has_one` `ElementalArea`
- `ElementalArea` `has_one` `Page` (which, we now have through `TopPage\DataExtension`)

When I go to generate a fixture for this particular relationship, I must determine which DataObject "comes first" in the fixture, but if the relationship loops, I am unable to do that. Currently, I must remove any of these sorts of "looping relationships".

### How does the above relate here?

My gut feeling is that the `TopPage` relationship on `ElementalArea` should not be a `has_one`, but instead should be a [`belongs_to`](https://docs.silverstripe.org/en/4/developer_guides/model/relations/#belongs-to), as it is already a `owned` `has_one` on whichever `Page` class applied the extension.

I believe, however, that the relationship from a `BaseElement` to `Page` could remain a `has_one` - as the `BaseElement` is not directly owned by the `Page`.

emteknetnz:
> My gut feeling is that the TopPage relationship on ElementalArea should not be a has_one

Much too late for changing that, it cannot be change for BC/semver.  It exists as a has_one in stable releases.

src/TopPage/DataExtension.php
```diff
@@ -238,4 +298,88 @@ protected function clearTopPage(): void
...
+    {
+        // Classes are ordered from generic to specific, top-down, left-right
+        $classes = ClassInfo::dataClassesFor($this->owner);
+
+        if (count($classes) === 0) {
+            return null;
+        }
+
+        // Find the first ancestor table which has the extension applied
+        // Note that this extension is expected to be sublassed
```
emteknetnz, line 374: typo `subclassed`

src/TopPage/FluentExtension.php
```diff
@@ -39,10 +39,49 @@ protected function assignTopPage(Page $page): void
...
+    {
+        parent::assignFixedTopPage();
+
+        $this->owner->TopPageLocale = FluentState::singleton()->getLocale();
+    }
+
+    /*
+     * @inheritdoc
+     */
+    protected function saveChanges(): void
```
emteknetnz, line 65: There's code duplication going on here so Idon't  think we need this function, instead change the parent function to `DataExtension::saveChanges(array $extraData = [])`

In the parent function in the SQLUpdate have this for the 2nd arg:

```
array_merge(['"TopPageID"' => $owner->TopPageID], $extraData)
```

src/TopPage/DataExtension.php
```diff
@@ -205,13 +231,47 @@ public function withTopPageUpdate(bool $update, callable $callback)
...
+     * For example: model duplication where parent is assigned and saved only after the duplication is done
+     * It's not possible to determine top page in such case however it might be possible to know the top page
+     * even before the operation starts from the specific context
+     * Setting the page id to null disables this feature
+     *
+     * @param int|null $topPageId
+     * @param callable $callback
+     * @return mixed
+     */
+    public function withFixedTopPage(?int $topPageId, callable $callback)
```
emteknetnz, line 246: Very odd having `?int` for the first arg.  In what scenario would you call `$myobj->withFixedTopPage(null, function ...)`?

src/TopPage/DataExtension.php
```diff
@@ -238,4 +298,88 @@ protected function clearTopPage(): void
...
+        $query->execute();
+    }
+
+    /**
+     * Perform a page lookup based on cached data
+     *
+     * @param int $id
+     * @return Page|null
+     */
+    protected function getTopPageFromCachedData(int $id): ?Page
```
emteknetnz, line 348: This function can be deleted, just call `$page = Page::get_by_id($id);` in the one place this is used

emteknetnz (changes requested):
Code looks good so far, some recommended changes

mfendeksilverstripe:
Thanks for feedback @chrispenny . The relationship setup `ElementalArea` `has_one` `Page` is intentional as it has performance implications. This feature is meant to provide fast lookup bottom up. This can only be achieved with this setup. Using `belongs_to` is simply too slow.

I agree that from logical point of view it may look like a redundancy. I recommend viewing this as a separate relation.

My suggestion is to add a back relation config which would make these two relations distinguishable as they should not be treated as one relation.

src/TopPage/DataExtension.php
```diff
@@ -40,12 +43,26 @@ class DataExtension extends BaseDataExtension
...
     private $topPageUpdate = true;
 
+    /**
+     * Global flag which indicates that automatic page determination is enabled or not
+     * If this is set to a page ID it will be used instead of trying to determine the top page
+     *
+     * @see DataExtension::withFixedTopPage()
+     * @var int|null
+     */
+    private $fixedID = null;
```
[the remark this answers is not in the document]
mfendeksilverstripe: Makes sense, I'll make the change.

src/TopPage/DataExtension.php
mfendeksilverstripe: `null` is a valid value to pass to this feature which effectively disables it. I agree that this is probably a marginal case but it may come in handy for example in unit tests.

src/TopPage/DataExtension.php
```diff
@@ -205,13 +231,47 @@ public function withTopPageUpdate(bool $update, callable $callback)
...
+            $this->fixedID = $original;
+        }
+    }
+
+    /**
+     * Get the ID of a page which is currently set as the fixed top page
+     *
+     * @return int|null
+     */
+    protected function getFixedId(): ?int
```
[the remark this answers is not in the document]
mfendeksilverstripe: Good point, I'll update this.

src/TopPage/DataExtension.php
mfendeksilverstripe: This is intentionally here to allow extensibility. Bespoke projects may override this method to customise how it works. I was thinking about adding an extension point here but it's not flexible enough as you may want to completely alter the behaviour here as opposed to just add stuff on top of the default. This is a performance driven feature after all. Example cases:

* inject tracking of this feature (tracks when this feature looks up pages from cached data)
* feature flag / enable / disable this feature
* load different page / page from different data source

I'll update the docs here to make it more clear why this is needed.

src/TopPage/DataExtension.php
```diff
@@ -238,4 +298,88 @@ protected function clearTopPage(): void
...
+     * Find table name which has the top page fields
+     *
+     * @return string|null
+     */
+    protected function getTopPageTable(): ?string
+    {
+        // Classes are ordered from generic to specific, top-down, left-right
+        $classes = ClassInfo::dataClassesFor($this->owner);
+
+        if (count($classes) === 0) {
```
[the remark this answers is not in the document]
mfendeksilverstripe: Nice spotting, thanks.

src/TopPage/FluentExtension.php
mfendeksilverstripe: I'll make the change, happy to have the code more DRY.

mfendeksilverstripe:
The failing tests are an upstream error as shown in my test PR:

https://github.com/silverstripe/silverstripe-elemental/pull/884

src/TopPage/FluentExtension.php
```diff
@@ -39,10 +39,36 @@ protected function assignTopPage(Page $page): void
...
+    /*
+     * @inheritdoc
+     */
+    protected function saveChanges(array $extraData = []): void
+    {
+        /** @var DataObject|FluentExtension $owner */
+        $owner = $this->owner;
+
+        parent::saveChanges([
+            '"TopPageLocale"' => $owner->TopPageLocale,
```
emteknetnz, line 71: You need to get the `$extraData` param here passed to DataExtension::saveChanges()

src/TopPage/DataExtension.php
```diff
@@ -205,13 +231,47 @@ public function withTopPageUpdate(bool $update, callable $callback)
...
+            $this->fixedTopPageID = $original;
+        }
+    }
+
+    /**
+     * Get the ID of a page which is currently set as the fixed top page
+     *
+     * @return int
+     */
+    protected function getFixedTopPageId(): int
```
emteknetnz, line 263: Convention is to use `ID` rather that `Id`

```suggestion
    protected function getFixedTopPageID(): int
```

src/TopPage/DataExtension.php
```diff
@@ -238,4 +298,95 @@ protected function clearTopPage(): void
...
     }
+
+    /**
+     * Assigns top page relation based on fixed id
+     *
+     * @see DataExtension::withFixedTopPage()
+     */
+    protected function assignFixedTopPage(): void
+    {
+        $this->owner->TopPageID = $this->getFixedTopPageId();
```
emteknetnz, line 309:
```suggestion
        $this->owner->TopPageID = $this->getFixedTopPageID();
```

src/TopPage/DataExtension.php
```diff
@@ -144,22 +170,22 @@ public function setTopPage(?Page $page = null): void
             return;
         }
 
-        $page = $page ?? $owner->getTopPage();
+        if ($this->getFixedTopPageId() > 0) {
```
emteknetnz, line 173:
```suggestion
        if ($this->getFixedTopPageID() > 0) {
```

src/TopPage/DataExtension.php
```diff
@@ -205,13 +231,47 @@ public function withTopPageUpdate(bool $update, callable $callback)
...
 
+    /**
+     * Use this to wrap any code which is supposed to run with fixed top page
+     * Useful when top page is known upfront and doesn't need to be determined
+     * For example: model duplication where parent is assigned and saved only after the duplication is done
+     * It's not possible to determine top page in such case however it might be possible to know the top page
+     * even before the operation starts from the specific context
+     * Setting the page id to 0 disables this feature
+     *
+     * @param int $topPageId
```
emteknetnz, line 242:
```suggestion
     * @param int $topPageID
```

src/TopPage/DataExtension.php
```diff
@@ -205,13 +231,47 @@ public function withTopPageUpdate(bool $update, callable $callback)
...
+     * @param callable $callback
+     * @return mixed
+     */
+    public function withFixedTopPage(int $topPageId, callable $callback)
```
emteknetnz, line 246:
```suggestion
    public function withFixedTopPage(int $topPageID, callable $callback)
```

src/TopPage/DataExtension.php
```diff
@@ -205,13 +231,47 @@ public function withTopPageUpdate(bool $update, callable $callback)
...
+    {
+        $original = $this->fixedTopPageID;
+        $this->fixedTopPageID = $topPageId;
```
emteknetnz, line 249:
```suggestion
        $this->fixedTopPageID = $topPageID;
```

tests/TopPage/TopPageTest.php
```diff
@@ -170,63 +170,78 @@ function () use ($populateTopPage): void {
...
+    /**
+     * This test is checking for page duplication in two cases
+     * Case 1: standard duplication
+     * Case 2: duplication with a fixed page setting
+     * The seconds case shows that it's possible to use the withFixedTopPage to set the top page to arbitrary value
+     * and completely bypass page determination logic
+     * This is needed in some edge cases were automatic determination is not possible due to the object not being
+     * assigned to the parent object at the time of duplication but rather later
+     *
+     * @param int $fixedPageId
```
emteknetnz, line 182:
```suggestion
     * @param int $fixedPageID
```

tests/TopPage/TopPageTest.php
```diff
@@ -170,63 +170,78 @@ function () use ($populateTopPage): void {
...
+     * @dataProvider fixedPagesProvider
+     */
+    public function testPageDuplication(int $fixedPageId): void
```
emteknetnz, line 185:
```suggestion
    public function testPageDuplication(int $fixedPageID): void
```

tests/TopPage/TopPageTest.php
```diff
@@ -170,63 +170,78 @@ function () use ($populateTopPage): void {
...
     {
         /** @var TopPage\DataExtension $extension */
         $extension = singleton(TopPage\DataExtension::class);
-        $extension->withTopPageUpdate(
-            true,
-            function (): void {
-                $this->populateTopPageForAllObjects();
+        $extension->withFixedTopPage($fixedPageId, function () use ($extension, $fixedPageId) {
```
emteknetnz, line 189:
```suggestion
        $extension->withFixedTopPage($fixedPageID, function () use ($extension, $fixedPageID) {
```

tests/TopPage/TopPageTest.php
```diff
@@ -170,63 +170,78 @@ function () use ($populateTopPage): void {
...
+            $extension->withTopPageUpdate(
+                true,
+                function () use ($fixedPageId): void {
```
emteknetnz, line 192:
```suggestion
                function () use ($fixedPageID): void {
```

tests/TopPage/TopPageTest.php
```diff
@@ -170,63 +170,78 @@ function () use ($populateTopPage): void {
...
-                    $this->assertCount(2, $items);
+                        $this->assertCount(2, $items);
 
-                    /** @var DataObject|TopPage\DataExtension $objectClone */
-                    $objectClone = $items->first();
+                        /** @var DataObject|TopPage\DataExtension $objectClone */
+                        $objectClone = $items->first();
 
-                    $this->assertEquals((int) $page->ID, (int) $objectClone->TopPageID);
+                        $expected = $fixedPageId ?: (int) $page->ID;
```
emteknetnz, line 234:
```suggestion
                        $expected = $fixedPageID ?: (int) $page->ID;
```

emteknetnz (changes requested):
Minor changes then happy to merge

mfendeksilverstripe:
@emteknetnz , I've pushed up the requested changes and also rebased the branch so hopefully the tests will now pass.

emteknetnz (approved):

PR #851, opened 2020-11-17
DEP Use graphql 3

emteknetnz:
~~Use graphql 3.x-dev when installing the 4.x-dev recipe~~

Specify graphql 3 in composer.json

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,6 +21,7 @@
         "silverstripe/admin": "^1.7@dev",
         "silverstripe/versioned": "^1.7@dev",
         "silverstripe/versioned-admin": "^1.7@dev",
+        "silverstripe/graphql": "^3.4",
         "symbiote/silverstripe-gridfieldextensions": "^3.1",
         "silverstripe/vendor-plugin": "^1"
     },


robbieaverill (commented):
Why don't you constrain this release line to GraphQL 3, then release a new minor elemental version which works on GraphQL 4?

emteknetnz:
The issue is with silverstripe/recipe-cms:4.x-dev which we install on .travis.yml, not elemental itself

Travis currently has a backlog of builds to run, though this branch is green on creative-commoners travis account https://travis-ci.com/github/creative-commoners/silverstripe-elemental/builds/201924599

robbieaverill:
Elemental does have a direct dependency on GraphQL though. I'm not as familiar with the decisions made here, but it does seem to me like this client-level abstraction you're proposing is just going to become a nightmare to maintain.

emteknetnz:
Elemental requires versioned-admin which requires graphql is how things currently work. 
We're currently mid release so I do not want to touch composer.json requirements right now

Graphql 4 compatibility in core modules was merged on 1.x-dev/4.x-dev branches in the last few days which has complicated travis builds for matrix's that install recipe 4.x-dev for the following modules:
- silverstripe-elemental
- silverstripe-versioned-admin
- silverstripe-asset-admin

>>  client-level abstraction you're proposing is just going to become a nightmare to maintain

I don't think it will? At this stage we're only expecting to have to update 3 travis files, and we'll probably be migrating to a shared travis config within the month which will make this sort of thing even easier

robbieaverill:
Ok, when 4.8 is released what are you intending to do with this change? In my mind if we're supporting GraphQL 3 and 4 then it'd be worth running tests on both versions - at least for behat

emteknetnz:
I'm not sure because I don't know if graphql will have a stable tag then. We will have the option of removing it because elemental will be compatible because of these unreleased commits on elemental 4.x-dev

image

If we wish to test both, then we could simply create another matrix entry to test on graphql 3 - silverstripe-admin does this now

image

emteknetnz:
Chatted to @Cheddam about this, we've agreed to change this to making the change in composer.json

emteknetnz:
Travis doesn't appear to be working on this repo/account?

Travis is green on creative-commoners - https://travis-ci.com/github/creative-commoners/silverstripe-elemental

Cheddam (approved):

chillu:
I've sent a follow up to restore GraphQL 4 support: https://github.com/dnadesign/silverstripe-elemental/pull/865

@emteknetnz Please ping @unclecheese or myself when you make those kinds of changes, when there's a significant overlap with our work (and it reverts an earlier change by Aaron).

PR #852, opened 2020-11-18
Use DBClassName so that dev/build can migrate legacy classnames

tractorcow:
I had a few issues where I'd register a class name map with DatabaseAdmin, but elemental wasn't picking this up.

Slashes are double escaped intentionally, in case it looks weird. :P

### DIFF
--- a/src/Models/ElementalArea.php
+++ b/src/Models/ElementalArea.php
@@ -25,7 +25,7 @@ use SilverStripe\Versioned\Versioned;
 class ElementalArea extends DataObject
 {
     private static $db = [
-        'OwnerClassName' => 'Varchar(255)',
+        'OwnerClassName' => 'DBClassName("SilverStripe\\\\ORM\\\\DataObject")',
     ];
 
     private static $has_many = [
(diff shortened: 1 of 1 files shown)


src/Models/ElementalArea.php
```diff
@@ -25,7 +25,7 @@
 class ElementalArea extends DataObject
 {
     private static $db = [
-        'OwnerClassName' => 'Varchar(255)',
+        'OwnerClassName' => 'DBClassName("SilverStripe\\\\ORM\\\\DataObject")',
```
michalkleiner, line 28: Should it be `DBClassName("SilverStripe\\\\ORM\\\\FieldType\\\\DBVarchar")` to keep it effectively a Varchar? Don't understand the change to DataObject.

tractorcow: That's the base class that DBClassName should use. The Value of 'OwnerClassName' is a subclass of dataobject.

robbieaverill: @michalkleiner I've just skim read this, but the change from varchar to DBClassName is to make it a proper polymorphic relationship, which means that dev/build in framework can perform modifications on it like it does for other polymorphic relations.

tractorcow: We should build a proper polymorphic belongs_to in core right @robbieaverill ? :P

robbieaverill: Yeah! I guess it this were replaced with Owner -> DataObject then it'd be even more real

PR #858, opened 2020-12-09
WIP FIX Don't display element actions to user if they cant select them

scott1702:
Resolves https://github.com/dnadesign/silverstripe-elemental/issues/856

**ToDo**
- [x] Tests


client/src/components/ElementActions/PublishAction.js
```diff
@@ -122,7 +122,10 @@ const PublishAction = (MenuComponent) => (props) => {
     <MenuComponent {...props}>
       {props.children}
 
-      {(formDirty || !element.IsLiveVersion) && <AbstractAction {...newProps} />}
+      {props.element.canPublish === false
```
emteknetnz, line 125: You could do these sorts of checks without the `===` and ternary?
`{(formDirty || !element.IsLiveVersion) && props.element.canPublish && <AbstractAction {...newProps} />}`

src/Models/BaseElement.php
```diff
@@ -80,6 +80,10 @@ class BaseElement extends DataObject
         'BlockSchema' => DBObjectType::class,
         'IsLiveVersion' => DBBoolean::class,
         'IsPublished' => DBBoolean::class,
+        'canCreate' => DBBoolean::class,
```
emteknetnz, line 83: Adding to a private static would be considered a semver addition so this will need to be branched off and target `4` instead of `4.4`

emteknetnz (changes requested):
Looks good:
- needs to target `4` due to semver
- could make the checks a bit more concise

src/Models/BaseElement.php
scott1702: Ah had targeted 4.4 assuming this would count as a bug fix. But if there are specific rules about adding to private static needed to bump semver then can target 4

client/src/components/ElementActions/PublishAction.js
scott1702: Wasn't super confident in the GraphQL stuff so hard checked `false` in case it canPublish didn't come through for some reason. But makes sense to assume it will always be there and shorthand this

src/Models/BaseElement.php
brynwhyman: > But if there are specific rules about adding to private static

Hey @scott1702 here's some more information on this:
* [Silverstripe CMS public APIs](https://docs.silverstripe.org/en/4/contributing/code/#picking-the-right-version) include private static class properties; and 
* According to SEMVER, a minor release "[MUST be incremented if new, backwards compatible functionality is introduced to the public API.](https://semver.org/#spec-item-7)"

scott1702: Perfect, thanks @brynwhyman!

emteknetnz:
Closing this in favour of https://github.com/dnadesign/silverstripe-elemental/pull/874 so that I can tweak the PR to get this one over the line

I'll put Scott down as a co-author in any further commit messages

scott1702:
Thanks @emteknetnz! Sorry I didn't get around to doing it myself

PR #859, opened 2020-12-09
Validation on $element before Duplicate

Jianbinzhu:
Check if  is canCreate before duplicate element. This will resolve https://github.com/dnadesign/silverstripe-elemental/issues/857

### DIFF
--- a/src/GraphQL/DuplicateElementMutation.php
+++ b/src/GraphQL/DuplicateElementMutation.php
@@ -56,6 +56,11 @@ class DuplicateElementMutation extends MutationCreator implements OperationResol
                 "The current user has insufficient permission to edit ElementalArea: $areaID"
             );
         }
+        if (!$element->canCreate($content['currentUser'])) {
+            throw new InvalidArgumentException(
+                "The current user has inssufficient permission to duplicate BaseElement: $elementID"
+            );
+        }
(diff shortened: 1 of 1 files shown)


src/GraphQL/DuplicateElementMutation.php
```diff
@@ -56,6 +56,11 @@ public function resolve($object, array $args, $context, ResolveInfo $info)
                 "The current user has insufficient permission to edit ElementalArea: $areaID"
             );
         }
+        if (!$element->canCreate($content['currentUser'])) {
```
emteknetnz, line 59: Seems less than ideal fixing this only in elemental rather than in the graphql module itself - is there a way we could fix this there instead?

emteknetnz: I've had a look withing the graphql module and I cannot find anything explicitly for handling duplication there, so I guess it makes sense to do this here

emteknetnz:
@Jianbinzhu I've picked up this issue in a [new PR](https://github.com/dnadesign/silverstripe-elemental/pull/877) to try and get this one over the line.  There's some additional UX changes that need to be made, though we can take care of those for you.

I've put you down as as a co-author in the git commit message

PR #863, opened 2021-01-11
v4: Add an extension point to BaseElement::AbsoluteLink()

chrispenny:
There is already a similar extension point available in the `Link()` method. Could we please also have this one available in `AbsoluteLink()`?

My current project uses direct routes for Elements (EG: `my-page/my-element`), so we would like to please be able to update both `Link()` and `AbsoluteLink()` appropriately.

Thank you!!

### DIFF
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -615,6 +615,8 @@ JS
         if ($page = $this->getPage()) {
             $link = $page->AbsoluteLink($action) . '#' . $this->getAnchor();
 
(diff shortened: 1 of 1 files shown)


chrispenny:
Failing test is the same exit code that exists on the current `master` build:
https://travis-ci.org/github/dnadesign/silverstripe-elemental/jobs/753469246

dnsl48 (approved):
lgtm

dnsl48:
The broken test is about PHPCS not installed. Irrelevant for this PR.
All PHPUnit tests are green.

PR #865, opened 2021-01-18
DEP Restore GraphQL 4 support

chillu:
This was originally added to the 4 release branch,
and then removed prior to the 4.5.0 release through
https://github.com/dnadesign/silverstripe-elemental/pull/851
to simplify Travis builds.

Explicitly declaring the module *doesn't* support GraphQL 4
is incorrect, and hinders early adoption - you'll need to
run a "silverstripe/graphql:4.x-dev as 3.x-dev" dependency.

It's also a different strategy compared to e.g. versioned-admin:
https://github.com/silverstripe/silverstripe-versioned-admin/commit/cae464fd25494e24d1e05d863c7bc1d4221d8d57

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "silverstripe/admin": "^1.7@dev",
         "silverstripe/versioned": "^1.7@dev",
         "silverstripe/versioned-admin": "^1.7@dev",
-        "silverstripe/graphql": "^3.4",
+        "silverstripe/graphql": "^3.4 || 4.x-dev",
         "symbiote/silverstripe-gridfieldextensions": "^3.1",
         "silverstripe/vendor-plugin": "^1"
     },
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
         "silverstripe/vendor-plugin": "^1"
     },
     "require-dev": {
-      "sminnee/phpunit": "^5.7"
+      "silverstripe/recipe-testing": "1.x-dev"
     },
     "suggest": {
         "silverstripe/elemental-blocks": "Adds a set of common SilverStripe content block types",


chillu:
Talked to @emteknetnz this will be fixed differently - by adding REQUIRE_GRAPHQL to the preconfigured build variations in https://github.com/silverstripe/silverstripe-travis-shared. We'll still need to restore the `composer.json` dependency bit with this PR though.

.travis.yml
```diff
@@ -7,4 +7,16 @@ env:
   global:
     - COMPOSER_ROOT_VERSION="4.x-dev"
     - REQUIRE_RECIPE="4.x-dev"
+    # Ensure builds default to current stable release.
+    # Run a specialised "preview release job" with GraphQL v4, see below.
+    - REQUIRE_GRAPHQL="3.x-dev"
```
emteknetnz, line 12: Won't need to update .travis.yml since we'll now do this in shared-config https://github.com/silverstripe/silverstripe-travis-shared/pull/4

composer.json
```diff
@@ -21,7 +21,7 @@
         "silverstripe/admin": "^1.7@dev",
         "silverstripe/versioned": "^1.7@dev",
         "silverstripe/versioned-admin": "^1.7@dev",
-        "silverstripe/graphql": "^3.4",
+        "silverstripe/graphql": "^3.4 || 4.x-dev",
```
emteknetnz, line 24: Would it be OK to make this change in `4.5` branch of elemental instead to increase test coverage? Or is to too much of change?

emteknetnz (changes requested):

.travis.yml
chillu: Yep I'll remove that again

composer.json
chillu: True, GraphQL v4 support was merged before the 4.5.0 stable release. So this would allow early adopters to run a non-stable version of GraphQL while staying on a stable version of Elemental. But then again, I don't really want to encourage that since we might need to further update elemental to keep up with changes in GraphQL v4. So on balance, I think the change is targeted correctly at `4`

emteknetnz (changes requested):
Could I get you to change the `sminnee/phpunit` in require-dev to `"silverstripe/recipe-testing": "1.x-dev"` so that it matches `4.5` .. this will include the phpcs requirement which is what causing the travis failure

Possibly this was missed during a merge-up
https://github.com/dnadesign/silverstripe-elemental/blob/4.5/composer.json#L29

chillu:
@emteknetnz Changed to recipe-testing now, keeping an eye on the build

Issue #866, opened 2021-01-20
PHP8 support


emteknetnz:
**ACs**
- Ensure that silverstripe/elemental works with PHP8
- Update composer.json to allow PHP8 installation e.g. "php": "^7.1 || ^8.0" (PR: https://github.com/silverstripe/silverstripe-elemental/pull/867)
- Ensure that this https://github.com/dnadesign/silverstripe-elemental/pull/874 and any other high impact bug fixes targeting next-minor `4` are also merged before tagging
- ~Tag 4.6.0 which also includes graphql4 support~
- Ensure the folowing elemental modules are also PHP8 compatible and installable:
  - silverstripe/elemental-fileblock
  - silverstripe/elemental-bannerblock 
  - dnadesign/silverstripe-elemental-userforms
  - dnadesign/silverstripe-elemental-subsites 
- Run through manual regression test suite for elemental
- Consider a module release following the work

obj63mc:
Was looking to update a few of my sites to php8 and with 4.6.0 released currently it seems that only Silverstripe Elemental is the only module not fully supporting php8 yet.  v4.3.2 works great with php8 so assuming its more the changes to graphql4 that need attention.  Just wondering if this is on the roadmap this year or what can be done to help get this over the line

emteknetnz:
I don't believe GraphQL 4 is a blocker in regards to PHP8 support.

We're aiming to get PHP8 support across the board for modules that Silverstripe Ltd supports later this year.  Currently that's most probably most things, but not everything, so we haven't announced official support yet.

tractorcow:
4.6.0 broke compatibility with graphql 3 (see https://github.com/silverstripe/silverstripe-elemental/pull/870), now that graphql typeNames map is no longer respected. Can you update the composer dependencies to reflect that?

graphql 3 users upgrading their silverstripe versions will need to lock to ~4.5.0 of elemental.

emteknetnz:
OK that wasn't supposed to happen, elemental 4.6.0 should work on both versions of graphql.  Thanks for pointing this out.

emteknetnz:
Graphql 4.6.0 convo - https://github.com/silverstripe/silverstripe-elemental/pull/870
PR (merged) https://github.com/silverstripe/silverstripe-elemental/pull/912

maxime-rainville:
There's nothing left to do on this.

PR #867, opened 2021-01-26
DEP PHP8 support

emteknetnz:


brynwhyman:
@emteknetnz, I reran the tests and the PHP8 travis tests are now passing, any reason why we shouldn't merge this? I've seen projects having to fork this module to get it running with PHP8 and avoid the composer constraint - even though there are no failures explicitly with this module.

We still have https://github.com/silverstripe/silverstripe-elemental/issues/866 which can stay open until the remaining tasks are done?

emteknetnz:
I've lost context on where the php8 work go to.  It's probably OK to merge this, we have the equivalent constraint within framework https://github.com/silverstripe/silverstripe-framework/blob/4/composer.json#L42

brynwhyman (approved):

PR #869, opened 2021-01-26
BUG AddElementPopoverComponent button is submit.

GuySartorelli:
By omitting the button type, this is automatically a submit button. This means pressing the enter key while in an input field triggers this button instead of a true form submit button, and also sends unnecessary form submit events when the form isn't actually being submitted.

### DIFF
--- a/client/src/components/ElementEditor/HoverBar.js
+++ b/client/src/components/ElementEditor/HoverBar.js
@@ -21,6 +21,7 @@ function StatelessHoverBar({
   const lineClasses = `${classNames('-line')} font-icon-plus-circled`;
   const label = i18n._t('ElementAddNewButton.ADD_BLOCK', 'Add block');
   const btnProps = {
+    type: 'button',
     className: classNames('-area', { '-area--focus': popoverOpen }),
     onClick: onToggle,
     'aria-label': label,


emteknetnz:
Unfortunately this pull-request changes the behavior in a bad way, at least in Firefox where I tested it.

Tabbing on to a hover bar then pressing `enter` will create a new element popover populated with the available block types.  This is the same behavior as clicking on the hover bar with a mouse.

You can then press enter again and it will dismiss the popover without adding a new element.

However with this pull-request, the behavior changes so that when you press enter the second time to dismiss the popover, instead it will cause the 'Content' tab in the top right tab set to open, which is very undesirable.

Sorry, I'll need to close this one.

michalkleiner:
Why close the PR and not let the author deal with the provided testing feedback?

GuySartorelli:
Because this has been closed, I've raised #890 for someone else to take a look at fixing.

emteknetnz:
> Why close the PR and not let the author deal with the provided testing feedback?

That's a fair point.

My feeling is that there would significant amount of underlying rework outside of just altering the hover bar component to get this to work.  With the approach taken in this pull request, when pressing enter while focused on the hover bar component was causing a completely different UX component to open.  It seems like a different approach is required.

PR #870, opened 2021-01-27
NEW: Compatibility with GraphQL 4 changes

unclecheese:
## Summary

Use new context provider API. No more BuildState.

Depends on https://github.com/silverstripe/silverstripe-graphql/pull/352


unclecheese:
Confirmed failures are due to missing `UserContextProvider` class.

tractorcow:
This change just broke compatibility with graphql 3. Do you want to fix the composer dependencies to reflect that 3 is no longer supported?

unclecheese:
Shouldn't have. We test everything in both v3 and v4. https://travis-ci.com/github/silverstripe/silverstripe-elemental/jobs/519607824

What issue are you seeing?

tractorcow:
my typeMap is no longer respected due to the removal of 
`return StaticSchema::inst()->typeNameForDataObject(static::class);`

tractorcow:
I guess my attempt to upgrade a framework 4.7 to 4.8 can't be done unless I upgrade to graphql 4 huh. Oh well.

unclecheese:
We are supporting both versions simultaneously. The method you're referring to still exists in GraphQL 3, right? Can't you just run GraphQL 3?

tractorcow:
Yes, the method exists, and elemental no longer calls it.

So with graphql config as below:

```yaml
      typeNames:
        myProject\HeroBlock: HeroBlock
```


I still get graphql results as below:

```json
"node": {
      "ID": "6",
      "BlockSchema": {
        "typeName": "myProject_HeroBlock",
        "title": "About the collection / Mō te kohinga",
        "showTitle": 1
  }
}
```

when it should be

```json
"node": {
      "ID": "6",
      "BlockSchema": {
        "typeName": "HeroBlock",
        "title": "About the collection / Mō te kohinga",
        "showTitle": 1
  }
}
```

unclecheese:
Right, OK.. The "backward compatibility" is focussed on the CMS, so there may still be upgrade issues in your frontend app. I think this is really as simple as adding a fork in the typeName function that uses the old GraphQL 3 method if it exists, right? Can we fix this in two lines of code?

unclecheese:
```php
    public static function getGraphQLTypeName(): string
    {
        return class_exists(StaticSchema::class)
            ? StaticSchema::inst()->typeNameForDataObject(static::class)
            : str_replace('\\', '_', static::class);
    }
```

tractorcow:
That'd probbably do it!

Sorry, i've been out of the silverstripe space for a while, not sure what the state of play is with the recent versions. :)

tractorcow:
Thanks for the quick response too. I don't need a fix urgently, I just wanted to report back my findings in case it was useful for you. I'm happy to keep my site on 4.7 in the mean time.

tractorcow:
convention over configuration is the way to go by the way. graphql 4 is very cool.

unclecheese:
Thanks! If you like convention over configuration, you'll love what we're doing with NextJS. :-)

unclecheese:
Can you just apply that patch locally and let me know if it works, and if so, I'll just make a PR now.

tractorcow:
Will do, 1 moment.

tractorcow:
looks good 👍 

```json
"node": {
                "ID": "6",
                "BlockSchema": {
                  "typeName": "HeroBlock",

```

tractorcow:
When I get some time, I'll just upgrade to graphql 4 (that's the real fix here, I have to admit).

unclecheese:
PR is up https://github.com/silverstripe/silverstripe-elemental/pull/912

PR #872, opened 2021-02-08
BUG ignored_classes Content field emptied by MigrateContentToElement task

GuySartorelli:
Classes in the ignored_classes configuration array don't have an elemental area. This fixes a bug where those classes have their Content field cleared even though it cannot be migrated into an elemental block.

### DIFF
--- a/src/Tasks/MigrateContentToElement.php
+++ b/src/Tasks/MigrateContentToElement.php
@@ -138,6 +138,9 @@ class MigrateContentToElement extends BuildTask
     protected function isMigratable($pageType)
     {
         $migratable = SiteTree::has_extension($pageType, ElementalPageExtension::class);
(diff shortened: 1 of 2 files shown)


Cheddam (commented):
Hey @GuySartorelli, thanks for contributing this patch! Could you write a [unit test case](https://github.com/dnadesign/silverstripe-elemental/blob/4/tests/Tasks/MigrateContentToElementTest.php) to cover the change?

GuySartorelli:
Can do @Cheddam, though I can't promise it will be done any time soon; I just slapped this pr together while I was implementing a workaround in a project.

GuySartorelli:
Gonna set this to draft until I get the unit test sorted.

GuySartorelli:
@Cheddam I finally got around to adding a unit test for this :)

Cheddam (approved):
Thanks for following through, @GuySartorelli!

PR #873, opened 2021-02-09
BUG Tweak URL generation so blocks can be edited when added to an Area on a regular DO

maxime-rainville:
This is a WIP. Not ready for peer review.

# Parent issue
- Fixes https://github.com/dnadesign/silverstripe-elemental/issues/871

### DIFF
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -695,17 +695,20 @@ JS
             return $link;
         }
 
-        if (!$page instanceof SiteTree && method_exists($page, 'CMSEditLink')) {
-            $link = Controller::join_links($page->CMSEditLink(), 'ItemEditForm');
-        } else {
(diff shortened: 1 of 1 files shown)


maxime-rainville:
A dev would need to add this code snippet to their DO for this PR to work.

```php
    /**
     * Sanitise a model class' name for inclusion in a link
     *
     * @param string $class
     * @return string
     */
    protected function sanitiseClassName($class)
    {
        return str_replace('\\', '-', $class);
    }

    public function CMSEditLink()
    {
        return Controller::join_links(
            TestModelAdmin::singleton()->Link($this->sanitiseClassName(self::class)),
            'EditForm/field',
            $this->sanitiseClassName(self::class),
            'item',
            $this->ID,
            'edit'
        );
    }

    public function Link() {
        return 'frontend-link';
    }
```

GuySartorelli:
This was fixed in a separate PR

PR #874, opened 2021-02-14
ENH Disable actions users does not have permissions for

emteknetnz:
Fix for https://github.com/dnadesign/silverstripe-elemental/issues/856

~~Blocked by https://github.com/dnadesign/silverstripe-elemental/issues/875~~

Original PR https://github.com/dnadesign/silverstripe-elemental/pull/858

Useful code for testing to easily disable permissions on a content block

```php
<?php

{
    public function getType()
    {
        return 'My Block';
    }

    public function canCreate($member = false, $context = [])
    {
        return true;
    }

    public function canDelete($member = false)
    {
        return false;
    }

    public function canPublish($member = false, $context = [])
    {
        return true;
    }

    public function canUnpublish($member = false, $context = [])
    {
        return true;
    }
}

```

========================================================================

Old screenshots (can ignore now)

**Before:**

image

**After - remove permissions for canEdit, canCreate, canDelete (kept permissions for canView):**

image

**After - Clicking on 'Content'**

image


emteknetnz:
@clarkepaul This is a different PR from the [related PR](https://github.com/dnadesign/silverstripe-elemental/pull/877) where there was a UX discussion.

In this PR the options that cannot be selected are simply removed.  Since the work is already done here the cost to implement is significantly less.

If you'd still like the work done as per https://github.com/dnadesign/silverstripe-elemental/pull/877#issuecomment-782950166 then let me know, and we'll look to update this PR instead.  Not sure about complexity level though, it may push the timeline on this one back.

clarkepaul:
I'm not sure why the _Unpublish_ action is visible to a user that only has CanView permissions, seems wrong to me.

Because the availability of these block actions are based on permissions, we do need to let people know that. In that the UI looks different for them, and acts differently from what the user guides show. At least they will be able to ask for permissions to the admin.
I think my suggestion for having the action disabled still makes the most sense, the following should be possible:

_Duplicate_ action:
- "Duplicate" to be italicised and use the lighter text colour variable (muted)
- when hovering link shows not allowed cursor
- title text on hover to says "Duplicate, insufficient permissions"

_Archive_ action:
- "Archive" to be italicised and use the lighter text colour variable (muted)
- when hovering link shows not allowed cursor
- title text on hover to says "Archive, insufficient permissions"

I think the same should be able to apply to _Publish_, but _Unpublish_ is okay to hide if a block is only in draft.

I realise these patterns might not be exactly the same in other areas of the CMS but we should aim for best practices/UX.

Side note those readonly fields look pretty wack, which browser are you using @emteknetnz ?

emteknetnz:
Firefox for linux

emteknetnz:
OK, have implemented as requested.

I couldn't get both the tooltip and cursor in the same screenshot so I've taken 2

**Disabled action**

image

**Mouse over cursor**

Screenshot_20210223_112606

clarkepaul:
I think it's perfect thanks

client/lang/src/en.json
```diff
@@ -2,17 +2,21 @@
   "ElementArchiveAction.CONFIRM_DELETE": "Are you sure you want to send this block to the archive?",
   "ElementArchiveAction.CONFIRM_DELETE_AND_UNPUBLISH": "Warning: This block will be unpublished before being sent to the archive. Are you sure you want to proceed?",
   "ElementArchiveAction.ARCHIVE": "Archive",
+  "ElementArchiveAction.ARCHIVE_DISABLED": "Archive, insufficient permissions",
```
dnsl48, line 5: `ARCHIVE_DISABLED` key name does not imply any relevance with lacking permission.
(`disabled` conventionally means the feature is off altogether, not something perms related).
Feels like we might use a different name that could be more relevant to the use case (`canArchive` permission check)

client/lang/src/en.json
```diff
@@ -2,17 +2,21 @@
...
   "ElementArchiveAction.DUPLICATE": "Duplicate",
+  "ElementArchiveAction.DUPLICATE_DISABLED": "Duplicate, insufficient permissions",
```
dnsl48, line 7: `DUPLICATE_DISABLED` key relevance with permissions (same as ARCHIVE_DISABLED - see above)

client/lang/src/en.json
```diff
@@ -2,17 +2,21 @@
...
   "ElementHeader.NOTITLE": "Untitled {type} block",
   "ElementPublishAction.SUCCESS_NOTIFICATION": "Published '{title}' successfully",
   "ElementPublishAction.ERROR_NOTIFICATION": "Error publishing '{title}'",
   "ElementPublishAction.PUBLISH": "Publish",
+  "ElementPublishAction.PUBLISH_DISABLED": "Publish, insufficient permissions",
```
dnsl48, line 12: `PUBLISH_DISABLED` key relevance with permissions (same as ARCHIVE_DISABLED - see above)

client/lang/src/en.json
```diff
@@ -2,17 +2,21 @@
...
   "ElementSaveAction.SUCCESS_NOTIFICATION": "Saved '{title}' successfully",
   "ElementSaveAction.ERROR_NOTIFICATION": "Error saving '{title}'",
   "ElementSaveAction.SAVE": "Save",
   "ElementUnpublishAction.SUCCESS_NOTIFICATION": "Removed '{title}' from the published page",
   "ElementUnpublishAction.ERROR_NOTIFICATION": "Error unpublishing '{title}'",
   "ElementUnpublishAction.UNPUBLISH": "Unpublish",
+  "ElementUnpublishAction.UNPUBLISH_DISABLED": "Unpublish, insufficient permissions",
```
dnsl48, line 19: `UNPUBLISH_DISABLED` key relevance with permissions (same as ARCHIVE_DISABLED - see above)

dnsl48 (changes requested):
Looks good overall!
Just one note regarding the translation key namings.

dnsl48:
## Merge Checklist

This list helps to ensure that PRs are in a good state before merging. Ideally it should be applied to the PR upon
initial triage, so that the contributor can check items off prior to the reviewer digging in. Some items may not be
relevant to every PR, and can be crossed out on a case-by-case basis.

* [x] The target branch is [correct](https://docs.silverstripe.org/en/4/contributing/code/#picking-the-right-version)
* [x] All commits are relevant to the change (e.g. no debug statements or arbitrary linting)
* [x] The commit messages follow [the contribution guidelines](https://docs.silverstripe.org/en/4/contributing/code/#commit-messages)
* [x] The patch follows [the contribution guidelines](https://docs.silverstripe.org/en/4/contributing/code/)
* [x] New features are covered with tests (back-end with unit tests, front-end with Behat)
* [ ] ~Any relevant User Help/Developer documentation is updated; for impactful changes, information is added to the
      changelog for the intended release~
* [x] CI is green
* [ ] At least one peer reviewer approved; no outstanding changes requested

dnsl48 (approved):

PR #877, opened 2021-02-18
FIX Prevent users without create permissions from duplicating blocks

emteknetnz:
Fix for https://github.com/dnadesign/silverstripe-elemental/issues/857

Related https://github.com/dnadesign/silverstripe-elemental/pull/874 (deals with UX)

Original PR https://github.com/dnadesign/silverstripe-elemental/pull/859

I've put down James Zhu as a co-author in the commit message

You can easily simulate not having create permissions by creating a regular content block, the add `return false;` to the top of `BaseElement::canCreate()`

Currently this fix will:
- prevents the user from duplicating the content block
- change the graphql response
- add some test coverage

image

===============================================================================

**Below this line now being dealt with on https://github.com/dnadesign/silverstripe-elemental/pull/874:**

However the user experience is poor, there is no feedback to the end user that anything has happened

There are two ways I can think of that would resolve this:

a) Not show the duplicate option in the drop down if the user lacks create permissions~~

image

~~b) Still show the duplicate option, though show a red toast when a user clicks this with the message "The current user has insufficient permission to edit ElementalArea: $areaID"

Let me know what you'd like @clarkepaul~~

### DIFF
--- a/src/GraphQL/DuplicateElementMutation.php
+++ b/src/GraphQL/DuplicateElementMutation.php
@@ -56,6 +56,11 @@ class DuplicateElementMutation extends MutationCreator implements OperationResol
                 "The current user has insufficient permission to edit ElementalArea: $areaID"
             );
         }
+        if (!$element->canCreate($context['currentUser'])) {
+            throw new InvalidArgumentException(
+                "The current user has insufficient permission to create or duplicate BaseElement: $elementID"
+            );
+        }
 
         try {
             // clone element
--- a/tests/GraphQL/DuplicateElementMutationTest.php
+++ b/tests/GraphQL/DuplicateElementMutationTest.php
@@ -1,0 +1,64 @@
+<?php
+
+namespace DNADesign\Elemental\Tests\GraphQL;
+
+use DNADesign\Elemental\GraphQL\DuplicateElementMutation;
+use DNADesign\Elemental\Models\BaseElement;
+use DNADesign\Elemental\Models\ElementalArea;
+use GraphQL\Type\Definition\ResolveInfo;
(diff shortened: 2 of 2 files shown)


clarkepaul:
Whats the pattern if a user without create permission tries to add a block? might make sense for it to be similar.

Ideally I'd want a disabled look on the _Duplicate_ action with a tooltip saying insufficient permissions, but not sure if we have much flexibility to add additional bits to dropdowns? failing that opt.2 sounds like the next best option.

emteknetnz:
> Whats the pattern if a user without create permission tries to add a block? might make sense for it to be similar.

I manually changed the `BaseElement::canCreate()` return value to `false`, which means the user cannot create any type of block.  When I click the 'Add block' button I get the following

image

@clarkepaul Does this change your recommendation to have a disabled look on the duplicate button?

clarkepaul:
Well that experience isn't great :(. 
I think my suggestion for having the action disabled still makes the most sense, the following should be possible: 

- "Duplicate" to be italicised and use the lighter text colour variable (might be called muted?)
- when hovering link shows not allowed cursor 
- title text on hover to says "Duplicate, insufficient permissions"

Ignore the background colour on video as I had hover set using the browser inspector—hover colour will show when hovered.
https://youtu.be/HlQD7O-T2sM

emteknetnz:
@clarkepaul There's another PR that does deals with UX [here](https://github.com/dnadesign/silverstripe-elemental/pull/874) - apologies I only just realised the other PR would have been a better place to talk UX

This PR is really just a backend change which is probably fine as is.

Cheddam (approved):

PR #884, opened 2021-03-11
Test commit

mfendeksilverstripe:
Just testing if upstream tests pass.

### DIFF
--- a/_config/config.yml
+++ b/_config/config.yml
@@ -1,7 +1,7 @@
 ---
 Name: elemental
 ---
-
+# test comment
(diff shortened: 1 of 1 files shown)


brynwhyman:
FYI you can check the status of supported module builds here: https://maxime-rainville.github.io/travis-dashboard/

And the current major branch (`4`) is on a cron to run a build weekly, as are all other modules :) 

FYI the fix for the failing build is being tracked here: https://github.com/silverstripe/silverstripe-elemental/issues/887

Issue #887, opened 2021-03-17
Broken build - Various issues


emteknetnz:
This affects `4.5` and `4`

https://travis-ci.com/github/silverstripe/silverstripe-elemental/jobs/497628625
- (graphql3 and graphql4) tests/Behat/features/archive-block-element.feature:23

----

The following two failures only affect the `4` branch, not `4.5`

GraphQL 3 installed:
https://travis-ci.com/github/silverstripe/silverstripe-elemental/jobs/497674380
- DNADesign\Elemental\Tests\GraphQL\DuplicateElementMutationTest::testResolvePermissions
Undefined index: id

GraphQL 4 installed:
https://travis-ci.com/github/silverstripe/silverstripe-elemental/jobs/497674381
- DNADesign\Elemental\Tests\GraphQL\DuplicateElementMutationTest::testResolvePermissions
Error: Class 'DNADesign\Elemental\GraphQL\DuplicateElementMutation' not found

### PRs
- [x] https://github.com/silverstripe/silverstripe-elemental/pull/892
- [x] https://github.com/silverstripe/silverstripe-elemental/pull/893

PR #892, opened 2021-04-14
MNT Update behat test to click archive button

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-elemental/issues/887

### DIFF
--- a/tests/Behat/features/archive-block-element.feature
+++ b/tests/Behat/features/archive-block-element.feature
@@ -24,7 +24,8 @@ Feature: Archive elements in the CMS
     Given I see a list of blocks
     When I press the "View actions" button
       Then I should see the archive button for block 1
-    When I press the "Archive" button
+    # Use a css selector instead of the "Archive" button otherwise it will get confused with the Archive ModelAdmin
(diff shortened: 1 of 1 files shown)


tests/Behat/features/archive-block-element.feature
```diff
@@ -24,7 +24,8 @@ Feature: Archive elements in the CMS
     Given I see a list of blocks
     When I press the "View actions" button
       Then I should see the archive button for block 1
-    When I press the "Archive" button
+    # Use a css selector instead of the "Archive" button otherwise it will get confused with the Archive ModelAdmin
+    When I click on the ".element-editor__actions-archive.dropdown-item" element
```
maxime-rainville, line 28: Just for context, `I press the "Archive" button` is kind of dumb because it doesn't distinguish between buttons and link.

Fixing it in behat-testing could be difficult difficult because it could potentially break a bunch of other things in all the other module. And it's not such a problem since it's relatively uncommon to have a link and a button on the same page with the same label.

maxime-rainville (approved):
Looks good and all tests are green.

PR #896, opened 2021-04-29
DEP Allow graphql ^4

emteknetnz:
Issue https://github.com/silverstripeltd/product-issues/issues/402

Behat failures look to be unrelated - this different open pull-request has the same failures https://github.com/silverstripe/silverstripe-elemental/pull/897

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "silverstripe/admin": "^1.7@dev",
         "silverstripe/versioned": "^1.7@dev",
         "silverstripe/versioned-admin": "^1.7@dev",
-        "silverstripe/graphql": "^3.4 || 4.x-dev",
+        "silverstripe/graphql": "^3.4 || ^4",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -21,7 +21,7 @@
         "silverstripe/admin": "^1.7@dev",
         "silverstripe/versioned": "^1.7@dev",
         "silverstripe/versioned-admin": "^1.7@dev",
-        "silverstripe/graphql": "^3.4 || 4.x-dev",
+        "silverstripe/graphql": "^3.4 || ^4 || 4.x-dev",
```
maxime-rainville, line 24: Not sure what `|| 4.x-dev` is meant to add here ... `4.x-dev` should match the `^4` constraint.

```suggestion
        "silverstripe/graphql": "^3.4 || ^4",
```

emteknetnz: True that, have updated

PR #897, opened 2021-04-30
FIX `getElementsForSearch()` to replace tags with spaces

andrewandante:
Having an issue where this method combines `<p>` tags into a single string, muddying my search results, i.e.

```html
<p>banana</p>
<p>hamburger</p>
```

becomes `bananahamburger` when this method is called.

Starting with a failing test - going to propose a fix which should replace these with spaces

### DIFF
--- a/src/Extensions/ElementalPageExtension.php
+++ b/src/Extensions/ElementalPageExtension.php
@@ -43,7 +43,8 @@ class ElementalPageExtension extends ElementalAreasExtension
                 /** @var ElementalArea $area */
                 $area = $this->owner->$key();
                 if ($area) {
-                    $output[] = strip_tags($area->forTemplate());
+                    // Replace HTML tags with spaces
+                    $output[] = strip_tags(str_replace('<', ' <', $area->forTemplate()));
                 }
             }
         } finally {
--- a/tests/ElementalPageExtensionTest.php
+++ b/tests/ElementalPageExtensionTest.php
@@ -97,4 +97,21 @@ class ElementalPageExtensionTest extends FunctionalTest
             'Duplicated page has duplicated area and duplicated elements, i.e. not shared'
         );
     }
+
+    public function testGetElementsForSearch()
+    {
+        /** @var TestPage $page */
+        $page = $this->objFromFixture(TestPage::class, 'page_with_html_elements');
+        $output = $page->getElementsForSearch();
+        $this->assertNotEmpty($output);
+
+        // Confirm tags have been stripped
+        $this->assertNotContains('<p>', $output);
+        $this->assertNotContains('</p>', $output);
+
+        // Confirm paragraphs don't get smushed together, also across elements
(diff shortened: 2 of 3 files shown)


src/Extensions/ElementalPageExtension.php
```diff
@@ -43,7 +43,8 @@ public function getElementsForSearch()
                 /** @var ElementalArea $area */
                 $area = $this->owner->$key();
                 if ($area) {
-                    $output[] = strip_tags($area->forTemplate());
+                    // Replace HTML tags with spaces
+                    $output[] = strip_tags(str_replace('<', ' <', $area->forTemplate()));
```
adrhumphreys, line 47: I think this presents a trade off that's worth it in this instance, as now if you have the string `There are <5 potatoes` it will become `There are < 5 potatoes` which means content previously indexable by `ElementSiteTreeFilterSearch::applyDefaultFilters` will no longer be picked up. But I think this is enough of an edge case that this is a good update :heart:

adrhumphreys (changes requested):
Can you check the tests? Other than those ones failing this is good and I'd be keen to get it merged in 👍🏽

src/Extensions/ElementalPageExtension.php
michalkleiner: It's easy to target HTML tags via regex, not sure what is the benefit of not doing that here, while possibly changing the meaning of the indexed content (marginally, depending on the settings, but can't say it's definitely not).

andrewandante: The idea is that we fall back to `strip_tags()` to do the heavy lifting in terms of HTML detection.

There's a lot of edge-cases either way really, if we hand-roll our own regex we are just as likely to add another issue, like "we wanted to make sure there were < 5 errors, but > 10 comments" being caught, or innerhtml such as:

```html
<button onclick="document.getElementById('alert').innerHTML='<strong>MESSAGE</strong>';">
```

Not being rendered properly. I admit that's unlikely but it's possible.

@adrhumphreys it will become `There are   <5 potatoes` and shouldn't break anything, the space is added before the `<` - I'll take a look at those tests, not sure exactly why they are failing, it _seems_ unrelated to my changes

andrewandante: Suspicion that the Behat failures are related to chromedriver 91? Having trouble replicating locally

emteknetnz:
@andrewandante I've rerun the tests.  Looks like there's a failure related to phpunit9 - https://app.travis-ci.com/github/silverstripe/silverstripe-elemental/jobs/555927169

andrewandante:
@emteknetnz PHPUnit tests updated, still seeing Behat failures though. I see that they are also failing on `4`, so can we assume they are unrelated?

emteknetnz:
Yes unrelated

emteknetnz (approved):

PR #909, opened 2021-06-17
Add config to disable the content search in ElementSiteTreeFilterSearch

alex-dna:
We have found that with large website, with hundreds of elemental page, the CMS Page search no longer works as every page gets rendered, which is too time consuming.
This allows to disable the content search via a config.


kinglozzer (approved):
LGTM, we’ve had the same issue on large sites

emteknetnz (commented):
Tested locally, works good

Issue #186, opened 2018-02-11
Add compatibility for adding link to "Anchor on a page" in WYSIWYG editors


robbieaverill:
![image](https://user-images.githubusercontent.com/5170590/36079567-34aa58d6-0fe9-11e8-8166-66fe7baa1b6c.png)

This functionality in the CMS lets you add a link to an anchor on an arbitrary page. With elemental enabled on a page this won't automatically populate a list of available anchors, since it'll be reading and parsing the page's Content field.

If we can augment the search location for HTML then we can tie it into the method that returns an entire page's elemental content for the sake of search indexing (see #163).

From a data point of view we definitely want to avoid duplicating all element data into the page's Content field.

Note that currently a WYSIWYG inside an elemental block is able to link to anchors within itself, but not within other elemental blocks on the same or on other blocks. On save, these links get marked as broken, even though they work.

## Acceptance Criteria
- [x] Using the "anchor on a page" link in a WYSIWYG field (either directly on a page or inside an elemental block), I can link to anchors inside elemental blocks on _a different_ page
- [x] Using the "anchor on a page" link in a WYSIWYG field, I can link to anchors inside the same elemental block I am editing
- [x] Using the "anchor on a page" link in a WYSIWYG field, I can link to anchors inside _other_ elemental blocks on _the same_ page
- [x] Links to anchors inside elemental blocks are not marked as broken
- [x] Links to anchors that do not exist on the page or in any elemental blocks _are_ marked as broken

## PRs
- #911 
- https://github.com/silverstripe/silverstripe-cms/pull/2741
- https://github.com/silverstripe/silverstripe-frameworktest/pull/106
- https://github.com/silverstripe/silverstripe-cms/pull/2742

brynwhyman:
Thinking a higher level API would be required for this. There's a number of modules that would also be affected by this too: `externallinks`, `fulltextsearch`, maybe more?

MichelleZZZZ:
Hi @robbieaverill 

I have a similar issue with Anchor on the Elemental block(not inline form). And I would like to ask if the fix might happen at some stage?

Cheers

robbieaverill:
Hi @MichelleZZZZ, I'm not sure what the roadmap looks like but @brynwhyman may be able to help

GuySartorelli:
Looks like this would be fixed by #911

PR #911, opened 2021-06-29
NEW Get anchor links from elemental pages

emteknetnz:
Adds anchor links in `HTMLText` fields on an elemental block to the "anchor on a page" link form in WYSIWYG fields.
The new method should be overridden for blocks that need to add additional anchors or exclude some anchors from this base list.

An argument can be made that this should render the block and use that to parse out any anchors - but the approach this PR takes is more in line with how this is currently done in `SiteTree::getAnchorsOnPage()`.

Also updated the `theHasAContentElementWithContent` fixture context regex to allow arbitrary html content in the test fixture.

## Dependencies
The following related PRs must be merged in first in order for tests to pass
- https://github.com/silverstripe/silverstripe-frameworktest/pull/106
- https://github.com/silverstripe/silverstripe-cms/pull/2741

## Parent Issues
- https://github.com/silverstripe/silverstripe-elemental/issues/186

### DIFF
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -257,7 +257,7 @@ class BaseElement extends DataObject implements CMSPreviewable
     public function canDelete($member = null)
     {
         $extended = $this->extendedCan(__FUNCTION__, $member);
-        
+
         if ($extended !== null) {
             return $extended;
         }
@@ -680,6 +680,38 @@ JS
     }
 
     /**
+     * Get anchors in this block's content.
+     * Used to populate the "anchor on a page" link in the WYSIWYG
+     *
+     * By default, this finds anchors in any HTMLText field on the block, but
+     * this method should be overridden if anchors are provided in other ways
+     * for this block or if not all HTMLText fields for this block are
+     * displayed on the front-end.
+     */
+    public function getAnchorsInContent(): array
+    {
+        $anchors = [$this->getAnchor()];
+        $anchorRegex = "/\\s+(name|id)\\s*=\\s*([\"'])([^\\2\\s>]*?)\\2|\\s+(name|id)\\s*=\\s*([^\"']+)[\\s +>]/im";
+        $allFields = DataObject::getSchema()->fieldSpecs($this);
+        foreach ($allFields as $field => $fieldSpec) {
+            $fieldObj = $this->owner->dbObject($field);
+            if ($fieldObj instanceof DBHTMLText) {
+                $parseSuccess = preg_match_all($anchorRegex, $fieldObj->getValue() ?? '', $matches);
+                if ($parseSuccess >= 1) {
+                    $fieldAnchors = array_values(array_filter(
+                        array_merge($matches[3], $matches[5])
+                    ));
+                    $anchors = array_merge($anchors, $fieldAnchors);
+                }
+            }
+        }
+        $anchors = array_unique($anchors);
+
+        $this->extend('updateAnchorsInContent', $anchors);
+        return $anchors;
+    }
+
+    /**
      * @param string|null $action
      * @return string|null
      * @throws \Psr\Container\NotFoundExceptionInterface
@@ -688,7 +720,7 @@ JS
     public function AbsoluteLink($action = null)
     {
         $page = $this->getPage();
-        
+
         if ($page && ClassInfo::hasMethod($page, 'AbsoluteLink')) {
             $link = $page->AbsoluteLink($action) . '#' . $this->getAnchor();
             $this->extend('updateAbsoluteLink', $link);
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
     },
     "require-dev": {
       "silverstripe/recipe-testing": "^2",
-      "silverstripe/frameworktest": "4.x-dev"
+      "silverstripe/frameworktest": "^0.4.5"
     },
     "suggest": {
         "silverstripe/elemental-blocks": "Adds a set of common SilverStripe content block types",
--- a/src/Extensions/ElementalPageExtension.php
+++ b/src/Extensions/ElementalPageExtension.php
@@ -2,7 +2,9 @@
 
 namespace DNADesign\Elemental\Extensions;
 
+use DNADesign\Elemental\Models\BaseElement;
 use DNADesign\Elemental\Models\ElementalArea;
+use SilverStripe\CMS\Model\SiteTree;
 use SilverStripe\Control\Controller;
 use SilverStripe\View\Parsers\HTML4Value;
 use SilverStripe\View\SSViewer;
@@ -47,23 +49,14 @@ class ElementalPageExtension extends ElementalAreasExtension
         SSViewer::set_themes(SSViewer::config()->get('themes'));
         try {
             $output = [];
-            foreach ($this->owner->hasOne() as $key => $class) {
-                if ($class !== ElementalArea::class) {
-                    continue;
-                }
-                /** @var ElementalArea $area */
-                $area = $this->owner->$key();
-                if ($area) {
-                    foreach ($area->Elements() as $element) {
-                        if ($element->getSearchIndexable()) {
-                            $content = $element->getContentForSearchIndex();
-                            if ($content) {
-                                $output[] = $content;
-                            }
-                        }
+            $this->loopThroughElements(function (BaseElement $element) use (&$output) {
+                if ($element->getSearchIndexable()) {
+                    $content = $element->getContentForSearchIndex();
+                    if ($content) {
+                        $output[] = $content;
                     }
                 }
-            }
+            });
         } finally {
             // Reset theme if an exception occurs, if you don't have a
             // try / finally around code that might throw an Exception,
@@ -71,6 +64,19 @@ class ElementalPageExtension extends ElementalAreasExtension
             SSViewer::set_themes($oldThemes);
         }
         return implode($this->owner->config()->get('search_index_element_delimiter') ?? '', $output);
+    }
+
+    /**
+     * @see SiteTree::getAnchorsOnPage()
+     */
+    public function updateAnchorsOnPage(array &$anchors): void
+    {
+        if (!($this->owner instanceof SiteTree)) {
+            return;
+        }
+        $this->loopThroughElements(function (BaseElement $element) use (&$anchors) {
+            $anchors = array_merge($anchors, $element->getAnchorsInContent());
+        });
     }
 
     public function MetaTags(&$tags)
@@ -89,6 +95,25 @@ class ElementalPageExtension extends ElementalAreasExtension
                 $body->removeChild($tag);
             }
             $tags = $html->getContent();
+        }
+    }
+
+    /**
+     * Call some function over all elements belonging to this page
+     */
+    private function loopThroughElements(callable $callback): void
+    {
+        foreach ($this->owner->hasOne() as $key => $class) {
+            if ($class !== ElementalArea::class) {
+                continue;
+            }
+            /** @var ElementalArea $area */
+            $area = $this->owner->$key();
+            if ($area) {
+                foreach ($area->Elements() as $element) {
+                    $callback($element);
+                }
+            }
         }
     }
 }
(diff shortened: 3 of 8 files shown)


GuySartorelli:
I've taken over this PR. Marking this as draft so nobody merges it just yet - the functionality is all there but I still need to add test coverage.

sabina-talipova:
I'm testing these changes and I can't seem to get the result (the list of anchors on the page) in any way. Perhaps I'm doing something wrong. Here are the steps I follow.
1. I create pages using the following template.
```php"
<?php

namespace {

    use SilverStripe\Forms\HTMLEditor\HTMLEditorField;

    class Page extends SiteTree
    {
        private static $db = [
            'Summary'     => 'HTMLText',
        ];
        private static $has_one = [
            'BlocksMain'         => ElementalArea::class,
            'BlocksRight'        => ElementalArea::class,
            'BlocksFooter'       => ElementalArea::class,
        ];

        {
    
            $summary = HtmlEditorField::create('Summary', 'Summary', false);
            $summary->setRows(5);
            $summary->setDescription('Test Description');

            $fields->addFieldsToTab('Root.Main',$summary, 'Summary');
            return $fields;
        }
    }
}
```
2. In each ElementalArea I add a Content block.
3. In the Content block I add content ` <p><a id="my_link" name="my_link">My_link</a>Test</p><p id="my_text">Tests2</p>` via Source code.
4. Save and publish ElementalArea.
5. In the Content Editor for the Summary, I click on "Insert Link" and select "Anchor on a page".
6. In the popup window that opens, in the "Select a page" I select this page with which I am working at the moment.
7. But then In the "Anchor" drop-down list, I see "No results found".

I also tried adding an anchor, which is placed in the Summary, to the ElementalArea Content Block. And I even refreshed the page. Still don't have the anchors.

GuySartorelli:
@sabina-talipova Note in the PR functionality added to the `ElementalPageExtension` - it is my expectation that people will be using that extension when adding elemental blocks to their pages. Perhaps this PR should also add some documentation that advises people who (for whatever reason) don't use that extension to also add this new functionality to their pages directly?

GuySartorelli:
@sabina-talipova
Can you please try following these steps and let me know if that works?

1. Create a new page with elemental blocks
2. On the new page, create a content elemental block
3. Add some anchor link to the content elemental block.
4. **Save the page**
5. In that, or any other block (including on a different page), click "Insert Link", then "Anchor on a page" and select the above page.

The anchor link you added should be in the anchors dropdown list.

I _suspect_ the issue you were having is that new anchors _when you don't save the page_ aren't showing up - that's a defect with the way I implemented #2741 which I will need to fix - but in the meantime the main functionality here should work.

sabina-talipova:
@GuySartorelli , it still doesn't work and it looks like it doesn't invoke updateAnchorsOnPage method from ElementalPageExtension. ElementalPageExtension adds ElementalArea on the Page, but doesn't run updateAnchorsOnPage. Probably, I should do something else.

GuySartorelli:
As discussed on slack, that sounds like a problem with the way I've changed the javascript in https://github.com/silverstripe/silverstripe-cms/pull/2741, which means I'm not quite done yet. Once you've finished your review, assign this back to me and I'll see what I can do about that.

sabina-talipova:
@GuySartorelli , 

I've finished testing in my local environment. Nice work! All works perfectly fine even with changes that I merged for `TinyMCE_sslink-anchor.js`. BUT with a few tiny enchincement that should be done.

First of all, in `SiteTree` class in the method `getAnchorsOnPage()` would be nice to implement your huck for DBHTMLText field. Since currently in this method we are looking for `$this->Content` (https://github.com/silverstripe/silverstripe-cms/blob/77b5bfa497d5026ffc806f1071a6c5d04d65bfb4/code/Model/SiteTree.php#L3338). But if I have other field's name , for example Summary, we skip it. That was the reason for my issue, I hadn't Content field at all. So if you implement your changes, something like this, it helps us to solve problem where we have more then one DBHTMLText on the page with different names.

```php
        $anchors = [];
        $allFields = DataObject::getSchema()->fieldSpecs($this);
        $anchorRegex = "/\\s+(?<attr>name|id)\\s*=\\s*([\"'])(?<token>[^\\2\\s>]*?)\\2|\\s+(name|id)\\s*=\\s*([^\"']+)[\\s +>]/im";

        foreach ($allFields as $field => $fieldSpec) {
            $fieldObj = $this->owner->dbObject($field);
            if ($fieldObj instanceof DBHTMLText) {
                $parseSuccess = preg_match_all($anchorRegex, $fieldObj->getValue() ?? '', $matches);
                if ($parseSuccess >= 1) {
                    $fieldAnchors = array_values(array_filter(
                        array_merge($matches['token'])
                    ));
                    $anchors = array_merge($anchors, $fieldAnchors);
                }
            }
        }
        $anchors = array_unique($anchors);

        $this->extend('updateAnchorsOnPage', $anchors);

        return $anchors;
```

But it is just suggestion.

sabina-talipova:
@GuySartorelli , 

I've finished testing in my local environment. Nice work! All works perfectly fine even with changes that I merged for `TinyMCE_sslink-anchor.js`. BUT with a few tiny enchincement that should be done.

First of all, in `SiteTree` class in the method `getAnchorsOnPage()` would be nice to implement your huck for DBHTMLText field. Since currently in this method we are looking for `$this->Content` (https://github.com/silverstripe/silverstripe-cms/blob/77b5bfa497d5026ffc806f1071a6c5d04d65bfb4/code/Model/SiteTree.php#L3338). But if I have other field's name , for example Summary, we skip it. That was the reason for my issue, I hadn't Content field at all. So if you implement your changes, something like this, it helps us to solve problem where we have more then one DBHTMLText on the page with different names.

```php
        $anchors = [];
        $allFields = DataObject::getSchema()->fieldSpecs($this);
        $anchorRegex = "/\\s+(?<attr>name|id)\\s*=\\s*([\"'])(?<token>[^\\2\\s>]*?)\\2|\\s+(name|id)\\s*=\\s*([^\"']+)[\\s +>]/im";

        foreach ($allFields as $field => $fieldSpec) {
            $fieldObj = $this->owner->dbObject($field);
            if ($fieldObj instanceof DBHTMLText) {
                $parseSuccess = preg_match_all($anchorRegex, $fieldObj->getValue() ?? '', $matches);
                if ($parseSuccess >= 1) {
                    $fieldAnchors = array_values(array_filter(
                        array_merge($matches['token'])
                    ));
                    $anchors = array_merge($anchors, $fieldAnchors);
                }
            }
        }
        $anchors = array_unique($anchors);

        $this->extend('updateAnchorsOnPage', $anchors);

        return $anchors;
```

But it is just suggestion.

src/Models/BaseElement.php
```diff
@@ -679,6 +679,38 @@ public function getAnchor()
...
+     *
+     * By default, this finds anchors in any HTMLText field on the block, but
+     * this method should be overridden if anchors are provided in other ways
+     * for this block or if not all HTMLText fields for this block are
+     * displayed on the front-end.
+     */
+    public function getAnchorsInContent(): array
+    {
+        $anchors = [$this->getAnchor()];
+        $anchorRegex = "/\\s+(name|id)\\s*=\\s*([\"'])([^\\2\\s>]*?)\\2|\\s+(name|id)\\s*=\\s*([^\"']+)[\\s +>]/im";
```
sabina-talipova, line 694: I used https://regex101.com/ for testing regular expression and I found that second part of this condition doesn't make sence. It looks like it searches only for quotes. I would suggest to remove this part or modify it for some special purpose.

sabina-talipova, line 694: Also would be nice to modify first part to provide ability to search name or id even if they have a spaces before and after.
`\s+(name|id)\s*?=\s*?(\s*?[\"']\s*?)(\s*?[^\2\s>]*?\s*?)\2`

src/Models/BaseElement.php
```diff
@@ -679,6 +679,38 @@ public function getAnchor()
...
+        $allFields = DataObject::getSchema()->fieldSpecs($this);
+        foreach ($allFields as $field => $fieldSpec) {
+            $fieldObj = $this->owner->dbObject($field);
+            if ($fieldObj instanceof DBHTMLText) {
+                $parseSuccess = preg_match_all($anchorRegex, $fieldObj->getValue() ?? '', $matches);
+                if ($parseSuccess >= 1) {
+                    $fieldAnchors = array_values(array_filter(
+                        array_merge($matches[3], $matches[5])
```
sabina-talipova, line 702: I would suggest to add name for the group in regular expression, like `\s+(name|id)\s*?=\s*?(\s*?[\"']\s*?)(?<token>\s*?[^\2\s>]*?\s*?)\2` and replace $matches[3] to $match['token']. Also, remove $matches[5], because the 5th group returns only quotes.

src/Models/BaseElement.php
GuySartorelli: Can you please specify which part you mean by "the second part"?
I took this regex straight from [`SiteTree::getAnchorsOnPage()`](https://github.com/silverstripe/silverstripe-cms/blob/77b5bfa497d5026ffc806f1071a6c5d04d65bfb4/code/Model/SiteTree.php#L3337) so if it's flawed we'll need to update it there too.

src/Models/BaseElement.php
GuySartorelli: The fifth group actually returns anything _other than_ quotes. `([^"']+)` The `^` there turns it into "anything other than".

I don't think adding named groups would be beneficial here - both of the two groups we are interested in just contain anchor names, so the two groups would have the same name which for obvious reasons won't work.

src/Models/BaseElement.php
sabina-talipova: How I can see we have two conditions:
It is looking for either `\s+(name|id)\s*=\s*([\"'])([^\2\s>]*?)\2`(first part)  or  `\s+(name|id)\s*=\s*([^\"']+)[\s +>]`(second part)
If you copy second part and test it in https://regex101.com/, use this string for test `<div><p id="my_p_link">My paragraph</p></div>`, it cannot find any matches. Probably, this regexpression should test some other sort of string.

src/Models/BaseElement.php
sabina-talipova: `([^"']+)` -  Sorry, my fault.
But anyway, it looks like it should find string something like this `<p name=my_p_link_2>My paragraph</p><p`, but not id / name title

src/Models/BaseElement.php
GuySartorelli: Right. So that second section is looking at markup where the quotes have been missed out, e.g. `<a name=anchor-name>`. this was taken from the existing `SiteTree::getAnchorsOnPage()` so I feel like omitting or changing it is likely to cause regressions.

src/Models/BaseElement.php
GuySartorelli: Yes, exactly. So in that example `my_p_link_2` would be the anchor.

src/Models/BaseElement.php
sabina-talipova: Don't worry. It was just a suggestion. Probably, we should just do more tests and enchance it.

GuySartorelli:
>  in `SiteTree` class in the method `getAnchorsOnPage()` would be nice to implement your solution for DBHTMLText field.

I won't make this change as a part of fixing the issue here. It could be a good enhancement so feel free to an issue for it in silverstripe/silverstripe-cms - but that is a separate problem to the issue this PR is fixing. That problem will exist for situations where people already have fields such as `Summary` even without elemental installed.

src/Models/BaseElement.php
sabina-talipova: I'll test second part of this condition and if it will be needed I'll create new ticket for enchancement.

sabina-talipova (approved):
All works fine in local environment together with changes from PR https://github.com/silverstripe/silverstripe-cms/pull/2742
Merge when all related PRs will be approved.

micschk:
Note the duplicated anchor-in-page regex [seems to pick up the "id" part of image-shortcodes](https://github.com/silverstripe/silverstripe-cms/issues/2837) (and possibly other shortcodes) as well.

PR #912, opened 2021-07-01
BUGFIX: Ensure graphql3 compat in type name generation

unclecheese:
See https://github.com/silverstripe/silverstripe-elemental/pull/870#issuecomment-872611069

### DIFF
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -1061,6 +1061,8 @@ JS
      */
     public static function getGraphQLTypeName(): string
     {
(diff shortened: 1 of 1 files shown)


tractorcow:
I have tested this and it works fine.

kinglozzer:
Looks legit, should we target this at `4.6` instead of `4`? Looks like a candidate for a quick patch release

unclecheese:
Yeah good point. 4.6.

unclecheese:
Have rebased against 4.6. Will merge on green.

PR #913, opened 2021-07-07
ENH: Provide more control over elemental block indexing.

GuySartorelli:
Provides a new configuration variable to exclude specific elemental block classes from being indexed in search.
Also provides a new extension point for modifying exactly what gets indexed for each block.

### DIFF
--- a/src/Extensions/ElementalPageExtension.php
+++ b/src/Extensions/ElementalPageExtension.php
@@ -26,6 +26,17 @@ class ElementalPageExtension extends ElementalAreasExtension
     ];
 
     /**
+     * The delimiter to separate distinct elements in indexed content.
+     *
+     * When using the getElementsForSearch() method to index all elements in a single field,
+     * a custom delimiter can be used help to avoid false positive results for phrase queries.
+     *
+     * @config
+     * @var string
+     */
+    private static $search_index_element_delimiter = ' ';
+
+    /**
      * Returns the contents of each ElementalArea has_one's markup for use in Solr or Elastic search indexing
      *
      * @return string
@@ -43,8 +54,14 @@ class ElementalPageExtension extends ElementalAreasExtension
                 /** @var ElementalArea $area */
                 $area = $this->owner->$key();
                 if ($area) {
-                    // Replace HTML tags with spaces
-                    $output[] = strip_tags(str_replace('<', ' <', $area->forTemplate()));
+                    foreach ($area->Elements() as $element) {
+                        if ($element->getSearchIndexable()) {
+                            $content = $element->getContentForSearchIndex();
+                            if ($content) {
+                                $output[] = $content;
+                            }
+                        }
+                    }
                 }
             }
         } finally {
@@ -53,7 +70,7 @@ class ElementalPageExtension extends ElementalAreasExtension
             // CMS layout can break on the response. (SilverStripe 4.1.1)
             SSViewer::set_themes($oldThemes);
         }
-        return implode($output);
+        return implode($this->owner->config()->get('search_index_element_delimiter'), $output);
     }
 
     public function MetaTags(&$tags)
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -175,6 +175,15 @@ class BaseElement extends DataObject implements CMSPreviewable
     private static $displays_title_in_template = true;
 
     /**
+     * Determines whether a block should be indexable in search.
+     *
+     * @config
+     * @var boolean
+     * @see ElementalPageExtension::getElementsForSearch()
+     */
+    private static $search_indexable = true;
+
+    /**
      * Store used anchor names, this is to avoid title clashes
      * when calling 'getAnchor'
      *
@@ -474,6 +483,33 @@ JS
     public function Top()
     {
         return (Controller::has_curr()) ? Controller::curr() : null;
+    }
+
+    /**
+     * Determines whether this elemental block is indexable in search.
+     *
+     * By default, this uses the configurable variable search_indexable, but
+     * this method can be overridden to provide more complex logic if required.
+     *
+     * @return boolean
+     */
+    public function getSearchIndexable(): bool
+    {
+        return (bool) $this->config()->get('search_indexable');
+    }
+
+    /**
+     * Provides content to be indexed in search.
+     *
+     * @return string
+     */
+    public function getContentForSearchIndex(): string
+    {
+        // Strips tags but be sure there's a space between words.
+        $content = trim(strip_tags(str_replace('<', ' <', $this->forTemplate())));
+        // Allow projects to update indexable content of third-party elements.
+        $this->extend('updateContentForSearchIndex', $content);
+        return $content;
     }
 
     /**
--- a/docs/en/searching-blocks.md
+++ b/docs/en/searching-blocks.md
@@ -30,3 +30,26 @@ SilverStripe\FullTextSearch\Search\FullTextSearch:
   indexes:
     - MyCustomIndex
 ```
+
+You can define whether each block is included in your search index using the
+`search_indexable` configuration variable, which is `true` by default:
+
+```yml
+App\Models\MyCustomElementalBlock:
+  search_indexable: false
+```
+
+You can also customise the content that is indexed for your blocks. By default
+the block is rendered in full using the templating engine, and the resultant
+markup is included in the index. You can override the `getContentForSearchIndex`
+method on your elemental blocks to change that. This is useful, for example, if
+your templates include hardcoded text or references to other content you don't
+want to be indexed.
+
+If you want to use a specific delimiter between each block, that can be configured
+as well. The default is a space, but you might for example want to use an ellipses
+to make it clear in search results where one piece of content ends and another begins.
+```yml
+Page:
+  search_index_element_delimiter: ' ... '
+```
--- a/tests/BaseElementTest.php
+++ b/tests/BaseElementTest.php
@@ -7,6 +7,7 @@ use DNADesign\Elemental\Extensions\ElementalPageExtension;
 use DNADesign\Elemental\Models\BaseElement;
 use DNADesign\Elemental\Models\ElementalArea;
 use DNADesign\Elemental\Models\ElementContent;
+use DNADesign\Elemental\Tests\Src\TestContentForSearchIndexExtension;
 use DNADesign\Elemental\Tests\Src\TestElement;
 use DNADesign\Elemental\Tests\Src\TestPage;
 use Page;
@@ -235,5 +236,22 @@ class BaseElementTest extends FunctionalTest
         $element1->write();
 
         $this->assertEquals(0, (int) $element1->Sort);
+    }
+
+    public function testGetContentForSearchIndex()
+    {
+        $element = $this->objFromFixture(ElementContent::class, 'content4');
+        // Content should have tags stripped with a space before what were the < characters
+        // One closing tag plus one opening tag means there should be two spaced between paragraphs
+        $this->assertEquals('One paragraph  And another one', $element->getContentForSearchIndex());
+    }
+
+    public function testUpdateContentForSearchIndex()
+    {
+        ElementContent::add_extension(TestContentForSearchIndexExtension::class);
+        $element = $this->objFromFixture(ElementContent::class, 'content4');
+        // Content should be updated by the extension
+        $this->assertEquals('This is the updated content.', $element->getContentForSearchIndex());
+        ElementContent::remove_extension(TestContentForSearchIndexExtension::class);
     }
 }
--- a/tests/ElementalPageExtensionTest.php
+++ b/tests/ElementalPageExtensionTest.php
@@ -114,4 +114,19 @@ class ElementalPageExtensionTest extends FunctionalTest
         $this->assertStringNotContainsString('oneMore', $output);
         $this->assertStringNotContainsString('paragraphsAnd', $output);
     }
+
+    public function testSearchIndexElementDelimiter()
+    {
+        /** @var TestPage $page */
+        $page = $this->objFromFixture(TestPage::class, 'page_with_html_elements');
+
+        // Confirm default delimiter of a single space is applied between elements
+        $output = $page->getElementsForSearch();
+        $this->assertStringContainsString('another one More paragraphs', $output);
+
+        // Confirm configured delimiter is applied between elements
+        Config::modify()->set(TestPage::class, 'search_index_element_delimiter', ' ... ');
+        $output = $page->getElementsForSearch();
+        $this->assertStringContainsString('another one ... More paragraphs', $output);
+    }
 }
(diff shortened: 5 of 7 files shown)


src/Extensions/ElementalPageExtension.php
```diff
@@ -43,7 +49,16 @@ public function getElementsForSearch()
                 /** @var ElementalArea $area */
                 $area = $this->owner->$key();
                 if ($area) {
-                    $output[] = strip_tags($area->forTemplate());
+                    // Only render the elemental blocks which are not excluded from search and don't pull from a page.
+                    foreach ($area->Elements() as $element) {
+                        if (!in_array($element->ClassName, Config::inst()->get(static::class, 'elements_excluded_from_search'))) {
+                            $forIndexing = $element->forTemplate();
```
michalkleiner, line 55: This always renders the element and then it decides whether it will throw it away based on the extension point.
Could it be reworked so excluded elements don't even render in the first place?

I can imagine something like adding `public function getIsIndexable(): bool` to the BaseElement and then each element can provide custom logic that would be used here to skip the rendering.

GuySartorelli: >This always renders the element and then it decides whether it will throw it away based on the extension point.

The intention was that it excludes it if it's in the `elements_excluded_from_search` array, but then on a block-by-block basis, developers have even more control. So it might be a block which isn't excluded by class, but perhaps there is a rule where if some fields are not filled in it shouldn't be displayed (or indexed). So in that case the developer might choose to not index that specific block.
They can also choose to just modify what is indexed, in case the display of the block is not the same as what should be indexed for the block (I have at least one block where some fields are pulled from other sources, so I wouldn't want to index those fields as a part of this block).
If on the other hand the block's class is in the `elements_excluded_from_search` array, that if statement will return false and it will automatically be excluded.

I like the idea of a `getIsIndexable` method in principle provided it has an extension point in it to allow modifying that value (I use a lot of third-party blocks and want to define what is indexable for my project). But I think being able to modify _what_ is indexed per block is also important.

michalkleiner: Fair point about the modification, but basically at that point you're dealing with rendered markup, not structured data. So if you wanted to avoid e.g. an API call, that would be too late anyway with the above code.

There's an epic somewhere to provide a better content API that would allow for that granularity.

GuySartorelli:
Not sure when I'll get a chance to update this to be in line with @michalkleiner's recommendations - setting to draft until then.

src/Extensions/ElementalPageExtension.php
GuySartorelli: I've made changes to allow per-block configuration both for whether a block is indexable in the first place, and if so, what content exactly will be indexed.
I think this is a much more sensible way of handling it. Please review.

GuySartorelli:
Tests are failing, but I don't see how the changes I've made could have caused that.... can someone who understands the `ElementSiteTreeFilterSearchTest` please take a look?

src/Models/BaseElement.php
```diff
@@ -457,6 +466,33 @@ public function Top()
...
+     * Provides content to be indexed in search.
+     *
+     * @return string|null HTML
+     */
+    public function forSearchIndex()
+    {
+        // Strips tags but be sure there's a space between words.
+        $index = trim(strip_tags(str_replace('<', ' <', $this->forTemplate())));
+        // Allow projects to update indexable content of third-party elements.
+        $this->extend('updateSearchIndex', $index);
```
michalkleiner, line 492:
```suggestion
        $this->extend('updateContentForSearchIndex', $content);
```

src/Models/BaseElement.php
```diff
@@ -457,6 +466,33 @@ public function Top()
...
+    {
+        return (bool)$this->config()->get('is_indexable');
+    }
+
+    /**
+     * Provides content to be indexed in search.
+     *
+     * @return string|null HTML
+     */
+    public function forSearchIndex()
```
michalkleiner, line 487:
```suggestion
    public function getContentForSearchIndex()
```

src/Models/BaseElement.php
```diff
@@ -457,6 +466,33 @@ public function Top()
...
+     * Determines whether this elemental block is indexable in search.
+     *
+     * By default, this uses the configurable variable is_indexable, but
+     * this method can be overridden to provide more complex logic if required.
+     *
+     * @return boolean
+     */
+    public function getIsIndexable(): bool
+    {
+        return (bool)$this->config()->get('is_indexable');
```
michalkleiner, line 479: Should this use the uninherited config so that inheritance wouldn't play role in excluding certain class from indexing?

GuySartorelli: Ooh yes, that's a good idea. I'll make that adjustment.

src/Extensions/ElementalPageExtension.php
```diff
@@ -52,7 +59,7 @@ public function getElementsForSearch()
             // CMS layout can break on the response. (SilverStripe 4.1.1)
             SSViewer::set_themes($oldThemes);
         }
-        return implode($output);
+        return implode(' ... ', $output);
```
michalkleiner, line 62: This will probably only need to be a space, we can't be adding dots into the content (not even an ellipsis).

GuySartorelli: Only adding a single space will help to ensure words aren't immediately adjacent to one another, but it will also mean that sometimes searching for a phrase could result in incorrectly finding a phrase between the end of one block and the start of another...

Would you object to me making this configurable? I will then have it as a space by default, but developers can choose to use whatever delimiter they choose between blocks.

michalkleiner: Yep, get your point around phrase queries. This mechanism of concatenating block content isn't ideal, same as the rendered version. The best way would be to where the content is added into multi-value solr content field separately, not as a single value.
Configurable delimiter will do for now.

src/Models/BaseElement.php
```diff
@@ -169,6 +170,15 @@ class BaseElement extends DataObject
...
     private static $displays_title_in_template = true;
 
+    /**
+     * Determines whether a block should be indexable in search.
+     *
+     * @config
+     * @var boolean
+     * @see ElementalPageExtension::getElementsForSearch()
+     */
+    private static $is_indexable = true;
```
emteknetnz, line 180:
```suggestion
    private static $search_indexable = true;
```
We probably want the word 'search' explicitly in the var name, so it's not confused with a database index.  Can also remove the `is_` prefix

src/Extensions/ElementalPageExtension.php
```diff
@@ -25,6 +25,17 @@ class ElementalPageExtension extends ElementalAreasExtension
...
+    /**
+     * The delimiter to separate distinct elements in indexed content.
+     *
+     * When using the getElementsForSearch() method to index all elements in a single field,
+     * a custom delimiter can be used help to avoid false positive results for phrase queries.
+     *
+     * @config
+     * @var string
+     */
+    private static $delimiter_for_indexed_elements = ' ';
```
emteknetnz, line 37: What's a use case for this to not be a space?

I can't really think of what we'd want to do something like put a dash `-` between element content in search results?

If we do retain this configurable var, could we rename it to `$search_index_element_delimiter`?

src/Models/BaseElement.php
```diff
@@ -470,6 +480,33 @@ public function Top()
...
 
+    /**
+     * Determines whether this elemental block is indexable in search.
+     *
+     * By default, this uses the configurable variable is_indexable, but
+     * this method can be overridden to provide more complex logic if required.
+     *
+     * @return boolean
+     */
+    public function getIsIndexable(): bool
```
emteknetnz, line 491:
```suggestion
    public function getSearchIndexable(): bool
```

src/Extensions/ElementalPageExtension.php
```diff
@@ -43,7 +54,14 @@ public function getElementsForSearch()
                 /** @var ElementalArea $area */
                 $area = $this->owner->$key();
                 if ($area) {
-                    $output[] = strip_tags($area->forTemplate());
+                    foreach ($area->Elements() as $element) {
+                        if ($element->getIsIndexable()) {
+                            $forIndexing = $element->getContentForSearchIndex();
```
emteknetnz, line 59:
```suggestion
                            $content = $element->getContentForSearchIndex();
```

emteknetnz (changes requested):
Seems good, some requested changes

Could you also add some unit tests?  Particularly for BaseElement::getContentForSearchIndex().

Let me know if not sure how to go about setting up tests for this and I'll point you in the right direction

src/Models/BaseElement.php
```diff
@@ -470,6 +480,33 @@ public function Top()
...
+    {
+        return (bool)$this->config()->get('is_indexable', Config::UNINHERITED);
+    }
+
+    /**
+     * Provides content to be indexed in search.
+     *
+     * @return string|null HTML
+     */
+    public function getContentForSearchIndex()
```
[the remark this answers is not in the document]
GuySartorelli: Would you be okay with `?string`? See discussion in PHPDoc conversation.

src/Models/BaseElement.php
```diff
@@ -470,6 +480,33 @@ public function Top()
...
+     */
+    public function getIsIndexable(): bool
+    {
+        return (bool)$this->config()->get('is_indexable', Config::UNINHERITED);
+    }
+
+    /**
+     * Provides content to be indexed in search.
+     *
+     * @return string|null HTML
```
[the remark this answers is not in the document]
GuySartorelli: This is intentionally nullable so that developers can return null from an overridden method in a subclass if they explicitly don't want content to be indexed. Let me know if you're okay with that approach - I should probably specify that in the PHPDoc for clarity.

src/Extensions/ElementalPageExtension.php
GuySartorelli: I initially had it as ellipses but @michalkleiner rightly pointed out that may be unexpected for some developers. In my projects I will want to use an ellipses so that search results show a clear separation between content from separate elemental blocks.

GuySartorelli:
@emteknetnz I'll make the requested changes as time permits, hopefully in the next week. Please refer to the discussion on the PHPDoc and let me know on that one.

I note there is also a conflict now so I'll resolve that at the same time.

src/Models/BaseElement.php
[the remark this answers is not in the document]
emteknetnz: What's the functional difference between `null` and `''` (empty string) though?  Is the something in the FullTextSearch module (or another search module?) that explicitly tests for `null`?

src/Extensions/ElementalPageExtension.php
emteknetnz: OK the ellipses use case makes sense.  Yes rename thanks

src/Models/BaseElement.php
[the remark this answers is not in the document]
GuySartorelli: It's a semantic difference rather than a functional one so I'm not too worried about it at the end of the day.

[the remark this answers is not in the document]
GuySartorelli: I've left this nullable for now and made the typehint nullable because it is quite common for developers to return null from methods such as this one in ss projects - happy to remove nullability in the end if you want, but I had to make a call.

GuySartorelli:
@emteknetnz I've made the requested changes (though see [discussion about phpdoc return type](#discussion_r788320014) which affects the [related typehint](#discussion_r788319930) except tests. I've also added some documentation.

> Let me know if not sure how to go about setting up tests for this and I'll point you in the right direction
Yes please. I thought the tests would be relatively straight forward but I spent a good 15 minutes and then realised I'm not really sure what exactly I _want_ to be testing... if you could give me a direction (even just some method signatures or comments to build off) that would be great.

GuySartorelli:
Failed tests are because I forgot `Extension` isn't inherently `Configurable` - I'll fix that at the same time as writing tests once I know what tests to write.

src/Models/BaseElement.php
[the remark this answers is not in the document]
emteknetnz: Returning a blank string also won't get indexed :-) i.e. what search term could and end user put into a search box to match blank?  I just can't see a use-case where `null` does something that `''` doesn't, so I'd prefer to keep the API simple and just have a `string` return type

src/Models/BaseElement.php
```diff
@@ -476,6 +486,33 @@ public function Top()
...
+     * Determines whether this elemental block is indexable in search.
+     *
+     * By default, this uses the configurable variable search_indexable, but
+     * this method can be overridden to provide more complex logic if required.
+     *
+     * @return boolean
+     */
+    public function getSearchIndexable(): bool
+    {
+        return (bool)$this->config()->get('search_indexable', Config::UNINHERITED);
```
emteknetnz, line 499:
```suggestion
        return (bool) $this->config()->get('search_indexable', Config::UNINHERITED);
```

emteknetnz (changes requested):
Tiny linting thing and change return type to `string`.  Cheers

emteknetnz:
@GuySartorelli This would be a good level of test coverage for the work you've done here:

Setup:
- Create a couple of test elements of ElementContent (comes with `HTML`  db field).
- Create a test Extension and put it on ElementContent that implements the `updateContentForSearchIndex` hook.
- Add elements to an ElementalArea + SiteTree

See and example [here](https://github.com/silverstripe/silverstripe-asset-admin/blob/1/tests/php/Forms/FileFormBuilderTest.php#L121) for how to add test extensions in a unit test

To change the config for the delimiter in the unit test, simply `MyClass::config()->set('config_var', $value);` - it will reset between tests so won't affect other tests (we'll I'm at least 99% sure on that :-)

Tests:
a) Test that changing `search_index_element_delimiter` delimits the output
b) Test that getContentForSearchIndex adds spaces to HTML
c) Test that an extension on getContentForSearchIndex updates the content

src/Models/BaseElement.php
GuySartorelli: Darnit I felt like I had missed something 😅

src/Models/BaseElement.php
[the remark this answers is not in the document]
GuySartorelli: I can agree to that logic! I'll make the change.

GuySartorelli:
Tests failed - working on that now.

GuySartorelli:
@emteknetnz I've added the requested tests and those are passing. There is one test that's failing, but I haven't touched that test and (from what I can tell) none of the changes I've made should be causing this to fail. Can you please take a look?

emteknetnz:
@GuySartorelli found the issue for the travis failure, it was an existing bug - have created a PR https://github.com/silverstripe/silverstripe-elemental/pull/957

emteknetnz (approved):
Travis failure is an existing bug that this PR highlighted, separate PR to fix it is here https://github.com/silverstripe/silverstripe-elemental/pull/957

PR #916, opened 2021-07-14
add sort index and fix typo

sunnysideup:


src/Models/BaseElement.php
```diff
@@ -90,6 +90,10 @@ class BaseElement extends DataObject
         'canUnpublish' => DBBoolean::class,
         'canDelete' => DBBoolean::class,
     ];
+    
+    private static $indexes = [
+        'Sort' => true,
```
ScopeyNZ, line 95:
```suggestion
        'Sort' => true,
```

src/Models/BaseElement.php
```diff
@@ -345,7 +349,7 @@ public function getCMSFields()
 
                 $fields->insertBefore($styleDropdown, 'ExtraClass');
 
-                $styleDropdown->setEmptyString(_t(__CLASS__.'.CUSTOM_STYLES', 'Select a style..'));
+                $styleDropdown->setEmptyString(_t(__CLASS__.'.CUSTOM_STYLES', 'Select a style...'));
```
emteknetnz, line 352:
```suggestion
                $styleDropdown->setEmptyString(_t(__CLASS__.'.CUSTOM_STYLES', 'Select a style..'));
```

emteknetnz: Updated the default string won't match the yml - https://github.com/silverstripe/silverstripe-elemental/blob/4/lang/en.yml#L11 - unfortunately changing strings in en.yml is pretty fraught and can cause the corresponding translations in transifex are likely revert, so we'd need to create an entirely new string here.  For this reason we should just keep this as is.

emteknetnz (changes requested):
Looks fine, though we'll need to revert the typo fix

src/Models/BaseElement.php
sunnysideup: fair enough, what is the easiest way to revert this?

sunnysideup: I have fixed this properly - hopefully I did it right.

emteknetnz: Adding new translations really needs to happen via transifex, not updating the yml directly (it's not an ideal process).

The existing translations are a mixture of '..' and '...', rather than updating yml, I think the following would make more sense (untested):
```
$str = _t(__CLASS__.'.CUSTOM_STYLES', 'Select a style..');
$str = preg_replace('#([^\.])\.\.$#', '$1...', $tr);
$styleDropdown->setEmptyString($str);
```

michalkleiner: As noble as fixing the typo in the default string (and the translations) is, I think it shouldn't be a part of this PR.
Ideally, a PR addresses one main thing only.

sunnysideup: I agree, but I am short on time so I will leave it up to you to fix then. The double dot looks like someone did not finish something (maybe because they ran out of time)?

PR #918, opened 2021-08-01
Fix Empty String

sunnysideup:


michalkleiner (approved):

emteknetnz:
Sorry we can't merge this.  The translation process which involves running the i18nTextCollector task (usually run quarterly via [cow](https://github.com/silverstripe/cow)) and pushing/pulling changes to https://www.transifex.com/ is extremely temperamental

Basically we cannot alter any existing keys or source translations (en.yml), or else when happens is that existing translation (e.g. de.yml) get wiped from the registry in transifex.com and and then removed from the yml when cow pulls the changes.

PR #924, opened 2021-08-12
Allow custom actions to be added

adrhumphreys:
The idea here is that you can then in modules simply update this section to add custom actions like so:

```js
import Injector from "lib/Injector";

const WorkflowAction = (CustomActionsComponent) => (props) => {
  const { children } = props;

  return (
    <CustomActionsComponent>
      {children}
      <button>Example</button>
    </CustomActionsComponent>
  );
};

export default () => {
  Injector.transform("workflow", (updater) => {
    updater.component(
      "ElementCustomActions",
      WorkflowAction,
      "WorkflowElementalActions"
    );
  });
};
```

By default if there are no custom actions then it will return null so that it doesn't add anything to the DOM

### DIFF
--- a/client/src/components/ElementActions/CustomActions.js
+++ b/client/src/components/ElementActions/CustomActions.js
@@ -1,0 +1,18 @@
+import React from "react";
+
+const CustomActions = (props) => {
+  const { children } = props;
+
+  // Don't render if there are no children
+  if (!Array.isArray(children) || !children.length > 0) {
+    return null;
+  }
+
+  return (
+    <div role="none" onClick={(event) => event.stopPropagation()}>
+      {children}
+    </div>
+  );
+};
+
+export default CustomActions;
--- a/client/src/components/ElementEditor/Header.js
+++ b/client/src/components/ElementEditor/Header.js
@@ -112,6 +112,7 @@ class Header extends Component {
       activeTab,
       expandable,
       ElementActionsComponent,
+      ElementCustomActionsComponent,
       handleEditTabsClick,
     } = this.props;
 
@@ -160,6 +161,11 @@ class Header extends Component {
           <h3 className={titleClasses}>{element.title || noTitle}</h3>
         </div>
         {!simple && <div className="element-editor-header__actions">
+          {ElementCustomActionsComponent ? (
+            <div role="none" onClick={(event) => event.stopPropagation()}>
+              <ElementCustomActionsComponent {...this.props} />
+            </div>
+          ) : null}
           {expandable &&
             <div
               role="none"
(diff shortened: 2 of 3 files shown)


client/src/components/ElementActions/CustomActions.js
```diff
@@ -0,0 +1,18 @@
...
+const CustomActions = (props) => {
+  const { children } = props;
+
+  // Don't render if there are no children
+  if (!Array.isArray(children) || !children.length > 0) {
+    return null;
+  }
+
+  return (
+    <div role="none" onClick={(event) => event.stopPropagation()}>
```
adrhumphreys, line 12: This is annoyingly required as the div it sits within has an event listener which will try and navigate away from the page if we don't stop the event from propagating

client/src/components/ElementEditor/Header.js
```diff
@@ -160,6 +161,7 @@ class Header extends Component {
           <h3 className={titleClasses}>{element.title || noTitle}</h3>
         </div>
         {!simple && <div className="element-editor-header__actions">
+          <ElementCustomActionsComponent {...this.props} />
```
adrhumphreys, line 164: I found that there's likely a need for all the props as we don't know what this will be used for (in my case I just needed the element ID)

emteknetnz:
What does the 'WorkflowElementalActions' string refer to in your code sample?

Could you provide a screenshot of what this looks like?

adrhumphreys:
Heya,

Just updated the code for the button to be the following (so it's just an icon):
```jsx
<button className="btn btn-secondary">
    <span className="font-icon-export" />
</button>
```

An example with a non inline editable block:
image

And then inline editable:
image

---
From what I understand, the string `WorkflowElementalActions` refers to the name of the update. So if someone wanted to apply an update to it only after this update had run they could do something like:

```js
Injector.transform(
  'ElementCustomActions', 
  (update) => update.component('ElementCustomActions', MyCustomWrapper),
  { after: 'WorkflowElementalActions' }
);
```

_This is what I understand that line to do but I could be wrong and you can just remove it and everything still works_

emteknetnz:
Do we actually need this new CustomActions component?  I'm kind of surprised that we cannot just add to the existing 'ElementActions' via Injector - https://github.com/silverstripe/silverstripe-elemental/blob/4/client/src/boot/registerTransforms.js#L78 i.e.

```js
Injector.transform('element-actions', (updater) => {
    updater.component("ElementActions", WorkflowAction, "MyWorkflowAction");
```

Have you tried using this instead?  If it does not work, is it because you'd need to run `yarn build` on within the `silverstripe-elemental` directory? (I'm honestly not sure how dynamic JS injector is)

adrhumphreys:
The main reason you can't take that route is it only supports inline blocks and blocks that are not inline don't get the actions added.

Otherwise if you just wanted to edit actions for inline blocks you could do something like:
```js
const ElementActions = (props) => {
   const { ActionCompontent } = props;

   return (
        <div className="workflow-element-actions">
            <WorkflowButton {...props} />
            <ActionCompontent {...props} />
        </div>
    );
}

Injector.transform("workflow", (updater) => {
  updater.component(
    "ElementActions",
    (ActionCompontent) => (props) => (
      <ElementActions ActionCompontent={ActionCompontent} {...props} />
    ),
    "WorkflowElementActions"
  );
});
```

Issue #512, opened 2018-11-19
Some of the "More Actions" should still show on non in-line editable blocks


ScopeyNZ:
### Affected Versions

4.x

### Description

When a block is _not_ in-line editable it still makes sense for some actions to exist:

- Publish
- Unpublish
- Archive

### Steps to reproduce

- Create a non in-line editable block type (or install one like `dnadesigns/silverstripe-elemental-userforms`)
- Add that block type to an elemental area
- Observe there's no "More actions" menu

## Designs

See block basics designs https://projects.invisionapp.com/dsm/silver-stripe/silverstripe-cms/nav/5fa7cb5f8c012000183561a5/asset/607e6266ade40316eb2697ee/tab/design?mode=preview

## ACs
- Confirm the final list of actions to be included in the list (note to not include the 'tabs' like Content and Settings)
  - Ensure that the new 'Duplicate' action is present in the list (https://github.com/silverstripe/silverstripe-elemental/issues/881)
- The finalised list of actions are implemented
- Ensure that the status icons on the block reflect any action made, i.e publishing the block updates the status icon
- Close this https://github.com/silverstripe/silverstripe-elemental/issues/881

### PRs
- [ ] ~~https://github.com/silverstripe/silverstripe-elemental/pull/925~~
- [ ] https://github.com/silverstripe/silverstripe-elemental/pull/963

robbieaverill:
Do we have designs with the `>` caret _and_ the dropdown? I don't remember seeing anything

ScopeyNZ:
I think the original designs didn't have the "meatballs" on non in-line editable blocks. But I also remember us discussing this a few weeks ago and noting it'd be a UX improvement. Worth tracking at least :slig

robbieaverill:
Ok cool. This is probably one of those options where the UX team popping down while we design on the fly would be the easiest solution, but for the sake of open source perhaps @sachajudd you could provide a quick design mockup of what you want it to look like in case a community member wants to pick this up before we get to it

sachajudd:
See collapsed [dnadesigns/silverstripe-elemental-userforms](https://projects.invisionapp.com/dsm/silver-stripe/silver-stripe/asset/components/5bda2b9c9029b00017b97a65) block.

chillu:
+1, just come across this as well.

clarkepaul:
Related, adding duplicate blocks to non-inline blocks https://github.com/silverstripe/silverstripe-elemental/issues/881

[...]

PR #925, opened 2021-08-16
ENH Show actions on non-inline editable blocks

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-elemental/issues/512

I've removed the 'Save' button for non inline-editable blocks as it doesn't make any sense in a context where there's no content that you can change

Video of a normal block + a non-inline-editable block - https://www.youtube.com/watch?v=LYFIMXor8FQ


emteknetnz:
@clarkepaul I've attached a video above on a non-inline editable block - I've removed the 'save' action from the list for non-inline-editable - could you check that you're happy with it?

client/src/components/ElementEditor/ElementActions.js
```diff
@@ -74,7 +74,34 @@ class ElementActions extends Component {
...
+        return ret;
+    };
+
+    let kids = children;
+    if (children && Array.isArray(children)) {
+      kids = flatten(children).filter(child => {
+        if (!child || !child.props) {
+          return false;
+        }
+        if (child.props.title === 'Save') {
```
emteknetnz, line 99: Having a hardcoded string here isn't great, though I'm not sure a better way to do this

Maybe add a "showIfNonInlineEditable" property on AbstractAction?  That seems worse though IMO

clarkepaul:
@emteknetnz looks all good to me, no issues that I could see. FYI the experience of editing a title on a non-inline block is fine as-is too.

client/src/components/ElementEditor/tests/ElementActions-test.js
```diff
@@ -10,26 +10,24 @@ import AbstractAction from 'components/ElementActions/AbstractAction';
...
 describe('ElementActions', () => {
-  const ActionMenuComponent = (props) => <div>{props.children}</div>;
-  const testTabs = [
-    { title: 'Content', name: 'Main' },
-    { title: 'Settings', name: 'Settings' },
-    { title: 'History', name: 'History' }
-  ];
-  const editTabsClick = () => {};
+  let props = {};
+  beforeEach(() => {
```
emteknetnz, line 14: I did some refactoring on the unit test cos there was a lot of code duplication between the existing tests

client/src/components/ElementEditor/ElementActions.js
```diff
@@ -17,6 +17,25 @@ class ElementActions extends Component {
...
   }
 
+  /**
+   * Flattens a nested array
+   * Polyfill for Array.flat() which isn't available for jest tests (or IE11)
+   *
+   * @param {Array} arr
+   * @return {Array}
+   */
+  flatten(arr) {
```
michalkleiner, line 27: Could/should we check if `Array.flat()` is available first and use that and only use the custom implementation when it isn't?

client/src/components/ElementEditor/ElementActions.js
```diff
@@ -74,7 +93,16 @@ class ElementActions extends Component {
    * @returns {ActionMenuComponent|null}
    */
   render() {
-    const { children, id, ActionMenuComponent } = this.props;
+    const { children, id, ActionMenuComponent, inlineEditable } = this.props;
+
+    // Flatten the recursive structure of children - [[[..., react.element], react.element]
+    // and filter AbstractActions's that can be shown depending if the Element is inlineEditable
+    let kids = children;
```
michalkleiner, line 100: Can we find a more descriptive variable name than `kids`, representing why it's different from `children`?

emteknetnz: I think I just used kids as a placeholder variable since I couldn't think of anything better at the time and intended to change it later.  Guess I forgot to :)

client/src/components/ElementEditor/ElementActions.js
emteknetnz: Just IE11 is missing Array.flat() - https://caniuse.com/array-flat

michalkleiner: We would need to add silverstripe-cms @ 4.9 as the minimal required version in that case?

client/src/components/ElementEditor/ElementActions.js
```diff
@@ -93,7 +121,7 @@ class ElementActions extends Component {
       >
         { this.renderEditTabs() }
         { this.renderDivider() }
-        { children }
```
maxime-rainville, line 96: I would add a "Context Provider" around `children` here. The context could tell sub component if they are rendered for an inline element or not.

client/src/components/ElementEditor/ElementActions.js
maxime-rainville: I really don't like this approach. This seems like an anti-pattern.

Element actions shouldn't have to worry about filtering out it's children. The `showForNonInlineEditableBlock` prop is literally not used in the `AbstractAction` ... it's just there as a flag to be filtered out here.

client/src/components/ElementActions/AbstractAction.js
```diff
@@ -14,9 +14,10 @@ const AbstractAction = (props) => {
     className: classNames(className, 'dropdown-item'),
     ...props,
   };
+  const { showForNonInlineEditableBlock, ...reactStrapProps } = itemProps;
```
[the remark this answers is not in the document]
maxime-rainville: Use a "context consumer" here and return `false` if `showForNonInlineEditableBlock` is `false` and the context is for a none-inline block.

client/src/components/ElementActions/AbstractAction.js
```diff
@@ -30,11 +31,13 @@ AbstractAction.propTypes = {
   name: PropTypes.string,
   type: elementTypeType,
   active: PropTypes.bool,
-  label: PropTypes.string
+  label: PropTypes.string,
+  showForNonInlineEditableBlock: PropTypes.bool
```
maxime-rainville, line 35: Not a big fan of the wording here. Could we rename this to something like `alwaysShow` or `showOnlyForInlineEditable`.

maxime-rainville (changes requested):
The approach is wrong. We should use context instead: https://reactjs.org/docs/context.html:
- Define an `InlineEditableContext`, either in a separate file or directly in `ElementActions` ... or maybe even in `Element`.
- In `ElementActions` or in `Element`, use `InlineEditableContext.Provider` to say if the current element is inlineEditable or not.
- In `AbstractAction`, use `InlineEditableContext.Consumer` to figure out if the current action is rendered inside an inline-editable element and return `false` if `showForNonInlineEditableBlock` is `false`.

maxime-rainville:
Superseded by https://github.com/silverstripe/silverstripe-elemental/pull/963

PR #929, opened 2021-08-26
Reorder function

Jianbinzhu:
Updated reorder function to account for when there are duplicate Sort values of `elements` within an `ElementalArea`


ScopeyNZ:
See #931

Issue #764, opened 2019-12-16
CMS users lose all blocks when attempting to save page with validation errors


scott1702:
## Overview

This may be more of a framework issue than an elemental issue, but haven't investigated enough to know.

When users attempt to save/publish a page, if the page has any validation errors then any content blocks which have been created get entirely removed as the page fails to save. This can be especially frustrating if there are required fields in other tabs which the user hasn't seen yet.

## Steps to reproduce:
- Create a page with required fields and an elemental area
- Fill out the page with numerous content blocks and leave the required field empty
- Save and/or publish the page

### Actual result:

- The content blocks are removed as the page has failed to save due to validation error

### Expected result:

- The content blocks are retained in an unsaved state, in the same manner as other fields

## Acceptance Criteria

* When a validation error against a _Page-level_ field is returned from attempting to save / publish the page, blocks and their unsaved contents are retained in the editor.
* When a validation error against a _Block-level_ field is returned from attempting to save / publish the page, blocks and their unsaved contents are retained in the editor.

## Out of scope

- [ ] Validation errors are displayed next to the elemental block - should be covered in the following: https://github.com/silverstripe/silverstripe-elemental/issues/329

## Notes

- We have some old PR that might fix this issue. If they don't fix the issue and are not salvageable, bring the card back into refinement.

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1246
- https://github.com/silverstripe/silverstripe-elemental/pull/930

ScopeyNZ:
@brynwhyman I was torn between `impact/high` and `impact/critical` here - feel free to re-adjust.

This doesn't surprise me, but it's obviously a bad bug. We had to work around the fact that SilverStripe replaces the whole form when saving with our new React editor. Redux should remember all of the form data, but it might be getting cleared on the response from the server, regardless of whether it was successful or not. I wonder if this can be _relatively_ easily fixed by selectively clearing redux here:

https://github.com/dnadesign/silverstripe-elemental/blob/4/client/src/legacy/ElementEditor/entwine.js#L72

GuySartorelli:
In my recent experience with version 4.4.0, the blocks themselves remain but the content within the block is cleared.

chrispenny:
It's worth mentioning that you will also experience the same thing if you apply validation requirements to an Element through the `validate()` extension point.

EG, I use the following `addFieldError()` method:
```
$validationResult->addFieldError('MyField', 'Required');
```

[...]

PR #930, opened 2021-08-26
FIX Retain value that failed validation on client side

emteknetnz:
Requires this PR https://github.com/silverstripe/silverstripe-admin/pull/1246 to get the ValidationResult client-side

### ~Video~
- ~https://www.youtube.com/watch?v=NvWJHFSHijo~
- ~custom "required" validator on content block title~
- ~RequiredFields validator on a custom page field~
- ~Page has multiple tabs that show invalid icons~

## Make sure you squash commits on merge
There are two commits so it is clear what was done originally vs what has been done more recently. The original commit may be useful for historical purposes but when merging, this PR should be merged with squashed commits.

## CI
The CI won't go green until https://github.com/silverstripe/silverstripe-admin/pull/1246 is merged in, so I've run an installer CI run: https://github.com/creative-commoners/silverstripe-installer/actions/runs/4985675862

Note that the failed unit test in the installer CI run is unrelated to this work.

## Issue
https://github.com/silverstripe/silverstripe-elemental/issues/764


emteknetnz:
1. yep

2. is technically possible, though its implementation is honestly quite hackish.  Also I think it's wrong to put field errors inside the general form error.  Also it will be confusing for users that have 2 or more elemental areas, I think warnings above the relevant elemental area would be better.

3. can do

7. that's doable

> highlighting blocks with some sort of visual indication which have error

Unfortunately not possible without very hacky code, because it would be jquery updating a rendered react component

clarkepaul:
> 2. "I think warnings above the relevant elemental area would be better." 

If there wasn't another competing error at the top of the page which is kinda repeating information, and also close in proximity perhaps. Watching the video it's very in your face when you have multiple errors showing with multiple error notifications as well, and then there are instances where the field shows an error too. Just think this might be doing more harm than good having so many errors present.

image

I understand there might be technical limitations which mean the experience is sub-optimal especially with #4 not being possible.

I think we absolutely need an error to be sticky near the top incase people might miss the fading toast notifications. An idea (might not be the right thing?), we might have the possibility to use sticky toasts with a dismiss (x). Then the top alert info could be transferred into the toast. I don't think we've used the Toast dismiss yet but this could be an okay scenario to use it.

> Also I think it's wrong to put field errors inside the general form error. 

This was actually seen as best practice years ago when lots of errors can present (pretty sure it still would be), means you can assess all the issues in one place to get a sense of what needs to be done. Helps to prevent people correcting one field and then try to save the form again but only to get the following issue for each repeated save.

Keen to hear your thoughts on moving the top alert info to a sticky toast @emteknetnz

yarn.lock
GuySartorelli, line 1: Not sure why these changes are here but given this is an automatically generated file I'm inclined to keep them

tests/Behat/features/edit-block-element.feature
GuySartorelli, line 1: Changes to this file are to make the tests more robust, mirroring the equivalent assertions in the new validation failed feature tests.

tests/Behat/Context/FeatureContext.php
```diff
@@ -240,19 +240,27 @@ public function stepIFillInForForBlock($value, $name, $blockNumber)
...
+        $regex = '/^' . preg_quote($content, '/') . '$/ui';
+
+        if ($negate) {
+            $message = sprintf('The field "%s" value is "%s", but "%s" expected.', $field, $actual, $content);
+            Assert::isTrue((bool) preg_match($regex, $actual), $message);
         } else {
-            $this->assertFieldContains($field, $content);
+            $message = sprintf('The field "%s" value is "%s", but it should not be.', $field, $actual);
+            Assert::isFalse((bool) preg_match($regex, $actual), $message);
         }
```
GuySartorelli, line 264: `assertFieldNotContains()` and `assertFieldContains()` failed because the `$field` variable wasn't a string (so obviously this part of the code just never worked but was never used).

The new assertions mirror what `assertFieldNotContains()` and `assertFieldContains()` methods _would_ have done. I don't use those directly because those will look fo the fields on the page, not on the block - and since we already have a reference to the field on the block, it's more sensible to just check against that directly here.

client/src/legacy/ElementEditor/entwine.js
```diff
@@ -60,16 +62,35 @@ jQuery.entwine('ss', ($) => {
...
+            $('.cms-edit-form').data('hasValidationErrors', false);
+            resetStores();
+          } else {
+            $('.cms-edit-form').data('hasValidationErrors', true);
+          }
+        } else {
+          // No validation result returned
+          // assume everything is valid and reset stores so data is refetched.
+          resetStores();
+        }
```
GuySartorelli: This is needed for compatibility with versions of silverstripe/admin that don't have the accompanying PR.

client/src/legacy/ElementEditor/entwine.js
```diff
@@ -60,16 +62,35 @@ jQuery.entwine('ss', ($) => {
     },
 
     onunmatch() {
-      resetStores();
+      // Reset the store if the user navigates to a different part of the CMS
+      // or after submission if there are no validation errors
+      if (!$('.cms-edit-form').data('hasValidationErrors')) {
+        resetStores();
+      }
```
GuySartorelli, line 67: The condition doesn't explicitly test for anything to see if it's swapping to a new page - but it doesn't need to. This condition evaluates to `true` when navigating to a new page. I've confirmed this under the following scenarios:
1. No form submission has happened yet. I click to a different page.
2. A form submission happened and there were no validation errors. I click to a different page.
3. A form submission happened and there _were_ validation errors. I click to a different page.

client/src/legacy/ElementEditor/entwine.js
emteknetnz: I think I'd prefer to just tighten the constraint within elementals composer.json to `^1.13.2`

emteknetnz (commented):
I clicked "Comment" rather than "Request changes" because I'm unable to request changes since I'm the original author on this PR

client/src/legacy/ElementEditor/entwine.js
GuySartorelli: Done. CI will break until admin is tagged.

emteknetnz (commented):
APPROVE
(I cannot tick approve because I'm the original author on the PR)

Tested locally, it's certainly a big approvement though the UX still isn't quite there

When I make inline edits to a content block such as changing the title and then save the page and fail page validation, the content block edits are retained, though the when the page edit form is refreshed the content block title appears unchanged in a closed state. When you open the content block to inline edit it the changes are now applied.

That said, it is now retaining the edits which is a great improvement

Issue #849, opened 2020-11-11
sorting bug, when moving element up one from bottom the sort order is not saved


haseganats:
We did a fresh install of elemental using SS4.6 and latest version of Elemental. What we found is that if move a block from bottom position up by one using the drag and drop UX, it works fine, BUT it does not save (i.e. when we save and publish the block is back in its original position). This does not happen if:

a. block is moved up more than one position
b. block is not move from bottom position (i.e. there is/are other blocks below it).

**Note**: [further replication steps in comment below](https://github.com/silverstripe/silverstripe-elemental/issues/849#issuecomment-922536928)

bergice:
@haseganats could you post your `composer show`? I'm unable to replicate this issue.

haseganats:
bramus/ansi-php                           3.0.3              ANSI Control Functions and ANSI Control Sequences (Colors, Erasing, etc.) for PHP CLI Apps
bramus/monolog-colored-line-formatter     2.0.3              Colored Line Formatter for Monolog
composer/ca-bundle                        1.2.8              Lets you find a path to the system CA bundle, and includes a fallback to the Mozilla CA bundle.
composer/installers                       v1.9.0             A multi-framework Composer library installer
dnadesign/silverstripe-elemental          4.4.1              Elemental pagetype and collection of Elements
doctrine/instantiator                     1.3.1              A small, lightweight utility to instantiate objects in PHP without invoking their constructors
embed/embed                               v3.4.9             PHP library to retrieve page info using oembed, opengraph, etc
guzzlehttp/psr7                           1.7.0              PSR-7 message implementation that also provides common utility methods
intervention/image                        2.5.1              Image handling and manipulation library with support for Laravel integration
league/csv                                8.2.3              Csv data manipulation made easy in PHP
league/flysystem                          1.0.70             Filesystem abstraction: Many filesystems, one API.
m1/env                                    2.2.0              Env is a lightweight library bringing .env file parser compatibility to PHP. In short - it enables you to read .env fil...
marcj/topsort                             1.1.0              High-Performance TopSort/Dependency resolving algorithm
monolog/monolog                           1.25.5             Sends your logs to files, sockets, inboxes, databases and various web services
myclabs/deep-copy                         1.10.1             Create deep copies (clones) of your objects
nikic/php-parser                          v4.10.2            A PHP parser written in PHP
phpdocumentor/reflection-common           2.2.0              Common reflection classes used by phpdocumentor to reflect the code structure
[...]

PR #931, opened 2021-08-26
Reorder function

Jianbinzhu:
Updated reorder function to account for when there are duplicate Sort values of elements within an ElementalArea

### DIFF
--- a/src/Services/ReorderElements.php
+++ b/src/Services/ReorderElements.php
@@ -138,18 +138,32 @@ class ReorderElements
     /**
      * Reorder all elements within the current ElementalArea
      * 
-     * @param $parentId The parent ID of the current element
+     * @param int $parentId The parent ID of the current element
      */
-    private function reorderAllElements($parentId)
+    private function reorderAllElements(int $parentId)
     {
         if (!$parentId) return;
         
         $elements = BaseElement::get()->filter('ParentID', $parentId)->sort('Sort', 'asc');
-        $sort = 1;
 
-        foreach ($elements as $element) {
-            $element->Sort = $sort++;
-            $element->write();
+        $baseTableName = Convert::raw2sql(DataObject::getSchema()->tableName(BaseElement::class));
+
+        // Update both the draft and live versions of the records
+        $tableNames = [$baseTableName];
+        if (BaseElement::has_extension(Versioned::class)) {
+            /** @var BaseElement&Versioned $element */
+            $tableNames[] = $this->element->stageTable($baseTableName, Versioned::LIVE);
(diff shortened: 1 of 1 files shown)


Jianbinzhu:
@ScopeyNZ I came across an issue where elements under the same ParentID have duplicate Sort value, and I can't seem to replicate that anymore. To fix existing records, I've created a sort all elements function. Please advise if there's a better way of doing this, thanks.

Screen Shot 2021-08-30 at 5 35 06 PM

ScopeyNZ:
I don't really like the idea of a side effect of somebody reordering their elements is a long running "fix the whole thing", that also doesn't have any guarantee of respecting the order they chose.

If there's a bug in this code which will mess up the sort order then we should address that, but other than that, I'm thinking we shouldn't have the code be concerned with garbage data.

brynwhyman:
This just came up in a different context, possibly a bit more information about the bug at play: https://github.com/silverstripe/silverstripe-elemental/issues/849#issuecomment-922536928

chrispenny:
Just adding my $0.02.

As Bryn mentioned, I experienced this issue here https://github.com/silverstripe/silverstripe-elemental/issues/849#issuecomment-922536928 

I think we could avoid a lot of these instances simply by making sure that we don't set a `Sort` value on our blocks until they actually have a Parent.

Jianbinzhu:
@brynwhyman  @chrispenny  Just found out that this solution is only partially working 😂 , it doesn't work when when have 2 elements with sort value of `1` and then trying to drag the second element to the top. `$elementToBeAfterID` on L72 will be 0 as the element to be after will be none. Which means that `$sortAfterPosition` will stay `0` and then we will be comparing `$sortAfterPosition 0` to `$currentPosition 1` and it won't hit that else condistion. I've created a task to fix the broken elements instead for now.

Another thing I found is that the duplicated sort value is caused by when an Elemental Block is archived and then restored

chrispenny:
@Jianbinzhu good find with the archive restoration. Based on @ScopeyNZ's comment, this would be an area where we would want to fix the "restoration process" itself.

EG: Maybe we have it set the `Sort` to the max `Sort` value + 1 from that `Parent` (put it down the bottom) any time we bring an archived block back to life.

ScopeyNZ:
In regards to the sort order on archive/publish. I think it would be fine to clear the sort order value when a block is archived. I'll raise an issue for sort being broken as elements are restored from the archive.

WRT this PR, I'm going to close it as I don't think it's the solution we're looking for. Thanks for the help getting to the bottom if this issue!

PR #919, opened 2021-08-01
Patch 2

sunnysideup:

PR #932, opened 2021-09-20
ENH Add index for Sort field.

### LINKED ISSUE #915: add index on sort for faster sorting
Adding an index on the `Sort` field will make retrieval faster.

GuySartorelli:
Fixes #915.


GuySartorelli:
Nevermind - see #919

PR #937, opened 2021-10-04
Bugfix #849: Set default Sort only when ParentID is available

### LINKED ISSUE #849: sorting bug, when moving element up one from bottom the sort order is not saved

chrispenny:
Fixes #849
Related to #931

- Do not set a default `Sort` value until a `ParentID` is available on the Element.
- Switch to "exit early" paradigm (easier to read).

### DIFF
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -285,17 +285,25 @@ class BaseElement extends DataObject
     {
         parent::onBeforeWrite();
 
-        if (!$this->Sort) {
-            if ($this->hasExtension(Versioned::class)) {
-                $records = Versioned::get_by_stage(BaseElement::class, Versioned::DRAFT);
-            } else {
-                $records = BaseElement::get();
-            }
-
-            $records = $records->filter('ParentID', $this->ParentID);
-
-            $this->Sort = $records->max('Sort') + 1;
+        // If a Sort has already been set, then we can exit early
+        if ($this->Sort) {
+            return;
         }
+
+        // If no ParentID is currently set for the Element, then we don't want to define an initial Sort yet
+        if (!$this->ParentID) {
+            return;
+        }
+
+        if ($this->hasExtension(Versioned::class)) {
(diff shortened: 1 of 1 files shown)


src/Models/BaseElement.php
```diff
@@ -285,17 +285,25 @@ public function onBeforeWrite()
...
-            $this->Sort = $records->max('Sort') + 1;
+        if ($this->hasExtension(Versioned::class)) {
+            $records = Versioned::get_by_stage(BaseElement::class, Versioned::DRAFT);
+        } else {
+            $records = BaseElement::get();
         }
+
+        $records = $records->filter('ParentID', $this->ParentID);
+
+        $this->Sort = $records->max('Sort') + 1;
```
chrispenny, line 306: No changes here - just moved to "exit early" paradigm.

src/Models/BaseElement.php
```diff
@@ -285,17 +285,25 @@ public function onBeforeWrite()
...
+        // If a Sort has already been set, then we can exit early
+        if ($this->Sort) {
+            return;
+        }
 
-            $records = $records->filter('ParentID', $this->ParentID);
+        // If no ParentID is currently set for the Element, then we don't want to define an initial Sort yet
+        if (!$this->ParentID) {
+            return;
+        }
```
chrispenny, line 296: The guts of the change.

Rationale detailed here:
https://github.com/silverstripe/silverstripe-elemental/issues/849#issuecomment-922536928

PR #938, opened 2021-10-05
ENH Add extension points to update element controller styles and template

michalkleiner:
For some use cases where a block requires a different holder template based on some other business logic requirements where the config system isn't sufficiently flexible during runtime, allow to update the styles and the controller/holder template via an extension to the element controller.


michalkleiner:
@emteknetnz thanks for the review. Committed the suggestions via GitHub, please use squash when merging. Thanks!

PR #941, opened 2021-10-12
API phpunit 9 support

emteknetnz:
Issue: https://github.com/silverstripe/silverstripe-framework/issues/10019


maxime-rainville (approved):
Not sure what Scrutinizer is complaining about and I don't care enough to debug it.

Merging now.

PR #944, opened 2021-10-30
Cosmetic change to italian translation

digitall-it:
Elemental Area can be translated to "Area Elementare" (lit. "Elemental Area") or Area Elemento (lit. "Element Area", preferred). As the latter has been chosen in the past, a plural form has been matched consequently.

### DIFF
--- a/lang/it.yml
+++ b/lang/it.yml
@@ -29,7 +29,7 @@ it:
     PLURALNAME: 'Aree Elementi'
     PLURALS:
       one: 'Un''Area Elementi'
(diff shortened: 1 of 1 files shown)


ScopeyNZ:
Thanks for the contribution! Our translations are managed by Transifex - see the docs here: https://docs.silverstripe.org/en/4/contributing/translations/ .

While I could merge this, your contribution would get overwritten when the next tag is created. Please sign up for a free account on Transifex (you can use GitHub SSO) and contribute there 😄 .

PR #951, opened 2021-12-16
FIX Use correct return type for permission checks

GuySartorelli:
The PHPDoc specifies that a boolean will be returned.

Returning a null causes a problem using SearchableService which, in `SearchableService::isSearchable`, can return the value from `canView` directly, expecting it to be a `boolean` (which then raises a `TypeError`).

### DIFF
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -203,7 +203,7 @@ class BaseElement extends DataObject
             }
         }
 
-        return (Permission::check('CMS_ACCESS', 'any', $member)) ? true : null;
+        return (Permission::check('CMS_ACCESS', 'any', $member)) ? true : false;
     }
 
     /**
@@ -226,7 +226,7 @@ class BaseElement extends DataObject
             }
         }
 
-        return (Permission::check('CMS_ACCESS', 'any', $member)) ? true : null;
+        return (Permission::check('CMS_ACCESS', 'any', $member)) ? true : false;
     }
 
     /**
@@ -253,7 +253,7 @@ class BaseElement extends DataObject
             }
         }
 
-        return (Permission::check('CMS_ACCESS', 'any', $member)) ? true : null;
+        return (Permission::check('CMS_ACCESS', 'any', $member)) ? true : false;
     }
 
     /**
@@ -271,7 +271,7 @@ class BaseElement extends DataObject
             return $extended;
         }
 
-        return (Permission::check('CMS_ACCESS', 'any', $member)) ? true : null;
+        return (Permission::check('CMS_ACCESS', 'any', $member)) ? true : false;
     }
 
     /**


michalkleiner:
You could just retype the returned value via `return (bool) Permission::check(...` though I would probably spend some time trying to understand if the tri-state mechanism had a reason here.

michalkleiner:
Alternatively, updating `SearchableService::isSearchable` method to retype whatever it receives to what it needs might be the better way of fixing this. And here only update the doc block to say `boolean|null`.

michalkleiner:
I don't have time to dig into this more this week, but another core team member could advise better on how to approach this.

ScopeyNZ:
The PHPDoc is wrong. `null` is a valid return type which basically means "defer". This is a tri-state. I'd raise an issue on SearchableService if it's incorrectly expecting a boolean, or not casting the value from `DataObject::canX`

I also don't know the reason for the tri-state, but returning `null` instead of `false` if you're not trying to explicitly deny should be the preferred approach.

ScopeyNZ:
I raised an issue: https://github.com/silverstripe/silverstripe-fulltextsearch/issues/304

I'll close this for now, but if you feel as though the tri-state is wrong, feel free to re-open

GuySartorelli:
I'd agree that it is a tristate if this was a subclass of `DataExtension` where returning `null` defers to the `DataObject`s canX method... But `BaseElement` subclasses `DataObject` which dilutes the `DataExtension` tristate down to a Boolean.

The phpdocs for `DataObject` canX methods also say they return Boolean which I think is correct.

https://github.com/silverstripe/silverstripe-framework/blob/96ace07f859cbf59f7623cb05418be526b1e94ee/src/ORM/DataObject.php#L3080-L3091

And in any case changing the phpdocs to say the can return null might constitute an API change, I would think as it changes what developers should expect when calling a canX method on _any_ dataobject subclass.

GuySartorelli:
I don't seem to have the ability to reopen this, but I do think it should be reopened - at the very least a bit more discussion around this tristate and whether the phpdocs for dataobject (and literally all subclasses of dataobject) should be updated.

ScopeyNZ:
Hmmn. I've always understood that `canView` is tri-state - at least the non existent interface for it should say so. Looking at the root implementation doesn't define the interface.

I don't agree that changing the PHPDoc constitutes an API change. The existing code has always worked with `null` as it is cast as `false` usually. This is definitely a discussion to have on the framework repo though.

The code here has been considered correct for so long, I'm more concerned with developers having the same understanding as I do, and relying on `null` results from the implementation in `BaseElement` being affected by the change here.

GuySartorelli:
I have raised silverstripe/silverstripe-framework#10184 to discuss this further.

I do agree that changing the return type from here could _potentially_ cause issues for some developers, and if the consensus is that `null` is not a valid return type for these methods probably some careful consideration will be required for how and when to amend them.

Issue #954, opened 2022-01-19
Inline actions broken in GraphQL4


emteknetnz:
Inline unpublish action (via the 3 dots icon) causes the following error in GraphQL4 only

![image](https://user-images.githubusercontent.com/4809037/150064338-6bf51bc4-65b6-4738-8dcd-7d8769603aa2.png)

Travis
https://app.travis-ci.com/github/silverstripe/silverstripe-elemental/jobs/556073834

```
--- Failed scenarios:
    tests/Behat/features/archive-block-element.feature:23
    tests/Behat/features/publish-block-element.feature:23
    tests/Behat/features/unpublish-block-element.feature:23
```

unclecheese:
Fix is here https://github.com/silverstripe/silverstripe-graphql/pull/431

unclecheese:
And here https://github.com/silverstripe/silverstripe-elemental/pull/955

PR #955, opened 2022-01-20
BUGFIX: Use new typeMapping graphql API

unclecheese:
Resolves https://github.com/silverstripe/silverstripe-elemental/issues/954

### DIFF
--- a/.idea/workspace.xml
+++ b/.idea/workspace.xml
@@ -1,0 +1,54 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="ChangeListManager">
+    <list default="true" id="291031ea-f6a1-4b21-b86a-556c57c22cb2" name="Default Changelist" comment="" />
+    <option name="SHOW_DIALOG" value="false" />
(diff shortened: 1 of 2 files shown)


unclecheese:
Merging ahead of green, because it is interdependent with this graphql fix: https://github.com/silverstripe/silverstripe-graphql/pull/431

emteknetnz:
@unclecheese please don't self merge unapproved pull-requests on stable branches.  I'm less concerned about graphql4's `master` branch as it's non-stable, however elemental 4 is stable.  The build for this PR ended up [red](https://github.com/silverstripe/silverstripe-elemental/pull/955/checks) and probably would not have been merged in its current state

emteknetnz:
I suppose its an unfortunate side-effect that the elemental 4 _graphql folder is effectively at non-stable level despite it being within stable branch

unclecheese:
Well the folder itself is basically inert. It's explicitly including graphql 4.x-dev in the build matrix that makes this stable module dependent on an unstable module, and I think that's a fairly new thing. But interdependent PRs, infrequent as they may be, are just unavoidable in our setup as far as I can tell.

PR #962, opened 2022-03-10
DEP Update eslint-config

emteknetnz:
Issue https://github.com/silverstripe/eslint-config/issues/9

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "redux-form": "^6.8.0"
   },
   "devDependencies": {
-    "@silverstripe/eslint-config": "^0.0.6",
+    "@silverstripe/eslint-config": "^0.1.0",
     "@silverstripe/webpack-config": "^1.5.0",
     "babel-jest": "^23.6.0",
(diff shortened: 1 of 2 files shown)


maxime-rainville (changes requested):
I pulled this PR down and ran and build. The `yarn.lock` file was updated to reflect the new devDependency.

How come the `yarn.lock` file hasn't been updated? And why isn't the Travis build failing ... it's meant to detect this kind of oversight?

emteknetnz:
We're not updating deps here (updating yarn.lock).  What we're doing is making it so the next time `yarn upgrade` is run, it will update the version of eslint-config

There's no material difference between eslint-config 0.0.6 and 0.1.0.  The only different is that if we release 0.1.1, the ^ caret now means that asset-admin etc will get 0.1.1.  Previously with ^0.0.6 the caret effectively did nothing.

emteknetnz:
Have updated PR after running `yarn install`

maxime-rainville (approved):

PR #963, opened 2022-03-19
ENH Render some buttons for blocks that can not be edited inline

maxime-rainville:
Some block types can not be edited in an Elemental Area (e.g.: UserForm blocks). Current behaviour is to not render any actions for those blocks in the Elemental Area.

However some action still make sense in that context (e.g.: Duplicate, Archive, Unpublish).

This PR allows those extra actions to be displayed.

# Parent issue
- Fixes https://github.com/silverstripe/silverstripe-elemental/issues/512


client/src/components/ElementEditor/tests/Header-test.js
```diff
@@ -170,7 +170,7 @@ describe('Header', () => {
         />
       );
 
-      expect(wrapper.text()).not.toContain('ElementActionsComponent');
+      expect(wrapper.text()).toContain('ElementActionsComponent');
```
maxime-rainville, line 173: This is shallow rendered, and it won't go all the way to rendering the actual buttons, so I can't make assertions beyond checking that its rendering the action list or not.

maxime-rainville (commented):
Turns out we didn't the Context provider non-sense at all. We just needed to pass the expandable prop all the way down.

emteknetnz (changes requested):
Tested locally, does what it should do.  Unpublish button remains on elemental-userforms which makes sense IMO.  Some minor changes requested

client/src/components/ElementEditor/ElementActions.js
```diff
@@ -91,8 +92,8 @@ class ElementActions extends Component {
         dropdownMenuProps={{ right: true }}
         dropdownToggleClassNames={dropdownToggleClassNames}
       >
-        { this.renderEditTabs() }
-        { this.renderDivider() }
+        { expandable && this.renderEditTabs() }
```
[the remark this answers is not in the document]
emteknetnz: @maxime-rainville remember to do this one

client/src/components/ElementEditor/ElementActions.js
```diff
@@ -59,12 +60,15 @@ class ElementActions extends Component {
    * @returns {DropdownItem|null}
    */
   renderDivider() {
-    const { children, editTabs } = this.props;
+    const { children, editTabs, expandable } = this.props;
 
-    if (editTabs && editTabs.length && React.Children.count(children)) {
```
maxime-rainville, line 64: I swap the return null and return DropdownItem around just so the approach between renderDivider and renderTabs are consistien

maxime-rainville (commented):
I was thinking it might be nice to cover this behavior with some behat test, but we can't really do it without requiring userform and the userfrom block because that's the only block that can't be rendered inline.

GuySartorelli:
> we can't really do it without requiring userform and the userfrom block because that's the only block that can't be rendered inline.

I haven't done anything with heat tests yet so excuse my ignorance but I take it then that behat tests can't take advantage of classes that implement `TestOnly`?

emteknetnz (approved):

maxime-rainville:
> I haven't done anything with behat tests yet so excuse my ignorance but I take it then that behat tests can't take advantage of classes that implement TestOnly?

@GuySartorelli Exactly ... your behat test need to be run against an actual functional website. Occasionally, we'll install frameworktest so we can test specific features that are not directly used in our core modules.

emteknetnz:
@GuySartorelli example behat equivalent of a TestOnly page - https://github.com/silverstripe/silverstripe-frameworktest/blob/4/code/multitab-validation/MultiTabPage.php

PR #966, opened 2022-03-30
Add extension point onAfterElementalAreaCreation

armsofnoodle:
Add an extension point for after the Elemental Area has been created. The extension point can be used to do things such as add blocks to the Elemental Area by default when a page is created. It differs from the exisiting extension point `onAfterRequireDefaultElementalRecords` in that this new one is run every time the elemental area is created (e.g. when a new page is created) rather than when requireDefaultRecords is run during a dev/build.

### DIFF
--- a/src/Extensions/ElementalAreasExtension.php
+++ b/src/Extensions/ElementalAreasExtension.php
@@ -285,6 +285,8 @@ class ElementalAreasExtension extends DataExtension
                 $area->OwnerClassName = get_class($this->owner);
                 $area->write();
                 $this->owner->$areaID = $area->ID;
+
+                $this->owner->extend('onAfterElementalAreaCreation', $area, $eaRelationship);
             }
         }
         return $this->owner;


michalkleiner:
It's an enhancement so should be branched off and target `4` branch.

If you need add elements when a page is being created, can you use `onAfterWrite` on the page itself?
I recall we add several default blocks on page creation in one project and never needed a new extension point in Elemental, so just curious if there are other ways of achieving the same.

GuySartorelli:
> If you need to add elements when a page is being created, can you use onAfterWrite on the page itself?

That's the approach I'd expect for this use case, as well. I can't think of a time when this hook would be useful that `onAfterWrite()` with a `$this->isInDB()` or `$this->owner->isInDB()` condition won't do.

@armsofnoodle is there a situation where the above won't work that this extension hook covers?

armsofnoodle:
The initial thought was to go with `onBeforeWrite()` but given that would rely on the the order of all the extensions' `onBeforeWrite()` being called in the same order each time it was a bit too fragile (as the elemental area needs to be created before we try assign blocks to it). Using `onAfterWrite()` with a `$this->owner->isInDB()` condition doesn't work either because `$this->owner->isInDB()` will return true if the page already exists and false if it doesn't. If the page doesn't exist then we can't add blocks to it, and if it does exist we don't necessarily want to add blocks to it because this being after `write()` that will add blocks every time we save the page.

I'm happy to investigate other routes if you think this still isn't necessarily required, if not though I'll update the branching and targeting 😃

GuySartorelli:
> Using `onAfterWrite()` with a `$this->owner->isInDB()` condition doesn't work either because `$this->owner->isInDB()` will return true if the page already exists and false if it doesn't. If the page doesn't exist then we can't add blocks to it

After looking at the order of operations, it looks like `onAfterWrite()` wouldn't be called until there is a DB record anyway, so that should always return true - my mistake, I misremembered that.

You could however use `$this->owner->isChanged('ID')` (only true the first time the object is written, unless you're doing some real weird stuff), or add your own flag in there - e.g. a `$db` field that is false by default, and you can set true the first time `onAfterWrite()` is called and you add your blocks in. Or if those blocks should _always_ exist, check if the blocks are in the page's `ElementalArea` and if not, add them.

Basically I think if there's a clean way you can achieve this through `onAfterWrite()`, that's probably preferable.

armsofnoodle:
Thanks for the suggestions, we were able to implement it using `onAfterWrite()`. Happy to close this PR 😄

chrispenny:
@GuySartorelli @armsofnoodle I would like to reopen this one and push for the extension point to be added.

`onAfterWrite()` does keep track of the changed state of the `ID`, however, it doesn't pair well with publishing actions. Take this example.

```php
<?php

namespace App\Models;

use SilverStripe\Dev\Debug;
use SilverStripe\Versioned\Versioned;

/**
 * @property string $Title
 * @mixin Versioned
 */
class MyModel extends DataObject
{
    private static string $table_name = 'MyModel';

    private static array $db = [
        'Title' => 'Varchar(255)',
    ];

    private static array $extensions = [
        Versioned::class,
    ];

    protected function onAfterWrite()
    {
        parent::onAfterWrite();

        Debug::dump('onAfterWrite: ' . ($this->isChanged('ID') ? 'changed' : 'not changed'));
    }
}
```

And I can test what happens quite simply here:
```php
<?php

namespace App\Tests\Models;

use App\Models\MyModel;
use SilverStripe\Dev\SapphireTest;

class MyModelTest extends SapphireTest
{
    public function testOnAfterWrite(): void
    {
        $model = MyModel::create();
        $model->Title = 'Starting Title';
        $model->write();
        $model->publishRecursive();
    }
}
```

The output is:
```
onAfterWrite: changed
=============================
onAfterWrite: changed
```

Basically... `onAfterWrite()` is also triggered when the model is published, and at that time it still believes that the `ID` has been changed. The result is that we end up with one set of Blocks being created when the Page is initially written, and then a second set of Blocks whenever the page is eventually published.

I know there would be ways around this (like adding additional logic to `onBeforePublish` to track when `onAfterWrite()` is called within that context), but I think that having to keep track of the context of the `onAfterWrite()` is a pain, and I think that this simple extension point would be a much nicer solution.

We're just asking to please have some level of access to the `ElementalArea` at the point when it is created.

GuySartorelli:
@michalkleiner What are your thoughts on that? I think it's a fair point but I also wonder if it wouldn't be better resolved by adding a more consistent way to identify in `onAfterWrite()` whether this is a write or a publish or a write-and-publish - or something like that?

chrispenny:
Hi @kinglozzer (CC: @armsofnoodle @GuySartorelli), thanks for jumping in! That looks to work!

I must admit that I don't fully understand what's going on though. Are we saying that the `ID` is sometimes stored as a `string` rather than an `int`, and so the default setting of `DataObject::CHANGE_STRICT` is seeing it as "changed"? That's definitely a [gotcha] that I was not aware of with our `IDs`.

kinglozzer:
> Are we saying that the ID is sometimes stored as a string rather than an int, and so the default setting of DataObject::CHANGE_STRICT is seeing it as "changed"?

Yep, exactly, though I must admit I thought we had solved it in core. Out of curiosity, are you using `MySQLDatabase` or `MySQLPDODatabase`? It might be worth trying switching between them to see if it’s driver-specific (from memory, PDO caused more issues than it solved, might be completely unrelated to this though!)

chrispenny:
We're using `MySQLDatabase`.

PR #968, opened 2022-04-04
HoverBar ID bugfixing

### LINKED ISSUE #926: Elemental 4.6 broken on pages with multiple ElementalAreas
We're using multiple ElemetalAreas on pages. Everything worked fine until we upgraded SS to 4.8 and Elemental to 4.6 (we were using SS 4.6 and Elemental 4.4 before).
Problem occurs when multiple Elemental Areas on single page contain at least one Elemental and you try to use "insert here" link (blue-lined plus) and you try to add new block to the top of the list.

Video:
Video of error

Test case:

1) install
`composer create-project silverstripe/installer ./ss4.elementals`

2) add elemental
`composer require dnadesign/silverstripe-elemental 4.6`

3) add extension `app/_config/elemental.yml`:
```yml
Page: 
  extensions: 
    - DNADesign\Elemental\Extensions\ElementalAreasExtension
```

4) edit app/src/Page.php:
```php
<?php

namespace {

    class Page extends SiteTree
    {
            
        private static $has_one = [
            'BlocksMain'         => ElementalArea::class,
            'BlocksRight'        => ElementalArea::class,
            'BlocksFooter'       => ElementalArea::class,
        ];
        
    }
}
```

5) build database
`/dev/build?flush`

6) login to CMS and open/create new page

7) add new content block to first elemental area and then add new content block to second elemental area

8) try to add new block to first elemental area by using "insert here" option and add it to first place. You can see both "insert here" links are activated from first and second elemental areas and two popups are open. Selecting Element type would (in most cases) create new element in second elemental area although you were trying to add one to the first block.

## Acceptance criteria
- [ ] When you use the hover bar to add a block, it adds it to the right tot he right Elemental Area, even when there's many areas on the same page.
- [ ] The position of the new block matches where you've clicked in the Elemental Area.

### Pull request
[FIX Hoverbar ID made uniq  -](https://github.com/silverstripe/silverstripe-elemental/pull/969) https://github.com/silverstripe/silverstripe-elemental/pull/969

#### Travis broken build
https://app.travis-ci.com/github/silverstripe/silverstripe-elemental/jobs/566559693

sabina-talipova:
### Parent issue
- fixes #926

### Description
Make HoverBarComponent attribute ID uniq by adding Elemental Area ID as suffix.
Add tests for ID HoverBarComponent


client/src/components/ElementEditor/ElementList.js
```diff
@@ -74,9 +74,9 @@ class ElementList extends Component {
     if (!isDraggingOver) {
       output = [
         <HoverBarComponent
-          key={0}
+          key={`0_${areaId}`} // Make key uniq
```
maxime-rainville, line 77: We need a key here because the element is in an array. The key just needs to be unique within that array ... it doesn't need to be unique within the page.

Although it doesn't hurt for it to be unique with the page.

src/GraphQL/Resolvers/Resolver.php
```diff
@@ -59,7 +59,8 @@ public static function resolveAddElementToArea(
     ): BaseElement {
         $elementClass = $args['className'];
         $elementalAreaID = $args['elementalAreaID'];
-        $afterElementID = $args['afterElementID'] ?? null;
+        // Remove areaaID suffix from 0_${areaaID} value and leave as it is ${elementID} value. We use it in line 99
+        $afterElementID = explode('_',$args['afterElementID'])[0] ?? null;
```
maxime-rainville, line 63: I really don't like this. We are changing the data that is being sent to the server in order to fix a UI bug.

maxime-rainville (changes requested):
While this fix works, I don't think it's the most elegant way of solving this issue.

The fundamental problem here is that when you click on a `<HoverBarComponent />` it will open the wrong `<AddElementPopoverComponent />` because many of those will have the same ID.

The proper way to fix this is to update this code snippet https://github.com/silverstripe/silverstripe-elemental/blob/3fe3c410c29952f6e319a820c3c9f8350a7a51a4/client/src/components/ElementEditor/HoverBar.js#L14-L50

These changes fixed it for me without the need to update the data that's being sent to the server. It's probably worth spending some time changing the way we generate the id, target and container props to avoid some double handling.

```diff
/**
 * Render an hoverbar without any state
 */
function StatelessHoverBar({
  AddElementPopoverComponent,
  elementTypes,
  elementId,
  areaId,
  popoverOpen,
  onToggle }) {
  const lineClasses = `${classNames('-line')} font-icon-plus-circled`;
  const label = i18n._t('ElementAddNewButton.ADD_BLOCK', 'Add block');
  const btnProps = {
    className: classNames('-area', { '-area--focus': popoverOpen }),
    onClick: onToggle,
    'aria-label': label,
    title: label,
-    id: `AddBlockHoverBarArea_${elementId}`
+    id: `AddBlockHoverBarArea_${areaId}_${elementId}`
  };

  return (
-    <div className={classNames('')} id={`AddBlockHoverBar_${elementId}`}>
+    <div className={classNames('')} id={`AddBlockHoverBar_${areaId}_${elementId}`}>
      <button {...btnProps}>
        <span className={classNames('-area-inner')}>
          <span className={lineClasses} />
        </span>
      </button>
      <AddElementPopoverComponent
        placement="bottom"
-        target={`AddBlockHoverBarArea_${elementId}`}
+        target={`AddBlockHoverBarArea_${areaId}_${elementId}`}
        isOpen={popoverOpen}
        elementTypes={elementTypes}
        toggle={onToggle}
-        container={`#AddBlockHoverBar_${elementId}`}
+        container={`#AddBlockHoverBar_${areaId}_${elementId}`}
        areaId={areaId}
        insertAfterElement={elementId}
      />
    </div>
  );
}
```

src/GraphQL/Resolvers/Resolver.php
GuySartorelli: A small note also about the comment - it's best to avoid mentioning line numbers in a comment, because as soon as someone adds or removes a line of code that comment will no longer be correct.

sabina-talipova: DONE

client/src/components/ElementEditor/ElementList.js
sabina-talipova: DONE

PR #973, opened 2022-04-12
MNT Compatibility with webonyx/graphql-php v14

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10250

### DIFF
--- a/tests/GraphQL/Legacy/SortBlockMutationCreatorTest.php
+++ b/tests/GraphQL/Legacy/SortBlockMutationCreatorTest.php
@@ -3,8 +3,8 @@
 namespace DNADesign\Elemental\Tests\Legacy\GraphQL;
 
 use DNADesign\Elemental\GraphQL\SortBlockMutationCreator;
+use DNADesign\Elemental\Tests\GraphQL\FakeResolveInfo;
 use DNADesign\Elemental\Tests\Src\TestElement;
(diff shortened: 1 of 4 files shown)


tests/GraphQL/Legacy/SortBlockMutationCreatorTest.php
```diff
@@ -3,6 +3,7 @@
 namespace DNADesign\Elemental\Tests\Legacy\GraphQL;
 
 use DNADesign\Elemental\GraphQL\SortBlockMutationCreator;
+use DNADesign\Elemental\Tests\GraphQL\FakeResolveInfo;
 use DNADesign\Elemental\Tests\Src\TestElement;
 use GraphQL\Type\Definition\ResolveInfo;
```
GuySartorelli, line 8:
```suggestion
```

GuySartorelli:
In general I'm happy with this - there's just the outstanding question of whether to use the `FakeResolveInfo` class which is newly declared in silverstripe/silverstripe-graphql instead of the one in this module. I can't see any reason to keep the one from this module.

emteknetnz:
I don't think we should refactor anymore than the absolute minimum to get this working given where graphql3 is in its lifecycle.

GuySartorelli (approved):
That's fair. I'm happy with this then.

PR #975, opened 2022-04-13
FIX getOwnerPage() should respect Versioned state.

wilr:
Previous to this getOwnerPage() would always return the Draft versioned
record. This would lead to cases were if a parent page contained draft
changes, unauthenicated users would be blocked from viewing content
blocks as `$this->getOwnerPage()->canView($member);` would fail since
logged out users could not view pages with draft changes.

### DIFF
--- a/src/Models/ElementalArea.php
+++ b/src/Models/ElementalArea.php
@@ -112,7 +112,9 @@ class ElementalArea extends DataObject
      */
     public function setOwnerPageCached(DataObject $page)
     {
-        $this->cacheData['owner_page'] = $page;
+        $cacheKey = 'owner_page_'. Versioned::get_reading_mode();
+
+        $this->cacheData[$cacheKey] = $page;
 
         return $this;
     }
@@ -193,8 +195,10 @@ class ElementalArea extends DataObject
         }
 
         // Allow for repeated calls to read from cache
-        if (isset($this->cacheData['owner_page'])) {
-            return $this->cacheData['owner_page'];
+        $cacheKey = 'owner_page_'. Versioned::get_reading_mode();
+
+        if (isset($this->cacheData[$cacheKey])) {
+            return $this->cacheData[$cacheKey];
         }
 
         if ($this->OwnerClassName && ClassInfo::exists($this->OwnerClassName)) {
@@ -217,6 +221,7 @@ class ElementalArea extends DataObject
 
                 if ($page) {
                     $this->setOwnerPageCached($page);
+
                     return $page;
                 }
             }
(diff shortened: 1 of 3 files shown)


kinglozzer (approved):

src/Models/ElementalArea.php
```diff
@@ -112,7 +112,9 @@ public function setElementsCached(ArrayList $elements)
      */
     public function setOwnerPageCached(DataObject $page)
     {
-        $this->cacheData['owner_page'] = $page;
+        $cacheKey = 'owner_page_'. Versioned::get_reading_mode();
+
+        $this->cacheData[$cacheKey] = $page;
```
GuySartorelli, line 117:
```suggestion
        $this->cacheData['owner_page_' . Versioned::get_reading_mode()] = $page;
```

src/Models/ElementalArea.php
```diff
@@ -193,8 +195,10 @@ public function getOwnerPage()
         }
 
         // Allow for repeated calls to read from cache
-        if (isset($this->cacheData['owner_page'])) {
-            return $this->cacheData['owner_page'];
+        $cacheKey = 'owner_page_'. Versioned::get_reading_mode();
+
```
GuySartorelli, line 199:
```suggestion
```

GuySartorelli (dismissed):

src/Models/ElementalArea.php
michalkleiner: I would leave it as it was for consistency with the changes below where it uses the variable in the same format.

src/Models/ElementalArea.php
michalkleiner: I think an empty line here is ok, I personally dislike variable assignments glued together to code blocks below.

GuySartorelli: I like them when the assignment is explicitly there to be used in the code block - ties it all together. But I'm okay keeping it there if the other change isn't being made. Bit of a (subjective) nitpick on my part.

src/Models/ElementalArea.php
GuySartorelli: The only reason it's assigned as a variable below is because the key is used for a conditional check before being used in the array. That said this is such a small thing, I'm happy to be overruled in this case.

GuySartorelli (approved):
Confirmed the new test fails without the other changes in this PR. Looks like a good change.

LABCAT:
Hi there,

Just wondering when this will be included in a tagged version of 4.x branch? 

We have a site that often experiences this issue so would really like to resolve this issue.

GuySartorelli:
There's nothing stopping this from being released as a patch other than that it was targetting `4`. I'll backport this to `4.9`

Issue #871, opened 2021-02-01
Elemental Area can be added to a non-Page DataObject and edited in a ModelAdmin


maxime-rainville:
[`BaseElement::CMSEditLink`](https://github.com/dnadesign/silverstripe-elemental/blob/4/src/Models/BaseElement.php#L707) assumes that blocks are attached to an Object with a `CMSEditLink` wethod. So when an `ElementalArea` is added to a `DataObject` edited via a ModelAdmin, you'll get this error `Object->__call(): the method 'CMSEditLink' does not exist on 'SilverStripe\\FrameworkTest\\Model\\Company'`

# Acceptance Criteria
- [x] Elemental Area can be added to DataObjects that are not sub classes of SiteTree
- [x] Proper guidance is given to developers on how to add an Elemental Area to a plain DataObject
- [x] Either developers have to define their own `CMSEditLink` method or we come with a way to "guess" the edit link

# Stretch goal
- Create a generic reusable API to define what ModelAdmin should be use to edit a DataObject and create the CMSEditLink from that.

# Notes
There was some recent work to allow previews to be added to ModelAdmin. Ideally, tho solution to this card should work with tho preview as well.

## PRs
- silverstripe/silverstripe-elemental#978
- #982
- https://github.com/silverstripe/silverstripe-frameworktest/pull/105

maxime-rainville:
Off the top of my head, I'm thinking we could
- update `BaseElement::CMSEditLink` to return `null` when the `ElementalArea is attached to an object without a `CMSEditLink` method.
- Implement a way for a DataObject to explicitly say, "go through this ModelAdmin to edit me"

brynwhyman:
Related/ duplicate: https://github.com/silverstripe/silverstripe-elemental/issues/718

GuySartorelli:
Possibly related: #965

sabina-talipova:
I suggest to add a few changes in the CMSEditLink() method in class BaseElement class. 
```
        if (!$page instanceof SiteTree) { 
            if (method_exists($page, 'CMSEditLink')) {
                $link = Controller::join_links($page->CMSEditLink(), 'ItemEditForm');
            } else {
                $link = null;
                $this->extend('updateCMSEditLink', $link);
                return $link;
            }
        } else {
            $link = $page->CMSEditLink();
        }

```

sabina-talipova:
[...]

PR #978, opened 2022-04-21
FIX / DOC Add support ElementalArea in DataObject

sabina-talipova:
## Description
Create new section in docs directory with example how to use ElementalArea together with DataObject.

## Changes
New condition was added in CMSEditLink() in BaseElement class for cause when $page is not an instance of SiteTree and Element is not inline editable. Also new condition for checking of existing the CMSEditLink() method in class that has relation with ElemantalArea.

## Parent Issue
- https://github.com/silverstripe/silverstripe-elemental/issues/871

### DIFF
--- a/docs/en/examples/elementalarea_with_dataobject.md
+++ b/docs/en/examples/elementalarea_with_dataobject.md
@@ -1,0 +1,111 @@
+# Using ElementalArea with DataObject
+
+## Creating Model and adding ElementalArea
+
+In this example the class `BlogPost` will be created, it will extend the `DataObject` class. Also a new `BlogPost` Admin section will be created in the Admin panel where CMS users can manage the `BlogPost` objects.
+
+Let's look at an example:
+
+**app/src/Admins/BlogPostsAdmin.php**
+
+```php
+<?php
+
+namespace App\Admins;
+
+use App\Models\BlogPost;
+use SilverStripe\Admin\ModelAdmin;
+
+class BlogPostsAdmin extends ModelAdmin
+{
+    private static string $url_segment = 'blog-posts-admin';
+
+    private static string $menu_title = 'Blog Posts';
+
+    private static string $menu_icon_class = 'font-icon-block-banner';
+
+    private static array $managed_models = [
+        BlogPost::class,
+    ];
+}
+```
+
+`$managed_models` - this static variable connects `BlogPostAdmin` with `BlogPost`, which will be created in the following code.\
+When you create a `DataObject` you should define the connection between your `DataObject` and `ElementalArea` by adding `$has_one` relationship.\
+In this example `BlogPost` will have one `ElementalArea`.
+
+**app/src/Models/BlogPost.php**
+
+```php
+<?php
+
+namespace App\Models;
+
+use App\Admins\BlogPostsAdmin;
+use DNADesign\Elemental\Models\ElementalArea;
+use SilverStripe\Control\Controller;
+use SilverStripe\ORM\DataObject;
+use SilverStripe\Versioned\Versioned;
+
+class BlogPost extends DataObject
+{
+    private static string $table_name = 'BlogPost';
+
+    private static $has_one = [
+        'ElementalArea' => ElementalArea::class,
+    ];
+
+    private static $owns = ['ElementalArea'];
+
+    // ...
+}
+ ```
+
+If you are using `ElementalArea` together with `DataObject`, it is important to define the `CMSEditLink()` method in the class.\
+`BaseElement::CMSEditLink()` relies on a valid CMS link being available in the parent `DataObject` - in this case, `BlogPost`. It is used to navigate directly to the editing section of the particular element.\
+***Note:*** For nested `GridField`s this method can get more complicated. Similarly, if your class is used in multiple admins, you will have to choose one to be the canonical admin section for the purposes of this method. This example only shows the simplest case, where the `BlogPost` class is used directly in `BlogPostsAdmin`.
+
+```php
+// ...
+
+class BlogPost extends DataObject
+{
+    // ...
+
+    public function CMSEditLink()
+    {
+        // In this example we use BlogPostsAdmin class as Controller
+        $admin = BlogPostsAdmin::singleton();
+
+        // Makes link more readable. Instead App\Models\ BlogPost, we get App-Models-BlogPost
+        $sanitisedClassname = str_replace('\\', '-', $this->ClassName);
+
+        // Returns link to editing section with elements
+        return Controller::join_links(
+            $admin->Link($sanitisedClassname),
+            'EditForm/field/',
+            $sanitisedClassname,
+            'item',
+            $this->ID,
+        );
+    }
+
+    // ...
+}
+
+```
+
+And finally, add `ElementalAreasExtension` to the `DataObject`
+
+**app/_config/elemental.yml**
+
+```yml
+App\Models\BlogPost:
+  extensions: 
+    - DNADesign\Elemental\Extensions\ElementalAreasExtension
+```
+
+## Related Documentation
+
+* [Preview](https://docs.silverstripe.org/en/4/developer_guides/customising_the_admin_interface/preview/)
+(note that this requires at least `silverstripe/admin` 4.11.0 and `dnadesign/silverstripe-elemental` 4.9.0)
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -243,13 +243,18 @@ class BaseElement extends DataObject
     public function canDelete($member = null)
     {
         $extended = $this->extendedCan(__FUNCTION__, $member);
+        
         if ($extended !== null) {
             return $extended;
         }
 
         if ($this->hasMethod('getPage')) {
             if ($page = $this->getPage()) {
-                return $page->canArchive($member);
+                if ($page->hasExtension(Versioned::class)) {
+                    return $page->canArchive($member);
+                } else {
+                    return $page->canDelete($member);
+                }
             }
         }
 
@@ -568,7 +573,8 @@ JS
     }
 
     /**
-     * @return null|SiteTree
+     * Despite the name of the method, getPage can return any type of DataObject
+     * @return null|DataObject
      * @throws \Psr\Container\NotFoundExceptionInterface
      * @throws \SilverStripe\ORM\ValidationException
      */
@@ -578,7 +584,6 @@ JS
         if (isset($this->cacheData['page'])) {
             return $this->cacheData['page'];
         }
-
 
         $class = DataObject::getSchema()->hasOneComponent($this, 'Parent');
         $area = ($this->ParentID) ? DataObject::get_by_id($class, $this->ParentID) : null;
@@ -638,9 +643,10 @@ JS
      */
     public function AbsoluteLink($action = null)
     {
-        if ($page = $this->getPage()) {
+        $page = $this->getPage();
+        
+        if ($page && ClassInfo::hasMethod($page, 'AbsoluteLink')) {
             $link = $page->AbsoluteLink($action) . '#' . $this->getAnchor();
-
             $this->extend('updateAbsoluteLink', $link);
 
             return $link;
@@ -657,9 +663,10 @@ JS
      */
     public function Link($action = null)
     {
-        if ($page = $this->getPage()) {
-            $link = $page->Link($action) . '#' . $this->getAnchor();
+        $page = $this->getPage();
 
+        if ($page && ClassInfo::hasMethod($page, 'Link')) {
+            $link = $page->Link($action) . '#' . $this->getAnchor();
             $this->extend('updateLink', $link);
 
             return $link;
@@ -712,35 +719,62 @@ JS
             return $this->cacheData['cms_edit_link'];
         }
 
+        $link = $this->getElementCMSLink($directLink);
+        $this->extend('updateCMSEditLink', $link);
+
+        if ($link) {
+            $this->cacheData['cms_edit_link'] = $link;
+        }
+
+        return $link;
+    }
+
+    /**
+     * @param bool $directLink
+     * @return null|string
+     */
+    private function getElementCMSLink(bool $directLink)
+    {
         $relationName = $this->getAreaRelationName();
         $page = $this->getPage();
 
+        $link = null;
+
         if (!$page) {
-            $link = null;
-            $this->extend('updateCMSEditLink', $link);
             return $link;
         }
 
-        if (!$page instanceof SiteTree && method_exists($page, 'CMSEditLink')) {
-            $link = Controller::join_links($page->CMSEditLink(), 'ItemEditForm');
-        } else {
+        if ($page instanceof SiteTree) {
             $link = $page->CMSEditLink();
+        } elseif (ClassInfo::hasMethod($page, 'CMSEditLink')) {
+            $link = Controller::join_links($page->CMSEditLink(), 'ItemEditForm');
         }
-
-        // In-line editable blocks should just take you to the page. Editable ones should add the suffix for detail form
+        // In-line editable blocks should just take you to the page.
+        // Editable ones should add the suffix for detail form.
         if (!$this->inlineEditable() || $directLink) {
-            $link = Controller::join_links(
-                singleton(CMSPageEditController::class)->Link('EditForm'),
-                $page->ID,
-                'field/' . $relationName . '/item/',
-                $this->ID,
-                'edit'
-            );
+            if ($page instanceof SiteTree) {
+                return Controller::join_links(
+                    singleton(CMSPageEditController::class)->Link('EditForm'),
+                    $page->ID,
+                    'field',
+                    $relationName,
+                    'item',
+                    $this->ID,
+                    'edit'
+                );
+            } else {
+                // If $page is not a Page, then generate $link base on $page->CMSEditLink()
+                return Controller::join_links(
+                    $link,
+                    'field',
+                    $relationName,
+                    'item',
+                    $this->ID,
+                    'edit'
+                );
+            }
         }
 
-        $this->extend('updateCMSEditLink', $link);
-
-        $this->cacheData['cms_edit_link'] = $link;
         return $link;
     }
 
--- a/tests/BaseElementTest.php
+++ b/tests/BaseElementTest.php
@@ -7,9 +7,13 @@ use DNADesign\Elemental\Extensions\ElementalPageExtension;
 use DNADesign\Elemental\Models\BaseElement;
 use DNADesign\Elemental\Models\ElementalArea;
 use DNADesign\Elemental\Models\ElementContent;
+use DNADesign\Elemental\Tests\Src\TestDataObject;
 use DNADesign\Elemental\Tests\Src\TestElement;
+use DNADesign\Elemental\Tests\Src\TestElementDataObject;
+use DNADesign\Elemental\Tests\Src\TestDataObjectWithCMSEditLink;
 use DNADesign\Elemental\Tests\Src\TestPage;
 use Page;
+use ReflectionClass;
 use SilverStripe\Control\Director;
 use SilverStripe\Core\Config\Config;
 use SilverStripe\Dev\FunctionalTest;
@@ -18,10 +22,19 @@ use SilverStripe\VersionedAdmin\Forms\HistoryViewerField;
 
 class BaseElementTest extends FunctionalTest
 {
-    protected static $fixture_file = 'ElementalPageExtensionTest.yml';
+    protected static $fixture_file = [
+        'ElementalPageExtensionTest.yml',
+        'ElementalAreaDataObjectTest.yml'
+    ];
 
     protected static $required_extensions = [
         Page::class => [
+            ElementalPageExtension::class,
+        ],
+        TestDataObject::class => [
+            ElementalPageExtension::class,
+        ],
+        TestDataObjectWithCMSEditLink::class => [
             ElementalPageExtension::class,
         ],
     ];
@@ -29,6 +42,9 @@ class BaseElementTest extends FunctionalTest
     protected static $extra_dataobjects = [
         TestPage::class,
         TestElement::class,
+        TestDataObject::class,
+        TestDataObjectWithCMSEditLink::class,
+        TestElementDataObject::class,
     ];
 
     public function testSimpleClassName()
@@ -235,5 +251,136 @@ class BaseElementTest extends FunctionalTest
         $element1->write();
 
         $this->assertEquals(0, (int) $element1->Sort);
+    }
+
+    public function getElementCMSLinkDataProvider()
+    {
+        return [
+            // Element in DataObject with $directLink === true
+            'element1' => [
+                TestElement::class,
+                'elementDataObject1',
+                'http://localhost/admin/1/ItemEditForm/field/ElementalArea/item/',
+                true
+            ],
+            // Element in DataObject with $inline_editable = false
+            'element2' => [
+                TestElementDataObject::class,
+                'testElementDataObject1',
+                'http://localhost/admin/1/ItemEditForm/field/ElementalArea/item/',
+            ],
+            // Element in DataObject with $inline_editable = true
+            'element3' => [
+                ElementContent::class,
+                'contentDataObject1',
+                'http://localhost/admin/1/ItemEditForm',
+            ],
+            // Element in Page with $inline_editable = true
+            'element4' => [
+                ElementContent::class,
+                'content1',
+                'http://localhost/admin/pages/edit/show/1',
+            ],
+            // Element in DataObject with $directLink === true
+            'element5' => [
+                ElementContent::class,
+                'content1',
+                'admin/pages/edit/EditForm/1/field/ElementalArea/item/1/edit',
+                true
+            ],
+            // DataObject without CMSEditLink method
+            'element6' => [
+                TestElement::class,
+                'elementDataObject2',
+                null
+            ],
+        ];
+    }
+
+    /**
+     * @dataProvider getElementCMSLinkDataProvider
+     */
+    public function testCMSEditLink(string $class, string $element, ?string $link, bool $directLink = false)
+    {
+        $object = $this->objFromFixture($class, $element);
+        $editLink = $object->CMSEditLink($directLink);
+
+        if ($link) {
+            $this->assertStringContainsString($link, $editLink);
+        } else {
+            $this->assertNull($link);
+        }
+    }
+
+    public function canDeleteProvider()
+    {
+        return [
+            // Element on Page
+            'element1' => [
+                ElementContent::class,
+                'content1',
+                'ADMIN',
+            ],
+            // Element in DataObject
+            'element2' => [
+                TestElement::class,
+                'elementDataObject1',
+                'CMS_ACCESS_CMSMain',
+            ],
+        ];
+    }
+
+    /**
+     * @dataProvider canDeleteProvider
+     */
+    public function testCanDelete(
+        string $class,
+        string $element,
+        string $permission
+    ) {
+        $this->logOut();
+        $this->logInWithPermission($permission);
+        $object = $this->objFromFixture($class, $element);
+        $canDelete = $object->canDelete();
+        $this->assertNotNull($canDelete);
+        $this->assertTrue($canDelete);
+    }
+
+    public function linksProvider()
+    {
+        return [
+            // Element on Page
+            'element1' => [
+                ElementContent::class,
+                'content1',
+                '/test-elemental/#e1',
+            ],
+            // Element in DataObject
+            'element2' => [
+                TestElement::class,
+                'elementDataObject1',
+                null,
+            ],
+        ];
+    }
+
+    /**
+     * @dataProvider linksProvider
+     */
+    public function testLinkWithDataObject(string $class, string $element, ?string $link)
+    {
+        $object = $this->objFromFixture($class, $element);
+        $this->assertEquals($link, $object->Link());
+    }
+
+    /**
+     * @dataProvider linksProvider
+     */
+    public function testAbsoluteLink(string $class, string $element, ?string $link)
+    {
+        $link = $link ? Director::absoluteURL($link) : $link;
+        $object = $this->objFromFixture($class, $element);
+        $absoluteLink = $object->AbsoluteLink();
+        $this->assertEquals($link, $absoluteLink);
     }
 }
--- a/tests/Behat/features/add-block-element-to-data-object.feature
+++ b/tests/Behat/features/add-block-element-to-data-object.feature
@@ -1,0 +1,44 @@
+@javascript
+Feature: Add elements in the CMS DataObject
+  As a CMS user
+  I want to add elements in the CMS DataObject
+  So that I can use multiple elements in a Dataobject
+  
+  Background:
+    Given I am logged in with "ADMIN" permissions
+
+  Scenario: I can add inline-editable elements to the DataObject
+    When I go to "/admin/elemental-behat-test-admin/"
+    Then I press the "Add Elemental Behat Test Object" button
+    Then I press the "Create" button
+    Then I wait 1 second
+    Then I press the "Add block" button
+    Then I press the "Content" button in the add block popover
+    Then I should see "Untitled Content block" as the title for block 1
+    When I click on block 1
+    And I should see "Title"
+    And I should see "Content"
+    And I fill in "Elemental Behat Test Title" for "Title" for block 1
+    Then I press the "Save" button
+    Then I should see "Elemental Behat Test Title" as the title for block 1
+
+  Scenario: I can add non-inline-editable elements to the DataObject
+    Given content blocks are not in-line editable
+        When I go to "/admin/elemental-behat-test-admin/"
+        Then I press the "Add Elemental Behat Test Object" button
+        Then I press the "Create" button
+        Then I wait 1 second
+        Then I press the "Add block" button
+        Then I press the "Content" button in the add block popover
+        When I click on block 1
+        And I should see "Title"
+        And I should see "Content"
+
+    Given I fill in "New Elemental Behat Test Title" for "Title"
+        And I fill in "<p>New sample content</p>" for the "Content" HTML field
+        And I press the "Publish" button
+        Then I should see a "Published content block" message
+        Then I should see "New Elemental Behat Test Title"
+        And the "Content" field should contain "<p>New sample content</p>"
+        Then I press the "Navigate up a folder" button
+        Then I should see "New Elemental Behat Test Title" as the title for block 1
--- a/tests/Src/TestDataObject.php
+++ b/tests/Src/TestDataObject.php
@@ -1,0 +1,27 @@
+<?php
+
+namespace DNADesign\Elemental\Tests\Src;
+
+use DNADesign\Elemental\Models\ElementalArea;
+use SilverStripe\Control\Controller;
+use SilverStripe\Control\Director;
+use SilverStripe\ORM\DataObject;
+use SilverStripe\Dev\TestOnly;
+
+class TestDataObject extends DataObject implements TestOnly
+{
+    private static $table_name = 'TestDataObject';
+
+    private static array $db = [
+        'Title' => 'Varchar(255)',
+        'Content' => 'HTMLText',
+    ];
+
+    private static $has_one = [
+        'ElementalArea' => ElementalArea::class,
+    ];
+
+    private static $owns = [
+        'ElementalArea',
+    ];
+}
--- a/tests/Src/TestDataObjectWithCMSEditLink.php
+++ b/tests/Src/TestDataObjectWithCMSEditLink.php
@@ -1,0 +1,45 @@
+<?php
+
+namespace DNADesign\Elemental\Tests\Src;
+
+use DNADesign\Elemental\Models\ElementalArea;
+use SilverStripe\Control\Controller;
+use SilverStripe\Control\Director;
+use SilverStripe\ORM\DataObject;
+use SilverStripe\Dev\TestOnly;
+use SilverStripe\Security\Permission;
+use SilverStripe\Security\Security;
+
+class TestDataObjectWithCMSEditLink extends DataObject implements TestOnly
+{
+    private static $table_name = 'TestDataObjectWithCMSEditLink';
+
+    private static array $db = [
+        'Title' => 'Varchar(255)',
+        'Content' => 'HTMLText',
+    ];
+
+    private static $has_one = [
+        'ElementalArea' => ElementalArea::class,
+    ];
+
+    private static $owns = [
+        'ElementalArea',
+    ];
+
+    public function CMSEditLink()
+    {
+        $link = Controller::join_links(
+            'admin/',
+            $this->ID,
+        );
+        return Director::absoluteURL($link);
+    }
+
+    public function canDelete($member = null)
+    {
+        $member = $member ? $member : Security::getCurrentUser();
+        $codes = ['CMS_ACCESS_CMSMain'];
+        return Permission::checkMember($member, $codes);
+    }
+}
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,8 @@
         "silverstripe/vendor-plugin": "^1"
     },
     "require-dev": {
-      "silverstripe/recipe-testing": "^2"
+      "silverstripe/recipe-testing": "^2",
+      "silverstripe/frameworktest": "4.x-dev"
     },
     "suggest": {
         "silverstripe/elemental-blocks": "Adds a set of common SilverStripe content block types",
--- a/tests/ElementalAreaDataObjectTest.yml
+++ b/tests/ElementalAreaDataObjectTest.yml
@@ -1,0 +1,38 @@
+DNADesign\Elemental\Models\ElementalArea:
+  areaDataObject1:
+    Title: Area 1
+    OwnerClassName: DNADesign\Elemental\Tests\Src\TestDataObjectWithCMSEditLink
+  areaDataObject2:
+    Title: Area 2
+    OwnerClassName: DNADesign\Elemental\Tests\Src\TestDataObject
+
+DNADesign\Elemental\Tests\Src\TestDataObjectWithCMSEditLink:
+  dataObject1:
+    Title: DataObject with CMSEditLink method 
+    ElementalAreaID: =>DNADesign\Elemental\Models\ElementalArea.areaDataObject1
+
+DNADesign\Elemental\Tests\Src\TestDataObject:
+  dataObject2:
+    Title: DataObject without CMSEditLink method
+    ElementalAreaID: =>DNADesign\Elemental\Models\ElementalArea.areaDataObject2
+
+DNADesign\Elemental\Tests\Src\TestElement:
+  elementDataObject1:
+    Title: Element 1
+    TestValue: 'Hello Test'
+    ParentID: =>DNADesign\Elemental\Models\ElementalArea.areaDataObject1
+  elementDataObject2:
+    Title: Element 2
+    TestValue: 'Hello Test'
+    ParentID: =>DNADesign\Elemental\Models\ElementalArea.areaDataObject2
+
+DNADesign\Elemental\Tests\Src\TestElementDataObject:
+  testElementDataObject1:
+    Title: Test Element inline-editable
+    TestValue: 'Hello Test'
+    ParentID: =>DNADesign\Elemental\Models\ElementalArea.areaDataObject1
+
+DNADesign\Elemental\Models\ElementContent:
+  contentDataObject1:
+    HTML: Some content
+    ParentID: =>DNADesign\Elemental\Models\ElementalArea.areaDataObject1
--- a/tests/Src/TestElementDataObject.php
+++ b/tests/Src/TestElementDataObject.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace DNADesign\Elemental\Tests\Src;
+
+use DNADesign\Elemental\Models\BaseElement;
+use SilverStripe\Dev\TestOnly;
+
+class TestElementDataObject extends BaseElement implements TestOnly
+{
+    private static $table_name = 'TestElementDataObject';
+
+    private static $db = [
+        'TestValue' => 'Text',
+    ];
+
+    private static bool $inline_editable = false;
+
+    public function getType()
+    {
+        return 'A test element in DataObject';
+    }
+}


docs/en/examples/using_elementalArea_with_dataObject.md
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    public function forTemplate($holder = true)
+    {
+        return $this->renderWith(['type' => 'Blogs', self::class]);
+    }
+}
+
+```
+Add template for DataObject to display data. In this example we create BlogPost.ss file in templates/App/Models/Blogs directory.
```
emteknetnz, line 230:
```suggestion
Add template for DataObject to display data. In this example we create a BlogPost.ss file.
```

docs/en/examples/using_elementalArea_with_dataObject.md
```diff
@@ -0,0 +1,245 @@
...
+
+**BlogPost.ss**
```
emteknetnz, line 232:
```suggestion
**templates/App/Models/BlogPost.ss**
```

docs/en/examples/using_elementalArea_with_dataObject.md
```diff
@@ -0,0 +1,245 @@
...
+        );
+    }
+
+    ...
+    
+}
+
+```
+
+Method Link() doesn't exist in DataObject class, so you should define this method in your class. Since PreviewLink() method of BaseElement uses Link() method to generate Preview window in the BlogPost Admin section. You should add canArchive() method in your class as well to set rules for canDalete() method.
```
emteknetnz, line 102:
```suggestion
Method Link() doesn't exist in DataObject class, so you should define this method in your class. Since PreviewLink() method of BaseElement uses Link() method to generate Preview window in the BlogPost Admin section.  Note: making a DataObject previewable requires silverstripe/admin 4.11 or later.

You should add canArchive() method in your class as well to set rules for canDelete() method.
```

emteknetnz (changes requested):
I wasted able to get the CMS preview to work.  Using the debugger, cmsPreview() nor PreviewLink were getting called. I followed the instructions inthe document I'll paste my code below.

emteknetnz:
Also could you rename `using_elementalArea_with_dataObject.md` to be lower case `using_elementalarea_with_dataobject.md` - you'll probably need to follow these [instructions](https://stackoverflow.com/a/20907647)

emteknetnz:
**app/_config/extensions.yml**
```yml
App\Models\BlogPost:
  extensions:
    - DNADesign\Elemental\Extensions\ElementalAreasExtension
```

**app/src/Admins/BlogPostsAdmin.php**
```php
<?php

namespace App\Admins;

use App\Models\BlogPost;
use SilverStripe\Admin\ModelAdmin;
use SilverStripe\View\SSViewer;

class BlogPostsAdmin extends ModelAdmin
{
    private static string $url_segment = 'blog-posts-admin';

    private static string $menu_title = 'Blog Posts';

    private static string $menu_icon_class = 'font-icon-block-banner';

    private static array $managed_models = [
        BlogPost::class,
    ];

    private static $allowed_actions = [
        'cmsPreview',
    ];

    private static $url_handlers = [
        '$ModelClass/cmsPreview/$ID' => 'cmsPreview',
    ];

    public function cmsPreview()
    {
        $id = $this->urlParams['ID'];
        $obj = $this->modelClass::get_by_id($id);
        if (!$obj || !$obj->exists()) {
            return $this->httpError(404);
        }
        // Include use of a front-end theme temporarily.
        $oldThemes = SSViewer::get_themes();
        SSViewer::set_themes(SSViewer::config()->get('themes'));
        $preview = $obj->forTemplate();

        // Make sure to set back to backend themes.
        SSViewer::set_themes($oldThemes);

        return $preview;
    }
}
```

**app/src/Models/BlogPost.php**
```php
<?php

namespace App\Models;

use App\Admins\BlogPostsAdmin;
use SilverStripe\Control\Controller;
use SilverStripe\ORM\CMSPreviewable;

class BlogPost extends DataObject implements CMSPreviewable
{

    private static string $table_name = 'BlogPost';

    private static array $db = [
        'Title' => 'Varchar(255)',
        'Content' => 'HTMLText',
    ];

    private static $has_one = [
        'ElementalArea' => ElementalArea::class,
    ];

    public function CMSEditLink()
    {
        // In this example we use BlogPostsAdmin class as Controller
        $admin = BlogPostsAdmin::singleton();
        // Makes link more readable. Instead App\Models\ BlogPost, we get App-Models-BlogPost
        $sanitisedClassname = str_replace('\\', '-', $this->ClassName);
        // Returns link to editing section with elements
        return Controller::join_links(
            $admin->Link($sanitisedClassname),
            'EditForm/field/',
            $sanitisedClassname,
            'item',
            $this->ID,
        );
    }

    public function Link($action = null)
    {
        $admin = BlogPostsAdmin::singleton();

        return $this->ID
            ? Controller::join_links(
                $admin->Link(str_replace('\\', '-', $this->ClassName)),
                $this->ID
                )
            : null;
    }

    public function canArchive($member = null)
    {
        return $this->canDelete($member);
    }

    public function AbsoluteLink($action = null)
    {
        return null;
    }

    public function PreviewLink($action = null)
    {
        $admin = BlogPostsAdmin::singleton();
        $link = $this->ID
            ? Controller::join_links(
                $admin->Link(str_replace('\\', '-', $this->ClassName)),
                'cmsPreview',
                $this->ID
                )
            : null;
        var_dump($link);
        return $link;
    }
    // The following methods are required to show data in template
    public function getMimeType()
    {
        return 'text/html';
    }
    public function forTemplate($holder = true)
    {
        //return $this->renderWith(self::class);
        return $this->renderWith(['type' => 'Blogs', self::class]);
        //return $this->renderWith('BlankPage');
    }
}
```

themes/simple/templates/App/Models/BlogPost.ss
```html
<% if $ElementalArea %>
    $ElementalArea
<% end_if %>
```

docs/en/examples/using_elementalArea_with_dataObject.md
```diff
@@ -0,0 +1,245 @@
+---
+title: Using ElementalArea with DataObject
+summary: Learn how to add ElementalArea to your models and manage it in ModalAdmin
```
GuySartorelli, line 3:
```suggestion
summary: Learn how to add an ElementalArea to your models and manage it in a ModelAdmin
```

docs/en/examples/using_elementalArea_with_dataObject.md
```diff
@@ -0,0 +1,245 @@
...
+
+    private static string $menu_icon_class = 'font-icon-block-banner';
+
+    private static array $managed_models = [
+        BlogPost::class,
+    ];
+}
+```
+
+``` $managed_models ``` - this static variable connects BlogPostAdmin with BlogPost, which will be created in the following code.
```
GuySartorelli, line 37:
```suggestion
`$managed_models` - this static variable connects BlogPostAdmin with BlogPost, which will be created in the following code.
```
Three backticks usually denote a block of code rather than an inline code reference.

docs/en/examples/using_elementalArea_with_dataObject.md
```diff
@@ -0,0 +1,245 @@
...
+When you create DataObject you should define connection between your DataObject and ElementalArea by adding ```$has_one``` relationship.
```
GuySartorelli, line 38:
```suggestion
When you create DataObject you should define connection between your `DataObject` and `ElementalArea` by adding `$has_one` relationship.
```

docs/en/examples/using_elementalArea_with_dataObject.md
GuySartorelli: Admin isn't up to 4 yet - that should read 1.11 :p

GuySartorelli (changes requested):

docs/en/examples/using_elementalArea_with_dataObject.md
sabina-talipova: The last version is 1.10. So it should be like "...requires silverstripe/admin 1.10 or later".

docs/en/examples/elementalarea_with_dataobject.md
```diff
@@ -0,0 +1,244 @@
...
+        );
+    }
+
+    // ...
+}
+
+```
+
+Method Link() doesn't exist in DataObject class, so you should define this method in your class. Since PreviewLink() method of BaseElement uses Link() method to generate Preview window in the BlogPost Admin section. Note: making a DataObject previewable requires silverstripe/admin 1.10 or later.
+You should add canArchive() method in your class as well to set rules for canDalete() method.
```
emteknetnz, line 106:
```suggestion
You should add canArchive() method in your class as well to set rules for canDelete() method.
```

docs/en/examples/elementalarea_with_dataobject.md
```diff
@@ -0,0 +1,244 @@
...
+            $this->ID,
+        );
+    }
+
+    // ...
+}
+
+```
+
+Method Link() doesn't exist in DataObject class, so you should define this method in your class. Since PreviewLink() method of BaseElement uses Link() method to generate Preview window in the BlogPost Admin section. Note: making a DataObject previewable requires silverstripe/admin 1.10 or later.
```
emteknetnz, line 105: Sorry, actually think it may be 1.11 - https://github.com/silverstripe/silverstripe-admin/pull/1250/commits/4603e3c35571485ec8c0bb18900985ff1eaeb7c4 isn't in 1.10

```suggestion
Method Link() doesn't exist in DataObject class, so you should define this method in your class. Since PreviewLink() method of BaseElement uses Link() method to generate Preview window in the BlogPost Admin section. Note: making a DataObject previewable requires silverstripe/admin 1.11 or later.
```

emteknetnz (changes requested):
Seems like these docs are duplication of what's already on framework - https://docs.silverstripe.org/en/4/developer_guides/customising_the_admin_interface/preview/ /  https://github.com/silverstripe/silverstripe-framework/blob/4/docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/04_Preview.md

We should aim to not have duplicated docs if possible - would it be possible to just remove the duplicated bits and link to the framework docs?

src/Models/BaseElement.php
```diff
@@ -721,21 +721,38 @@ public function CMSEditLink($directLink = false)
             return $link;
         }
 
-        if (!$page instanceof SiteTree && method_exists($page, 'CMSEditLink')) {
-            $link = Controller::join_links($page->CMSEditLink(), 'ItemEditForm');
+        if (!$page instanceof SiteTree) {
```
emteknetnz, line 724:
```suggestion
        if (!($page instanceof SiteTree)) {
```

src/Models/BaseElement.php
```diff
@@ -721,21 +721,38 @@ public function CMSEditLink($directLink = false)
...
         // In-line editable blocks should just take you to the page. Editable ones should add the suffix for detail form
         if (!$this->inlineEditable() || $directLink) {
-            $link = Controller::join_links(
-                singleton(CMSPageEditController::class)->Link('EditForm'),
-                $page->ID,
-                'field/' . $relationName . '/item/',
-                $this->ID,
-                'edit'
-            );
+            if (!$page instanceof SiteTree) {
```
emteknetnz, line 738:
```suggestion
            if (!($page instanceof SiteTree)) {
```

src/Models/BaseElement.php
```diff
@@ -721,21 +721,38 @@ public function CMSEditLink($directLink = false)
...
         }
 
-        if (!$page instanceof SiteTree && method_exists($page, 'CMSEditLink')) {
-            $link = Controller::join_links($page->CMSEditLink(), 'ItemEditForm');
+        if (!$page instanceof SiteTree) {
+            if (ClassInfo::hasMethod($page, 'CMSEditLink')) {
+                $link = Controller::join_links($page->CMSEditLink(), 'ItemEditForm');
+            } else {
+                $link = null;
+                $this->extend('updateCMSEditLink', $link);
```
emteknetnz, line 729: Shouldn't this extension hook still be be called even if `ClassInfo::hasMethod($page, 'CMSEditLink')` ?

emteknetnz (changes requested):

src/Models/BaseElement.php
sabina-talipova: Probably, not yet. Here we just generate only a some part of the link. Further we use this part to generate the rest part of the link and after that we use $this->extend('updateCMSEditLink', $link);

emteknetnz: seems like it should be 
```
$link = null;
if (ClassInfo::hasMethod($page, 'CMSEditLink')) {
    $link = Controller::join_links($page->CMSEditLink(), 'ItemEditForm');
}
$this->extend('updateCMSEditLink', $link);
```
Any custom code implementing the link can add a `if (is_null($link))` check and proceed from there.  It's entirely possible if $link is non-null they still need to update the link to something else

emteknetnz (changes requested):
Seems a lot better now that the Previewable examples have been removed.

I've followed the example code locally, the BlogPost's aren't publishable, which is unexpected given the content blocks are inline publishable

src/Models/BaseElement.php
emteknetnz: actually it's a bit confusing that $link can be set multiple times in this method, for instance

```php
        if (!($page instanceof SiteTree)) {
            if (ClassInfo::hasMethod($page, 'CMSEditLink')) {
                $link = Controller::join_links($page->CMSEditLink(), 'ItemEditForm');
            } else {
                $link = null;
                $this->extend('updateCMSEditLink', $link);
                return $link;
            }
        } else {
            $link = $page->CMSEditLink();
        }

        // In-line editable blocks should just take you to the page. Editable ones should add the suffix for detail form
        if (!$this->inlineEditable() || $directLink) {
            if (!($page instanceof SiteTree)) {
                $link = Controller::join_links(
                    $link,
                    'field',
                    $relationName,
                    'item',
                    $this->ID,
                    'edit'
                );
```

ideally we'd have something like 

```php
$link = null
if ($conditionA) {
  $link = ...
} elseif ($conditionB) {
  $link = ...
}
$this->extend('updateCMSEditLink', $link);
return $link;
```

src/Models/BaseElement.php
```diff
@@ -712,35 +712,65 @@ public function CMSEditLink($directLink = false)
...
-            );
-        }
+    /**
+     * 
+     * @param bool $directLink Indicates that the GridFieldDetailEdit
+     *             form link should be given even if the block can be edited in-line.
+     * @return null|string
+     * @throws \SilverStripe\ORM\ValidationException
+     */
+    public function getElementCMSLink($directLink = false)
```
emteknetnz, line 732:
```suggestion
    private function getElementCMSLink(bool $directLink)
```
Only add public/protected (new API) if you need to, otherwise use private.  Also don't use optional params unless you need to

src/Models/BaseElement.php
```diff
@@ -712,35 +712,65 @@ public function CMSEditLink($directLink = false)
...
-                $this->ID,
-                'edit'
-            );
-        }
+    /**
+     * 
+     * @param bool $directLink Indicates that the GridFieldDetailEdit
+     *             form link should be given even if the block can be edited in-line.
+     * @return null|string
+     * @throws \SilverStripe\ORM\ValidationException
```
emteknetnz, line 730: Where does this ValidationException come from?  If it's from one of the nested methods then it should be defined there but not here

src/Models/BaseElement.php
```diff
@@ -712,35 +712,65 @@ public function CMSEditLink($directLink = false)
...
+     */
+    public function getElementCMSLink($directLink = false)
+    {
+        $relationName = $this->getAreaRelationName();
+        $page = $this->getPage();
 
-        $this->extend('updateCMSEditLink', $link);
+        $link = null;
+
+        if ($page) {
```
emteknetnz, line 739:
```suggestion
        if (!$page) {
            return null;
        }
```
Do an early return and unindent the block below

src/Models/BaseElement.php
```diff
@@ -712,35 +712,65 @@ public function CMSEditLink($directLink = false)
...
+            // If $page is not a Page
+            if (!($page instanceof SiteTree)) {
```
emteknetnz, line 741: Add a comment to `function getPage()` mentioning that now, despite the name of the method, that this can return any type of DataObject, not just SiteTree

emteknetnz, line 741:
```suggestion
            if ($page instanceof SiteTree) {
```
And then reverse the blocks below

src/Models/BaseElement.php
```diff
@@ -712,35 +712,65 @@ public function CMSEditLink($directLink = false)
...
+     * 
+     * @param bool $directLink Indicates that the GridFieldDetailEdit
+     *             form link should be given even if the block can be edited in-line.
+     * @return null|string
+     * @throws \SilverStripe\ORM\ValidationException
+     */
+    public function getElementCMSLink($directLink = false)
+    {
+        $relationName = $this->getAreaRelationName();
+        $page = $this->getPage();
```
emteknetnz, line 735:
```suggestion
        $page = $this->getPage();
```

src/Models/BaseElement.php
```diff
@@ -712,35 +712,65 @@ public function CMSEditLink($directLink = false)
...
 
-        $this->extend('updateCMSEditLink', $link);
+        $link = null;
```
emteknetnz, line 737:
```suggestion
```

src/Models/BaseElement.php
```diff
@@ -712,35 +712,65 @@ public function CMSEditLink($directLink = false)
...
+
+        if ($page) {
+            // If $page is not a Page
+            if (!($page instanceof SiteTree)) {
+                // Envoke method only if it exists in the $page
+                if (ClassInfo::hasMethod($page, 'CMSEditLink')) {
+                    $link = Controller::join_links($page->CMSEditLink(), 'ItemEditForm');
```
emteknetnz, line 744:
```suggestion
                    return Controller::join_links($page->CMSEditLink(), 'ItemEditForm');
```

src/Models/BaseElement.php
```diff
@@ -712,35 +712,65 @@ public function CMSEditLink($directLink = false)
...
+                }
+            } else {
+                $link = $page->CMSEditLink();
```
emteknetnz, line 747:
```suggestion
                return $page->CMSEditLink();
```

src/Models/BaseElement.php
```diff
@@ -712,35 +712,65 @@ public function CMSEditLink($directLink = false)
...
+            }
+            // In-line editable blocks should just take you to the page.
+            // Editable ones should add the suffix for detail form.
+            if (!$this->inlineEditable() || $directLink) {
+                if ($page instanceof SiteTree) {
+                    $link = Controller::join_links(
```
emteknetnz, line 753:
```suggestion
                    return Controller::join_links(
```

src/Models/BaseElement.php
```diff
@@ -712,35 +712,65 @@ public function CMSEditLink($directLink = false)
...
+                        singleton(CMSPageEditController::class)->Link('EditForm'),
+                        $page->ID,
+                        'field/' . $relationName . '/item/',
+                        $this->ID,
+                        'edit'
+                    );
+                } else {
+                    // If $page is not a Page, then generate $link base on $page->CMSEditLink()
+                    $link = Controller::join_links(
```
emteknetnz, line 762:
```suggestion
                    return Controller::join_links(
```

src/Models/BaseElement.php
```diff
@@ -712,35 +712,65 @@ public function CMSEditLink($directLink = false)
...
+                        'item',
+                        $this->ID,
+                        'edit'
+                    );
+                }
+            }
+        }
 
-        $this->cacheData['cms_edit_link'] = $link;
         return $link;
```
emteknetnz, line 774:
```suggestion
        return null;
```

src/Models/BaseElement.php
```diff
@@ -712,35 +712,65 @@ public function CMSEditLink($directLink = false)
...
+                // Envoke method only if it exists in the $page
+                if (ClassInfo::hasMethod($page, 'CMSEditLink')) {
+                    $link = Controller::join_links($page->CMSEditLink(), 'ItemEditForm');
+                }
+            } else {
+                $link = $page->CMSEditLink();
+            }
+            // In-line editable blocks should just take you to the page.
+            // Editable ones should add the suffix for detail form.
+            if (!$this->inlineEditable() || $directLink) {
```
emteknetnz, line 751: Seems like this entire block be moved above the other one?

emteknetnz (changes requested):

src/Models/BaseElement.php
sabina-talipova: I cannot return it here. I use $link below

src/Models/BaseElement.php
sabina-talipova: I cannot return it here. I use $link below.

docs/en/examples/elementalarea_with_dataobject.md
```diff
@@ -0,0 +1,161 @@
+---
+title: Using ElementalArea with DataObject
+summary: Learn how to add ElementalArea to your models and manage it in a ModalAdmin
+---
+
```
GuySartorelli, line 5:
```suggestion
```
Title and summary are used in the docs in framework, because those get displayed on silverstripe.org. This isn't the case for this module (e.g. see https://raw.githubusercontent.com/silverstripe/silverstripe-elemental/4/docs/en/searching-blocks.md)

docs/en/examples/elementalarea_with_dataobject.md
```diff
@@ -0,0 +1,161 @@
...
+# Using ElementalArea with DataObject
+
+## Creating Model and adding ElementalArea
+
+In this example the class BlogPost will be created, it will extend the DataObject class. Also a new BlogPost Admin section will be created in the Admin panel where CMS users can manage the BlogPosts.
```
GuySartorelli, line 10:
```suggestion
In this example the class `BlogPost` will be created, it will extend the `DataObject` class. Also a new `BlogPost` Admin section will be created in the Admin panel where CMS users can manage the `BlogPost`s.

Note that you don't necessarily need to create an admin section to apply an `ElementalArea` to a `DataObject` - for example you could have a `DataObject` attached to a page.
```
All references to class names, methods, or properties should be quoted with backticks (\`).

docs/en/examples/elementalarea_with_dataobject.md
```diff
@@ -0,0 +1,161 @@
...
+<?php
+
+namespace App\Admins;
+
+use App\Models\BlogPost;
+use SilverStripe\Admin\ModelAdmin;
+
+class BlogPostsAdmin extends ModelAdmin
+{
+
```
GuySartorelli, line 26:
```suggestion
```

docs/en/examples/elementalarea_with_dataobject.md
```diff
@@ -0,0 +1,161 @@
...
+    private static array $managed_models = [
+        BlogPost::class,
+    ];
+}
+```
+
+`$managed_models` - this static variable connects BlogPostAdmin with BlogPost, which will be created in the following code.
+When you create a DataObject you should define the connection between your DataObject and ElementalArea by adding `$has_one` relationship.
+In this example BlogPost will have one ElementalArea.
```
GuySartorelli, line 41:
```suggestion
`$managed_models` - this static variable connects `BlogPostAdmin` with `BlogPost`, which will be created in the following code.
When you create a `DataObject` you should define the connection between your `DataObject` and `ElementalArea` by adding `$has_one` relationship.
In this example `BlogPost` will have one `ElementalArea`.
```

GuySartorelli: Note that the new lines won't actually appear when the file is rendered - if your intention is for each of these to display on a new line, you'll need to add two spaces at the end of the lines (https://www.markdownguide.org/basic-syntax#line-breaks).
I won't repeat this comment elsewhere but it applies to all such situations in the document.

docs/en/examples/elementalarea_with_dataobject.md
```diff
@@ -0,0 +1,161 @@
...
+
+    private static $owns = ['ElementalArea'];
+
+    // ...
+}
+ ```
+
+If you are using ElementalArea together with DataObject, it is important to define a few following methods in the class.
+First of all, the CMSEditLink() method should be defined to generate a link for opening the editing section of the particular BlogPost.
+BaseElement uses a link, that was created in CMSEditLink() method in BlogPost class, to generate new links for opening the editing section of the particular Element.
```
GuySartorelli, line 81:
```suggestion
If you are using `ElementalArea` together with `DataObject`, there are some important methods to implement on the class.
First of all, the `CMSEditLink()` method should be defined to generate a link for opening the editing section of the particular `BlogPost`.
`BaseElement` uses a link, that was created in `CMSEditLink()` method in the `BlogPost` class, to generate new links for opening the editing section of the particular element.

Note that for nested `GridField`s this method can get more complicated. Similarly, if your class is used in multiple admins, you will have to choose one to be the canonical admin section for the purposes of this method. This example only shows the simplest case, where the `BlogPost` class is used directly in `BlogPostsAdmin`.
```

docs/en/examples/elementalarea_with_dataobject.md
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    // ...
+}
+
+```
+
+Method Link() doesn't exist in DataObject class, so you should define this method in your class. You should add the canArchive() method in your class as well to set rules for the canDelete() method.
+You should also define the AbsoluteLink() method for opening the tabs in the Elements.
+Use the following code as an example of how to simply create these methods.
```
GuySartorelli, line 115:
```suggestion
The `Link()` method doesn't exist in the`DataObject` class, so you should define this method in your class. You should add the `canArchive()` method in your class as well to set rules for the `canDelete()` method.
You should also define the `AbsoluteLink()` method for opening the tabs in the elements.
Use the following code as an example of a simple implementation of these methods.
```

docs/en/examples/elementalarea_with_dataobject.md
```diff
@@ -0,0 +1,161 @@
...
+
+    public function AbsoluteLink($action = null)
+    {
+        return null;
+    }
+}
+
+```
+
+Add ElementalAreasExtension to the DataObject
```
GuySartorelli, line 149:
```suggestion
Add `ElementalAreasExtension` to the `DataObject`
```

docs/en/examples/elementalarea_with_dataobject.md
GuySartorelli: Why is `Link()` needed? If this is just for preview purposes, we can leave it out of this documentation and just refer to the new preview documentation.
Also what do you mean by "opening the tabs in the elements"?

docs/en/examples/elementalarea_with_dataobject.md
GuySartorelli: Why is `CMSEditLink()` needed? If this is just for preview purposes, we can leave it out of this documentation and just refer to the new preview documentation.
I think there's a report that uses it, right? If so then yeah definitely leave it in but can you please just confirm that for me?

docs/en/examples/elementalarea_with_dataobject.md
```diff
@@ -0,0 +1,161 @@
...
+
+        return $this->ID
+            ? Controller::join_links(
+                $admin->Link(str_replace('\\', '-', $this->ClassName)),
+                $this->ID
+                )
+            : null;
+    }
+
+    public function canArchive($member = null)
```
GuySartorelli, line 136: Why are we explicitly mentioning `canArchive` here? The class doesn't have to be `Versioned`.

docs/en/examples/elementalarea_with_dataobject.md
```diff
@@ -0,0 +1,161 @@
...
+use SilverStripe\Control\Controller;
+use SilverStripe\ORM\DataObject;
+use SilverStripe\Versioned\Versioned;
+
+class BlogPost extends DataObject
+{
+    private static string $table_name = 'BlogPost';
+
+    private static $extensions = [
+        Versioned::class,
```
GuySartorelli, line 61: Why are we explicitly saying this is `Versioned`? This should work whether the dataobject is `Versioned` or not, shouldn't it?

docs/en/examples/elementalarea_with_dataobject.md
```diff
@@ -0,0 +1,161 @@
...
+    ];
+
+    private static array $db = [
+        'Title' => 'Varchar(255)',
+        'Content' => 'HTMLText',
+    ];
```
GuySartorelli, line 67:
```suggestion
```
No need to include this in the example.

src/Models/BaseElement.php
```diff
@@ -579,7 +579,7 @@ public function getPage()
             return $this->cacheData['page'];
         }
 
-
+        // Despite the name of the method, method can return any type of DataObject
```
GuySartorelli, line 582: Which method are you talking about here? If you mean `getPage`, use the suggestion below - otherwise use the name of whatever method you're talking about.

Also, move the comment into the PHPDocs for the method instead, so developers can take advantage of their IDE displaying it to them. You should also update the typing mentioned in the PHPDoc since at the moment it explicitly says `null|SiteTree`.
```suggestion
        // Despite the name of the method, getPage can return any type of DataObject
```

src/Models/BaseElement.php
```diff
@@ -712,35 +712,65 @@ public function CMSEditLink($directLink = false)
...
-        if (!$page instanceof SiteTree && method_exists($page, 'CMSEditLink')) {
-            $link = Controller::join_links($page->CMSEditLink(), 'ItemEditForm');
-        } else {
+        // If $page is a Page
+        if ($page instanceof SiteTree) {
             $link = $page->CMSEditLink();
+        } else {
+            // Envoke method only if it exists in the $page
+            if (ClassInfo::hasMethod($page, 'CMSEditLink')) {
+                $link = Controller::join_links($page->CMSEditLink(), 'ItemEditForm');
```
GuySartorelli, line 745: Is it correct for `ItemEditForm` to be at the end of the url for inline-editable blocks?

tests/BaseElementTest.php
```diff
@@ -236,4 +247,55 @@ public function testOnBeforeWriteNoParent()
...
+    }
+
+    /**
+     * @param string $class
+     * @param string $element
+     * @param string|null $link
+     * @dataProvider getElementCMSLinkDataProvider
+     */
+
+    public function testGetElementCMSLink($class, $element, $link, $directLink = false)
```
GuySartorelli, line 290:
```suggestion
     * @dataProvider getElementCMSLinkDataProvider
     */
    public function testGetElementCMSLink(string $class, string $element, ?string $link, bool $directLink = false): void
```
Strongly type the method instead of adding phpdoc params.

tests/BaseElementTest.php
```diff
@@ -236,4 +247,55 @@ public function testOnBeforeWriteNoParent()
...
+    {
+        $obj = $this->objFromFixture($class, $element);
+        
+        $class = new ReflectionClass($class);
+        $method = $class->getMethod('getElementCMSLink');
```
GuySartorelli, line 295: Update this method to test `CMSEditLink()` instead, since ultimately that's the method that matters. We could remove this private method in the future, but `CMSEditLink()` must continue working as expected regardless.

GuySartorelli: Also since this is testing the logic that handles both `SiteTree` and non-`SiteTree` instances, make sure there's at least one `ElementalArea` with a `SiteTree` owner in this test - and confirm that you're testing both inline and non-inline editable for `SiteTree` and non-`SiteTree` alike.

GuySartorelli (changes requested):
I haven't actually tried running the example locally yet, but I figured I may as well submit this much of the review while I do that.

There are some questions here that you may have already gone through with Steve - if so, sorry about the duplication. Just want to make sure I fully understand what is happening here.

docs/en/examples/elementalarea_with_dataobject.md
```diff
@@ -0,0 +1,161 @@
...
+
+```yml
+App\Models\BlogPost:
+  extensions: 
+    - DNADesign\Elemental\Extensions\ElementalAreasExtension
+```
+
+## Related Documentation
+
+* [Preview](https://docs.silverstripe.org/en/4/developer_guides/customising_the_admin_interface/preview/)
```
GuySartorelli, line 161:
```suggestion
* [Preview](https://docs.silverstripe.org/en/4/developer_guides/customising_the_admin_interface/preview/) (note that this requires at least `silverstripe/admin` 4.11.0 and `dnadesign/silverstripe-elemental` 4.9.0.
```

docs/en/examples/elementalarea_with_dataobject.md
sabina-talipova: I follow Steve's comment https://github.com/silverstripe/silverstripe-elemental/pull/978#pullrequestreview-955619575

docs/en/examples/elementalarea_with_dataobject.md
sabina-talipova: Method canDelete() in BaseElement class invoke method canArchive on $page. Probably, it's expected that page is Versioned. And from Steve's comment https://github.com/silverstripe/silverstripe-elemental/pull/978#pullrequestreview-955619575, it's expected that possibly DataObject instance can be Versioned as well.

docs/en/examples/elementalarea_with_dataobject.md
GuySartorelli: I disagree with Steve's comment :p
The block being `Versioned` should be entirely independent of the page or `DataObject` being `Versioned`.
The `DataObject` is allowed to be unversioned. The block can still have its internal draft and published states, and that's fine. Just double check that if the object (in this case `BlogPost` is saved, that publishes any unpublished blocks (which I think it will by virtue of the `$owns` config).

docs/en/examples/elementalarea_with_dataobject.md
sabina-talipova: I'm not sure, that we can split word `BlogPost`s for formatting.

docs/en/examples/elementalarea_with_dataobject.md
GuySartorelli: It would be nicer if you add a check in `BaseElement` - something like
```php
if ($page->hasExtension(Versioned::class)) {
    // use $page->canArchive()
} else {
    // use $page->canDelete()
}
```

docs/en/examples/elementalarea_with_dataobject.md
sabina-talipova: No, user should define CMSEditLink() not for preview purpose.

GuySartorelli: Sweet, that's fine then 👍

docs/en/examples/elementalarea_with_dataobject.md
sabina-talipova: Since BaseElement implemented CMSPreviewable before (in this version CMSPreviewable interface was removed from  the BaseElement class), we have :
```php
                HiddenField::create('LiveLink', false, Director::absoluteURL($this->Link())),
                HiddenField::create('StageLink', false, Director::absoluteURL($this->PreviewLink())),
  ```
  in public function getCMSFields(). I mentioned this before and I also said that we should refactor BaseElement class and remove all unused methods.

Comment (GuySartorelli) [reply docs/en/examples/elementalarea_with_dataobject.md:115]: These methods are used in the next minor so that's fine.
If the issue is that `BaseElement::Link()` calls `$this->getPage()->Link()` then there should be a check in there. Something like this:

```php
if (ClassInfo::hasMethod($page, 'Link')) {
    $link = $page->Link($action);
    // do any transformations on the link e.g. concatenations
    return $link;
}

return null;
```

That way we're not imposing a link on the dataobject where it doesn't make sense to have one.

sabina-talipova: If we have Element that has additional tabs in Element, tab itself has link like "http://localhost.com/admin/blog-posts-admin/App-Models-BlogPost/EditForm/field/App-Models-BlogPost/item/4/ItemEditForm/field/ElementalArea/item/9/edit**#Root_Settings**"

GuySartorelli: Oh, you mean tabs in the CMS edit form! Okay. That won't come from the `Link()` method on the `Page`  though, surely. Otherwise it would looks like this for blocks pages:
`http://localhost.com/some-page-path/ItemEditForm/field/ElementalArea/item/9/edit#Root_Settings`
I think this is coming from the form itself, rather than relying on some `Link()` method to be defined on the `DataObject` subclass that owns the element. I could even see it _maybe_ using `CMSEditLink`, but `Link()` is for front-end purposes.

src/Models/BaseElement.php
sabina-talipova: We don't use 'ItemEditForm' in inline-editable block, we even can return NULL for this block.
But initially was 
```php
        if (!$page instanceof SiteTree && method_exists($page, 'CMSEditLink')) {
            $link = Controller::join_links($page->CMSEditLink(), 'ItemEditForm');
        } else {
            $link = $page->CMSEditLink();
        }
```
I don't want to remove it.

GuySartorelli: Ahh yup, that's fair. I missed that it was already there, so that's fine.

docs/en/examples/elementalarea_with_dataobject.md
GuySartorelli: I've tested this locally without `Link()` on the `BlogPost` class - so long as there's a check like the one I mention above it works fine even with tabs in the CMS.

docs/en/examples/elementalarea_with_dataobject.md
```diff
@@ -0,0 +1,111 @@
...
+
+    private static $owns = ['ElementalArea'];
+
+    // ...
+}
+ ```
+
+If you are using `ElementalArea` together with `DataObject`, it is important to define a following method in the class.\
+The `CMSEditLink()` method should be defined to generate a link for opening the editing section of the particular `BlogPost`.\
+`BaseElement` uses a link, that was created in `CMSEditLink()` method in `BlogPost` class, to generate new links for opening the editing section of the particular element.
```
GuySartorelli, line 66:
```suggestion
If you are using `ElementalArea` together with `DataObject`, it is important to define the `CMSEditLink()` method in the class.\
`BaseElement::CMSEditLink()` relies on a valid CMS link being available in the parent `DataObject` - in this case, `BlogPost`. It is used to navigate directly to the editing section of the particular element (for example in the "Content blocks in use" report).

Note that for nested `GridField`s this method can get more complicated. Similarly, if your class is used in multiple admins, you will have to choose one to be the canonical admin section for the purposes of this method. This example only shows the simplest case, where the `BlogPost` class is used directly in `BlogPostsAdmin`.
```
The language here was a bit confusing - I _think_ this is what you were trying to say, but if I got it wrong just let me know and we can figure out a clear way to communicate the correct message.

I've also added a note here about some edge cases developers may encounter which aren't covered in this example.

tests/BaseElementTest.php
```diff
@@ -236,4 +252,66 @@ public function testOnBeforeWriteNoParent()
...
+                'elementDataObject2',
+                null
+            ],
+        ];
+    }
+
+    /**
+     * @dataProvider getElementCMSLinkDataProvider
+     */
+
```
GuySartorelli, line 309:
```suggestion
```

docs/en/examples/elementalarea_with_dataobject.md
```diff
@@ -0,0 +1,111 @@
...
+
+```yml
+App\Models\BlogPost:
+  extensions: 
+    - DNADesign\Elemental\Extensions\ElementalAreasExtension
+```
+
+## Related Documentation
+
+* [Preview](https://docs.silverstripe.org/en/4/developer_guides/customising_the_admin_interface/preview/)\
```
GuySartorelli, line 110:
```suggestion
* [Preview](https://docs.silverstripe.org/en/4/developer_guides/customising_the_admin_interface/preview/)
```
We don't want a new line here - it looks a bit awkward.

src/Models/BaseElement.php
```diff
@@ -639,11 +643,12 @@ public function getAnchor()
     public function AbsoluteLink($action = null)
     {
         if ($page = $this->getPage()) {
-            $link = $page->AbsoluteLink($action) . '#' . $this->getAnchor();
-
-            $this->extend('updateAbsoluteLink', $link);
-
-            return $link;
+            if (ClassInfo::hasMethod($page, 'AbsoluteLink')) {
```
GuySartorelli, line 646: These two if statements should be put together:
```php
$page = $this->getPage();
if ($page && ClassInfo::hasMethod($page, 'AbsoluteLink')) {
```

src/Models/BaseElement.php
```diff
@@ -658,11 +663,12 @@ public function AbsoluteLink($action = null)
     public function Link($action = null)
     {
         if ($page = $this->getPage()) {
-            $link = $page->Link($action) . '#' . $this->getAnchor();
-
-            $this->extend('updateLink', $link);
+            if (ClassInfo::hasMethod($page, 'Link')) {
```
GuySartorelli, line 666: These two if statements should be put together:
```php
$page = $this->getPage();
if ($page && ClassInfo::hasMethod($page, 'Link')) {
```

src/Models/BaseElement.php
```diff
@@ -712,35 +718,65 @@ public function CMSEditLink($directLink = false)
...
-
-        if (!$page instanceof SiteTree && method_exists($page, 'CMSEditLink')) {
-            $link = Controller::join_links($page->CMSEditLink(), 'ItemEditForm');
-        } else {
+        // If $page is a Page
+        if ($page instanceof SiteTree) {
             $link = $page->CMSEditLink();
+        } else {
+            // Envoke method only if it exists in the $page
+            if (ClassInfo::hasMethod($page, 'CMSEditLink')) {
```
GuySartorelli, line 750:
```suggestion
        } else if (ClassInfo::hasMethod($page, 'CMSEditLink')) {
```

src/Models/BaseElement.php
```diff
@@ -712,35 +718,65 @@ public function CMSEditLink($directLink = false)
...
+                $link = Controller::join_links($page->CMSEditLink(), 'ItemEditForm');
+            }
```
GuySartorelli, line 752:
```suggestion
```

src/Models/BaseElement.php
```diff
@@ -712,35 +718,65 @@ public function CMSEditLink($directLink = false)
...
         if (!$page) {
-            $link = null;
-            $this->extend('updateCMSEditLink', $link);
             return $link;
         }
-
-        if (!$page instanceof SiteTree && method_exists($page, 'CMSEditLink')) {
-            $link = Controller::join_links($page->CMSEditLink(), 'ItemEditForm');
-        } else {
+        // If $page is a Page
```
GuySartorelli, line 745:
```suggestion
```
The code is self documenting here.

GuySartorelli (changes requested):
Nice work!
There are a few minor tweaks to make on the documentation, and some code quality tweaks to make, but I think it's just about ready.
I haven't done so in my review, but once you've made these changes can you please make sure to run `phpcs`? (You probably already do this but just in case)

At some stage we'll want behat coverage to ensure elemental blocks continue to work in arbitrary `DataObject`s in the future... I'll let you decide whether you want to do that with this PR or create an issue for it to be added later, given the time you've already spent on this one.

docs/en/examples/elementalarea_with_dataobject.md
sabina-talipova: I think we don't need this -> (for example in the "Content blocks in use" report). We don't have any code examples related to "Content blocks in use"

sabina-talipova:
Regarding "**_After thinking about this I'm taking this decision for you :p just creat an issue, it should be a separate piece of work. Adding behat tests for this will be a big piece of work._**"
I think, it would be better to implement behat tests together with this task, then leave it for long time. Probably, test cases for  Elements on a Page, I just need to add tests for DataObject. Am I right?

src/Models/BaseElement.php
sabina-talipova: Just in case, I use ELSEIF instead ELSE IF, because of this warning "Usage of ELSE IF is discouraged; use ELSEIF instead"

docs/en/examples/elementalarea_with_dataobject.md
GuySartorelli: Fair enough. My intention was to give context to the reason `CMSEditLink()` is needed on the `DataObject` subclass, but if you don't think that part's needed I'm okay leaving it out.

src/Models/BaseElement.php
GuySartorelli: Yup, fair enough. 👍

GuySartorelli:
> I just need to add tests for DataObject. Am I right?

You'd also need to add fixtures for it, which can't be done easily as `TestOnly` objects in behat, so it would mean adding new fixtures to https://github.com/silverstripe/silverstripe-frameworktest and adding that repository as a dev dependency for elemental. 
I won't stop you from adding that to this PR if that's what you'd prefer to do, but if you get frustrated with it I'd also be okay merging this without behat tests.

> I also want to add test cases for canDelete, Link and AbsoluteLink methods.

That sounds like a good idea

tests/BaseElementTest.php
```diff
@@ -236,4 +252,141 @@ public function testOnBeforeWriteNoParent()
...
     }
+
+    public function getElementCMSLinkDataProvider()
+    {
+        return [
+            // Element in DataObject with $directLink === true
+            'element1' => [
+                TestElement::class,
+                'elementDataObject1',
+                'http://localhost/test/1/ItemEditForm/field/ElementalArea/item/',
```
GuySartorelli, line 263: This link represents the `CMSEditLink`, correct? Shouldn't it start with `http://localhost/admin/`?

tests/BaseElementTest.php
```diff
@@ -236,4 +252,141 @@ public function testOnBeforeWriteNoParent()
...
+                'elementDataObject1',
+                'CMS_ACCESS_CMSMain',
+            ],
+        ];
+    }
+
+    /**
+     * @dataProvider canDeleteProvider
+     */
+    public function testCanDelete(
```
GuySartorelli, line 342: This checks that the result is not null, but it doesn't validate that the method works as expected - for example this won't confirm that a user with `canDelete()` permissions on the underlying `DataObject` has `canDelete()` permission on the element, nor (arguably more importantly) that users _without_ permission on the object don't have permission on the element.

tests/BaseElementTest.php
```diff
@@ -236,4 +252,141 @@ public function testOnBeforeWriteNoParent()
...
+    }
+
+    /**
+     * @dataProvider linksProvider
+     */
+    public function testLinkWithDataObject(string $class, string $element, ?string $link)
+    {
+        $object = $this->objFromFixture($class, $element);
+        $canDelete = $object->Link();
+        $this->assertEquals($link, $canDelete);
```
GuySartorelli, line 379:
```suggestion
        $this->assertEquals($link, $object->Link());
```
I assume the `$canDelete` variable name was a copy-paste error - but in any case there's no need to assign an intermediate variable here.

tests/Behat/features/add-block-element-to-data-object.feature
```diff
@@ -0,0 +1,33 @@
...
+    And I should see "Title"
+
+@runthis2
+  Scenario: I can add non-inline-editable elements to the DataObject
+    When I go to "/admin/test-blog-posts-admin/"
+    Then I press the "Add Test Blog Post" button
+    Then I press the "Create" button
+    Then I wait 1 second
+    Then I press the "Add block" button
+    Then I press the "Form" button in the add block popover
```
GuySartorelli, line 29: Where does the "form" button come from? If that's the userform elemental block, that's not included in this module's dev dependencies.

tests/Behat/features/add-block-element-to-data-object.feature
```diff
@@ -0,0 +1,33 @@
...
+    Then I should see "Untitled Form block" as the title for block 1
+    When I click on block 1
+    Then I should see new page with form
```
GuySartorelli, line 32: I'm not sure what the purpose of this is... are you testing that an elemental block that isn't inline-editable opens correctly? If so I recommend looking at the "I can edit a non in-line editable block" scenario in `edit-block-element.feature` and using the logic that already exists.

tests/Src/TestDataObject.php
```diff
@@ -0,0 +1,28 @@
...
+
+use DNADesign\Elemental\Models\ElementalArea;
+use SilverStripe\Control\Controller;
+use SilverStripe\Control\Director;
+use SilverStripe\ORM\DataObject;
+use SilverStripe\Dev\TestOnly;
+
+class TestDataObject extends DataObject implements TestOnly
+{
+
```
GuySartorelli, line 13:
```suggestion
```

tests/Src/TestDataObjectWithCMSEditLink.php
```diff
@@ -0,0 +1,37 @@
...
+
+use DNADesign\Elemental\Models\ElementalArea;
+use SilverStripe\Control\Controller;
+use SilverStripe\Control\Director;
+use SilverStripe\ORM\DataObject;
+use SilverStripe\Dev\TestOnly;
+
+class TestDataObjectWithCMSEditLink extends DataObject implements TestOnly
+{
+
```
GuySartorelli, line 13:
```suggestion
```

GuySartorelli (changes requested):
Tests are failing. Ignore the code coverage one for now - there's not really anything you can do about that. I'm hoping it will resolve itself by the time you push a fix for the other failing tests.

tests/BaseElementTest.php
sabina-talipova: I added "test" name for route specially to check that new URL from CMSEditLink method from DataObject. But if you think that "admin" is correct, I'll change.
DONE!

tests/BaseElementTest.php
sabina-talipova: Actually, the main purpose of this method check that it invoke canDelete or canArchive methods in class and don't return NULL or throws Exception. I don't have an aim to check permissions.  
DONE!

GuySartorelli:
There's a [note above](https://github.com/silverstripe/silverstripe-elemental/pull/978#discussion_r867577608) to address, plus some changes requested in silverstripe/silverstripe-frameworktest#105 - once those are addressed I'll test the behat tests locally

tests/Behat/features/add-block-element-to-data-object.feature
```diff
@@ -0,0 +1,47 @@
+@javascript
+Feature: Add elements in the CMS DataObject
+  As a CMS user
+  I want to add elements in the CMS DataObject
+  So that I can use multiple elements in a Dataobject
+  
+  Background:
+    Given I am logged in with "ADMIN" permissions
+
+@runthis
```
GuySartorelli, line 10:
```suggestion
```

tests/Behat/features/add-block-element-to-data-object.feature
```diff
@@ -0,0 +1,47 @@
...
+    Then I press the "Content" button in the add block popover
+    Then I should see "Untitled Content block" as the title for block 1
+    When I click on block 1
+    And I should see "Title"
+    And I should see "Content"
+    And I fill in "Elemental Behat Test Title" for "Title" for block 1
+    Then I press the "Save" button
+    Then I should see "Elemental Behat Test Title" as the title for block 1
+
+@runthis2
```
GuySartorelli, line 26:
```suggestion
```

GuySartorelli (changes requested):
Sorry, should have noticed this in the last review.
Looks good now, I'll check that the behat tests run with the fixtures in frameworktest when I have a moment. If that passes fine, I'll approve the fixtures and then get travis to run the tests again here before merging.

GuySartorelli:
Behat tests work locally. Restarting travis tests now that the fixtures are merged.

GuySartorelli (changes requested):
Tests are failing that [aren't failing on the `4.8` branch](https://app.travis-ci.com/github/silverstripe/silverstripe-elemental/builds/250482412)

GuySartorelli (approved):
Tested locally, works as expected.
The only remaining failures are also observed on the 4.8 branch, and are unrelated to this PR.

PR #936, opened 2021-09-28
FIX: Get CMS preview working for non inline-editable elements.

GuySartorelli:
Fixes #905. Depends on silverstripe/silverstripe-admin#1250

This was a feature in an earlier version of the module and most of the code is still in place.

Edit: While this technically depends on the PR from silverstripe-admin, it won't cause any defects if it does get merged in early, and merging it in even without that PR would make it easier for project-specific code to enable previews for elementals.

PR #980, opened 2022-04-27
MNT Add behat test for previewing elemental blocks

GuySartorelli:
Requires silverstripe/silverstripe-behat-extension#214

Tests non-inline-editable blocks can be previewed (functionality was fixed/added in #936)

### DIFF
--- a/tests/Behat/features/preview-a-block.feature
+++ b/tests/Behat/features/preview-a-block.feature
@@ -1,0 +1,43 @@
+Feature: Preview a non-inline-editable block
+  As an author
+  I want to preview the parent page of the non-inline-editable block I'm editing in the CMS
+  So that I can see how it would look like to my visitors
+
+  Background:
+    Given I add an extension "DNADesign\Elemental\Extensions\ElementalPageExtension" to the "Page" class
+      And I add an extension "DNADesign\Elemental\Tests\Src\TestReplacePageContentExtension" to the "Page" class
+      And content blocks are not in-line editable
+      And a "page" "Preview Blocks Page" with a "Alice's Block" content element with "Alice's content" content
+
+    Given I am logged in with "ADMIN" permissions
+      And I go to "/admin/pages"
+      And I left click on "Preview Blocks Page" in the tree
+      And I set the CMS mode to "Edit mode"
+      And I press the "Publish" button
+      And I click on block 1
+
+  @javascript
+  Scenario: I can show a preview of the current block's page from the pages section
+    When I set the CMS mode to "Preview mode"
+    Then I can see the preview panel
+      And the preview contains "Alice's content"
+
+    Given I set the CMS mode to "Edit mode"
+
+  @javascript
+  Scenario: I can see an updated preview when editing inline-editable blocks
+    When I fill in "<p>New published content</p>" for the "Content" HTML field
+      And I press the "Publish" button
+      And I fill in "<p>New preview content</p>" for the "Content" HTML field
+      And I press the "Save" button
+      And I set the CMS mode to "Preview mode"
+
+    When I switch the preview to "Published"
+    Then the preview does not contain "New preview content"
+      And the preview contains "New published content"
+
+    When I switch the preview to "Draft"
+    Then the preview does not contain "New published content"
+      And the preview contains "New preview content"
+
+    Given I set the CMS mode to "Edit mode"
--- a/tests/Src/TestReplacePageContentExtension.php
+++ b/tests/Src/TestReplacePageContentExtension.php
@@ -1,0 +1,14 @@
+<?php
+
+namespace DNADesign\Elemental\Tests\Src;
+
+use SilverStripe\Core\Extension;
+use SilverStripe\Dev\TestOnly;
+
+class TestReplacePageContentExtension extends Extension implements TestOnly
+{
+    public function getContent()
+    {
+        return $this->owner->ElementalArea;
+    }
+}


tests/Behat/Context/FixtureContext.php
```diff
@@ -74,4 +74,43 @@ public function contentBlocksAreNotInLineEditable()
...
+<h1>$Title</h1>
+<div class="content">
+    $Content
+    $ElementalArea
+</div>
+TEMPL;
+
+        $templFile = 'Layout/Page.ss';
+        $templPath = $this->getDestinationTemplatePath($templFile, true);
+        file_put_contents($templPath, $template);
```
emteknetnz, line 93: Where will this file go?  Into the current projects template path?  That doesn't seem right if that's the case.

tests/Behat/Context/FixtureContext.php
```diff
@@ -74,4 +74,43 @@ public function contentBlocksAreNotInLineEditable()
...
+
+    $themeConfig = <<<'CONF'
+---
+Name: mytheme-override
+Before: mytheme
+---
+SilverStripe\View\SSViewer:
```
emteknetnz, line 100: Is something like this done elsewhere in behat?  This seems wrong, can't we just update config like normal?

tests/Behat/features/preview-a-block.feature
```diff
@@ -0,0 +1,44 @@
...
+    Given I add an extension "DNADesign\Elemental\Extensions\ElementalPageExtension" to the "Page" class
+      And content blocks are not in-line editable
+      And the page template includes elemental blocks
+      And a "page" "Blocks Page" with a "Alice's Block" content element with "Alice's content" content
+
+    Given I am logged in with "ADMIN" permissions
+      And I go to "/admin/pages"
+      And I left click on "Blocks Page" in the tree
+      And I set the CMS mode to "Edit mode"
+    Then I should see an edit page form
```
emteknetnz, line 16: Shouldn't be doing assertions in 'Background', that's really just a test setup bit

tests/Behat/features/preview-a-block.feature
```diff
@@ -0,0 +1,44 @@
...
+
+    Given I click on block 1
+    Then I should see "Alice's Block"
+      And the "Content" field should contain "Alice's content"
+
+  @javascript
+  Scenario: I can show a preview of the current block's page from the pages section
+    When I set the CMS mode to "Preview mode"
+    Then I can see the preview panel
+      And the preview contains "Alice's content"
```
emteknetnz, line 28: Indentation is wonky

tests/Behat/Context/FixtureContext.php
GuySartorelli: By "something like this" do you mean adding config files? If so, yes - literally in the same file this is done for other config.

tests/Behat/Context/FixtureContext.php
GuySartorelli: It gets added to `app/templates/Layout/Page.ss`
I'm open to other suggestions - I don't want to override the simple theme template 'cause then I have to restore it correctly afterwards - I guess I could create a temporary theme instead... but I don't see what the difference is at the end of the day.

tests/Behat/features/preview-a-block.feature
GuySartorelli: Yeah I thought it was odd too - but this test is largely copying from other tests which also do this (see the edit block feature for example) - It's kinda good to know the setup is working as expected before we even get to the scenarios without having to repeat the same assertions again which is probably why it's done in the other ones.
That said I can change it if you feel strongly about it.

tests/Behat/features/preview-a-block.feature
GuySartorelli: Can you be more specific? Indentation in features are inconstent between modules and often between features in a given module, so it's not super clear what the indentation _should_ be - seems to be that "And" should be indented, as it's adding on top of a when/given/then step.

emteknetnz: Oh right, I never noticed how inconsistent it is

Probably doesn't matter then.  Personally I'd just left align everything, with a 2 space indent after each Background or Scenario, though I guess it doesn't really matter since there's no standard

tests/Behat/features/preview-a-block.feature
emteknetnz: Would prefer the assertions in the first scenario

tests/Behat/Context/FixtureContext.php
emteknetnz: Are there any other behat tests that add template files where we could copy their approach?

GuySartorelli: Not that I could find, that's why I need https://github.com/silverstripe/silverstripe-behat-extension/pull/214 as well.

tests/Behat/features/preview-a-block.feature
GuySartorelli: Fair. It does mean anyone running the second scenario locally won't have those assertions run but I can live with that. I'll make that change before pulling this out of draft.

GuySartorelli:
@emteknetnz I found a way around the template conundrum! I'm surprised a `TestOnly` extension works to be honest, since it's in the context of the actual runtime of the site - but it works! 🎉

GuySartorelli:
Some of the (unrelated to this PR) tests are failing, but they pass locally if I merge this into `4`. I've rebased which should resolve the issues travis was having.

emteknetnz (approved):
Merge on green

GuySartorelli:
Ahh! These tests pass if I run them independently, but if I run the suite as a whole they fail. All of the tests that run _after_ the new test are the ones which are failing.
I have discovered that this is caused by `I set the CMS mode to "Preview mode"` - It was still on preview mode when the remaining tests were meant to be performed because that is stored in the session. I have updated the tests to set back to edit mode.

@emteknetnz Since this is a change to the PR I probably need re-approval.

emteknetnz:
Travis is red

GuySartorelli:
Travis is now green! Composer got stuck (transport error) - restarting the job fixed it.

emteknetnz (approved):

PR #982, opened 2022-05-04
FIX Use PreviewLink of the owner object to preview blocks

GuySartorelli:
#871 allows `ElementalArea` to be added to any arbitrary `DataObject`. We should always try to set the correct preview link where possible. Some `DataObject`s may have a `PreviewLink` method for preview in the CMS, but no front-end link.

Finally, we should fallback to returning a null value if no link could be obtained for preview, so that the preview panel can correctly display a "no preview available" message.

## Parent Issues
- #871 

## Related
- silverstripe/silverstripe-admin#1304

### DIFF
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -701,7 +701,7 @@ JS
 
     /**
      * @param string|null $action
-     * @return string
+     * @return string|null
      * @throws \Psr\Container\NotFoundExceptionInterface
      * @throws \SilverStripe\ORM\ValidationException
      */
@@ -721,16 +721,28 @@ JS
 
     /**
      * @param string|null $action
-     * @return string
+     * @return string|null
      * @throws \Psr\Container\NotFoundExceptionInterface
      * @throws \SilverStripe\ORM\ValidationException
      */
     public function PreviewLink($action = null)
     {
-        $action = $action . '?ElementalPreview=' . mt_rand();
-        $link = $this->Link($action);
-        $this->extend('updatePreviewLink', $link);
+        $link = null;
+        if ($page = $this->getPage()) {
+            if (ClassInfo::hasMethod($page, 'Link')) {
+                $link = $page->Link($action);
+            }
+            if (!$link && ($page instanceof CMSPreviewable)) {
+                $link = $page->PreviewLink($action);
+            }
+            if ($link) {
+                // The ElementalPreview getvar is used in ElementalPageExtension
+                // The anchor must be at the end of the URL to function correctly
+                $link .= '?ElementalPreview=' . mt_rand() . '#' . $this->getAnchor();
+            }
+        }
 
+        $this->extend('updatePreviewLink', $link);
         return $link;
     }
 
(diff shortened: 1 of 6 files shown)


src/Models/BaseElement.php
```diff
@@ -714,17 +714,29 @@ public function Link($action = null)
...
     public function PreviewLink($action = null)
     {
-        $action = $action . '?ElementalPreview=' . mt_rand();
-        $link = $this->Link($action);
-        $this->extend('updatePreviewLink', $link);
+        if ($page = $this->getPage()) {
+            $link = null;
+            if (ClassInfo::hasMethod($page, 'Link')) {
+                $link = $page->Link($action);
+            } else if ($page instanceof CMSPreviewable) {
```
sabina-talipova, line 727: Could you, please, change `else if` to `elseif` to get rid of PSR warning.

src/Models/BaseElement.php
```diff
@@ -714,17 +714,29 @@ public function Link($action = null)
...
+                $link = $page->PreviewLink($action);
+            }
+            if ($link) {
+                // The ElementalPreview getvar is used in ElementalPageExtension
+                // The anchor must be at the end of the URL to function correctly
+                $link .= '?ElementalPreview=' . mt_rand() . '#' . $this->getAnchor();
+                $this->extend('updatePreviewLink', $link);
```
sabina-talipova, line 734: Probably, we should provide users with ability to extend PreviewLink even if link is NULL.

sabina-talipova:
Could you add testing for PreviewLink, please.

sabina-talipova:
Just minor changes. Nice to have this enhancement!

src/Models/BaseElement.php
GuySartorelli: I've changed this logic a bit instead, to make it a little more robust.

tests/BaseElementTest.php
```diff
@@ -401,4 +401,50 @@ public function testAbsoluteLink(string $class, string $element, ?string $link)
...
+     * @dataProvider previewLinksProvider
+     */
+    public function testPreviewLink(string $class, string $elementIdentifier, ?string $link)
+    {
+        /** @var BaseElement $element */
+        $element = $this->objFromFixture($class, $elementIdentifier);
+
+        if ($link) {
+            $regex = '/^' . preg_quote($link . '?ElementalPreview=', '/') .'\d*#' . $element->getAnchor() . '$/';
+            $this->assertTrue(preg_match($regex, $element->PreviewLink()));
```
sabina-talipova, line 445: Just small things preg_match(...) returns 1 (found), 0 (not found) or false (error).
assertTrue doesn't work with integers. 
Could you replace this with 
```php
            $result = preg_match($regex, $element->PreviewLink()) ? true : false;
            $this->assertTrue($result);
```
or cast to Boolean type explicitly 
```php $this->assertTrue((bool)preg_match($regex, $element->PreviewLink()));```

GuySartorelli: Can do.

GuySartorelli: Done. Good catch.

Issue #708, opened 2019-08-01
Removing an element’s PHP class breaks the CMS editor


kinglozzer:
```
TypeError: Cannot read property 'title' of undefined
    at t.value (bundle.js?m=1557373499:1)
    at Ln (vendor.js?m=1560140517:1)
    at Dn (vendor.js?m=1560140517:1)
    at Nn (vendor.js?m=1560140517:1)
    at Jn (vendor.js?m=1560140517:1)
    at Xn (vendor.js?m=1560140517:1)
    at _r (vendor.js?m=1560140517:1)
    at hr (vendor.js?m=1560140517:1)
    at dr (vendor.js?m=1560140517:1)
    at rr (vendor.js?m=1560140517:1)
```

graphql response looks like:

<img width="855" alt="Screenshot 2019-08-01 at 16 02 10" src="https://user-images.githubusercontent.com/1655548/62304376-c58c7280-b475-11e9-88c8-9f23e4787de1.png">

Manually deleting the records (e.g. id 117 from above) fixes it

## Acceptance criteria
- [x] The elemental area still renders
- [x] Blocks of deleted classes are not rendered on the front end
- [x] Blocks of deleted classes are displayed with a warning in the CMS. Users can not edit them, but can delete them.
- [x] We seek feedback from designers and bespoke teams on the best way to communicate the situation to the end user.

## PRs
- https://github.com/silverstripe/silverstripe-elemental/pull/991

robbieaverill:
What should we do here? The CMS says "this page type is obsolete" and I guess changes its class to something more generic on save (whatever is populated first in the class type dropdown). Do we need to do something similar for elements?

We need to handle this error gracefully anyway, maybe a "couldn't render this element: [error message here]" or something, assuming there is an error message.

cc @silverstripeux re: "How do we display elements in the editor that have had their PHP classes removed in code"

clarkepaul:
I like the idea of the error message but being able to extract the raw data so it could be copypasted into another block, and for the frontend we don't show anything.

My thought would be to show a collapsed block with the error message, expand it to show the table data (doesn't need to be pretty at all) with any additional error info if available.

kinglozzer:
Looking at what the graphql endpoint returns, I wonder if the original intention was for the block editor to display a generic “Block” as the block type... so maybe this a regression?

[...]

PR #991, opened 2022-07-13
FIX Obsolete classes don't break the elemental area.

GuySartorelli:
When the class for an elemental block is deleted, currently the elemental area simply fails to load.
This PR allows the elemental area to load, with a placeholder where the broken element is.

- Broken elements have no 'chevron' icon button (cannot expand or open an edit form)
- Broken elements always display a red "font-icon-block" icon as per the below example.
- Broken elements have "This element is of obsolete type {classname}" for the title
  - This is the same message displayed for pages when a page class is removed
- Broken elements' summaries include the title of the block (if any) and some basic info (you can archive, won't display on front end)
- You can archive the broken element - it has no other options
- You can create new blocks
- You can edit your other blocks
- You can save/publish/etc the other blocks
- You can save the page
- You can publish the page

Screenshot from 2022-07-13 14-31-01

## Parent issue
- https://github.com/silverstripe/silverstripe-elemental/issues/708


GuySartorelli:
UX advice:
- The example is clear enough but the wording could use some tweaking - particularly to indicate that this broken block won't display in the front end.

client/lang/src/en.json
```diff
@@ -19,6 +20,9 @@
     "ElementUnpublishAction.UNPUBLISH_PERMISSION_DENY": "Unpublish, insufficient permissions",
     "ElementAddElementPopover.SEARCH_BLOCKS": "Search blocks",
     "ElementAddNewButton.ADD_BLOCK": "Add block",
+    "ElementalElement.ARCHIVE_BROKEN": "Archive this block",
+    "ElementalElement.BROKEN_DESCRIPTION": "This block is broken and will not display on the front-end. You can archive it to remove it from this elemental area.",
+    "ElementalElement.BROKEN_DESCRIPTION_TITLE": "This block had the title \"{title}\". It is broken and will not display on the front-end. You can archive it to remove it from this elemental area.",
```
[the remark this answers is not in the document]
GuySartorelli: This message is for content authors. The expectation is that a developer will _know_ what changes they have made, and will therefore know how they can solve it if this message occurs during development. If we start adding messages for the developer _as well_ as for the content author then this message will get even longer than it already is.

sabina-talipova:
When I rename block, I get `Fatal error: Cannot declare class SilverStripe\ElementalBannerBlock\Block\BannerBlocks, because the name is already in use in /var/www/mysite/www/vendor/silverstripe/elemental-bannerblock/src/Block/BannerBlock.php on line 0`. Are you changes should solve this problem?

GuySartorelli:
> When I rename block, I get `Fatal error: Cannot declare class SilverStripe\ElementalBannerBlock\Block\BannerBlocks, because the name is already in use in /var/www/mysite/www/vendor/silverstripe/elemental-bannerblock/src/Block/BannerBlock.php on line 0.` Are you changes should solve this problem?

FYI for anyone seeing this in the future: We discussed this on slack. Showing friendly messages for errors like that one are out of scope for this PR.

GuySartorelli:
I've modified the formatting for summaries of broken blocks. Broken block summaries will now wrap, which allows us to keep all of the parts of the message as already written.

Non-broken block summaries still display as they already did. Changing that would be out of scope

sabina-talipova:
> > When I rename block, I get `Fatal error: Cannot declare class SilverStripe\ElementalBannerBlock\Block\BannerBlocks, because the name is already in use in /var/www/mysite/www/vendor/silverstripe/elemental-bannerblock/src/Block/BannerBlock.php on line 0.` Are you changes should solve this problem?
> 
> FYI for anyone seeing this in the future: We discussed this on slack. Showing friendly messages for errors like that one are out of scope for this PR.

I figured out what was the reason for this. When I renamed blocks to break them, some changes brake composer autoupload PSR-4 rules.

sabina-talipova:
Now solution works nice. Do you need to fix Travis failed build?

GuySartorelli:
No, we're ditching Travis this sprint anyway - the fact that it's failing on a behat test that passed in github actions CI tells me that the travis failure is almost certainly just a timing issue and rerunning it would let it pass, so it's not something to be actually concerned about now that we have github actions CI running.

tests/Behat/features/broken-element.feature
```diff
@@ -0,0 +1,37 @@
...
+      Then I should see "Alice's Block"
+      And I should see "Bob's Block"
+      And I should see "This element is of obsolete type BrokenClass"
+      And I should see "This one is broken"
+
+  @modal
+  Scenario: I can archive a block
+    Given I see a list of blocks
+    When I press the "View actions" button
+      Then I should see the archive button for block 2
```
sabina-talipova, line 27: Small question. Should we check that we don't see "Save" and "Publish" actions?

src/Models/BaseElement.php
```diff
@@ -293,6 +293,15 @@ public function canCreate($member = null, $context = array())
...
 
+    public function write($showDebug = false, $forceInsert = false, $forceWrite = false, $writeComponents = false)
+    {
+        // Skips writes for broken blocks, so that we can still publish the page to allow all other blocks to publish.
+        if ($this->ObsoleteClassName) {
+            return $this->ID;
+        }
+        return parent::write($showDebug, $forceInsert, $forceWrite, $writeComponents);
+    }
+
```
sabina-talipova, line 304: Could you test this method with broken and no broken elements?

sabina-talipova, line 304: Could you test this method with broken and no broken elements?

tests/Behat/features/broken-element.feature
GuySartorelli: Probably worth checking for that, yeah. Will add that.

src/Models/BaseElement.php
GuySartorelli: I'm not sure how (or if) I can get a broken element in a unit test context...

tests/Behat/features/broken-element.feature
GuySartorelli: Done. Also checking for duplicate button for completeness.
Adding this required some small refactoring of the behat `FeatureContext` - all the same behat test scenarios still work but now there is no duplication of code, and we can check for any action button in a uniform way instead of having a separate method for each possible action.

sabina-talipova (approved):

PR #994, opened 2022-07-29
MNT Use update-js action

emteknetnz:
Issue https://github.com/silverstripe/gha-update-js/issues/5


GuySartorelli (approved):
Failure is just scrutinizer which we're getting rid of this sprint anyway.

PR #999, opened 2022-08-20
DOC Correct title for userhelp

maxime-rainville:
Parent issue https://github.com/silverstripe/doc.silverstripe.org/issues/246


GuySartorelli (approved):
The failing scrutinizer build can be ignored - we don't use that service anymore.

PR #1000, opened 2022-08-22
MNT Fix github action for deploying userdocs

GuySartorelli:
Set the branches that activate the action to match the ones used by userhelp according to [the userhelp sources json](https://github.com/silverstripe/doc.silverstripe.org/blob/master/sources-user.js)

Also more clearly and accurately identify the specific action (main.yml is hardly specific).

## Parent Issue
- https://github.com/silverstripe/doc.silverstripe.org/issues/246

### DIFF
--- a/.github/workflows/deploy-userhelp-docs.yml
+++ b/.github/workflows/deploy-userhelp-docs.yml
@@ -1,13 +1,14 @@
-name: Build Docs
+name: Deploy Userhelp Docs
 on:
   push:
     branches:
-      - 'master'
+      - '5'
+      - '4'
     paths:
       - 'docs/en/userguide/**'
 jobs:
-  build:
-    name: build-docs
+  deploy:
+    name: deploy-userhelp-docs
     runs-on: ubuntu-latest
     steps:
       - name: Run build hook


GuySartorelli:
The scrutinizer failure can be ignored for this PR - we don't use it anymore, and it will be removed as a part of https://github.com/silverstripeltd/product-issues/issues/567

maxime-rainville (commented):
In theory, if any of the repos are updated the user help should be rebuilt. Would it make more sense to build this on a cron instead?

GuySartorelli:
> In theory, if any of the repos are updated the user help should be rebuilt. Would it make more sense to build this on a cron instead?

_maybe_? I don't think either approach is really more-or-less valuable than the other.
The current approach means it only gets rebuilt when there are actually changes to make - but it might result in rebuilding the site multiple times in quick succession for weird situations where we find multiple related issues.

A cron approach would mean we rebuild unnecessarily a lot (userdocs are not updated frequently at _all_) but would mean we don't have to worry about adding and updating these actions anywhere that userdocs sit.

I can see some value in both approaches but I don't think there is _enough_ value in the cron approach to warrant changing it right this very second. Maybe create a new issue for that if you think it's worthwhile?

maxime-rainville (approved):

PR #1005, opened 2022-08-29
ENH Replace ADMIN permissions with less permissions in Behat test 

sabina-talipova:
### Description
Replace user with "Admin" permissions to user with "Editor" permissions.

### Parent issue
- https://github.com/silverstripe/silverstripe-behat-extension/issues/220

### DIFF
--- a/tests/Behat/features/add-block-element-to-data-object.feature
+++ b/tests/Behat/features/add-block-element-to-data-object.feature
@@ -5,7 +5,8 @@ Feature: Add elements in the CMS DataObject
   So that I can use multiple elements in a Dataobject
 
   Background:
-    Given I am logged in with "ADMIN" permissions
+    Given the "group" "EDITOR" has permissions "CMS_ACCESS_CMSMain"
+    And I am logged in as a member of "EDITOR" group
 
   Scenario: I can add inline-editable elements to the DataObject
     When I go to "/admin/elemental-behat-test-admin/"
--- a/tests/Behat/features/element-editor.feature
+++ b/tests/Behat/features/element-editor.feature
@@ -6,12 +6,12 @@ Feature: View types of elements in an area on a page
 
   Background:
     Given I add an extension "DNADesign\Elemental\Extensions\ElementalPageExtension" to the "Page" class
-    And a "group" "AUTHOR group" has permissions "Access to 'Pages' section"
+    And a "group" "AUTHOR" has permissions "Access to 'Pages' section"
     And a "page" "Blocks Page" with a "Alice's Block" content element with "Some content" content
     And the "page" "Blocks Page" has a "Bob's Block" content element with "Some content II" content
 
   Scenario Outline: I can see the title and summary of each element
-    Given I am logged in with "<group>" permissions
+    Given I am logged in as a member of "AUTHOR" group
     When I go to "/admin/pages"
       And I left click on "Blocks Page" in the tree
     Then I should see a list of blocks
@@ -26,7 +26,7 @@ Feature: View types of elements in an area on a page
       | AUTHOR |
 
   Scenario: Opening the "more actions" menu will not expand a block
-    Given I am logged in with "ADMIN" permissions
+    Given I am logged in as a member of "AUTHOR" group
     When I go to "/admin/pages"
     And I left click on "Blocks Page" in the tree
       Then I should see a list of blocks
@@ -35,7 +35,7 @@ Feature: View types of elements in an area on a page
       Then I should not see "Title"
 
   Scenario: I can see the block type when I hover over an element's icon
-    Given I am logged in with "ADMIN" permissions
+    Given I am logged in as a member of "AUTHOR" group
     When I go to "/admin/pages"
       And I left click on "Blocks Page" in the tree
 
@@ -45,7 +45,7 @@ Feature: View types of elements in an area on a page
 
   @unsavedChanges
   Scenario: I can preview a block and hide the form again
-    Given I am logged in with "ADMIN" permissions
+    Given I am logged in as a member of "AUTHOR" group
     When I go to "/admin/pages"
       And I left click on "Blocks Page" in the tree
     When I see a list of blocks
@@ -69,7 +69,7 @@ Feature: View types of elements in an area on a page
   Scenario: I can operate blocks with a mouse
     Given a "virtual page" "Virtual Page"
 
-    Given I am logged in with "ADMIN" permissions
+    Given I am logged in as a member of "AUTHOR" group
     When I go to "/admin/pages"
 
     # Content blocks are not applied to other page types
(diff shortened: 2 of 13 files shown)


tests/Behat/features/element-editor.feature
```diff
@@ -11,7 +11,7 @@ Feature: View types of elements in an area on a page
     And the "page" "Blocks Page" has a "Bob's Block" content element with "Some content II" content
 
   Scenario Outline: I can see the title and summary of each element
-    Given I am logged in with "<group>" permissions
+    Given I am logged in with "AUTHOR" permissions
```
GuySartorelli: What is "AUTHOR" permissions?

tests/Behat/features/add-block-element-to-data-object.feature
```diff
@@ -5,7 +5,8 @@ Feature: Add elements in the CMS DataObject
   So that I can use multiple elements in a Dataobject
 
   Background:
-    Given I am logged in with "ADMIN" permissions
+    Given the "group" "EDITOR group" has permissions "Access to 'Pages' section" and "ELEMENTAL_EDIT"
+    And I am logged in with "EDITOR" permissions
```
GuySartorelli: What are "EDITOR" permissions?

GuySartorelli (commented):
The general approach looks good overall. Some bits confused me slightly so I've asked for clarification - I think I know what's going on but just want to validate my understanding.

tests/Behat/features/element-editor.feature
sabina-talipova: `And a "group" "AUTHOR group" has permissions "Access to 'Pages' section"`

How I see we create special group with restricted permissions, "only Page section". I can rename AUTHOR to EDITOR to more consistent in naming.

GuySartorelli: Ahh, okay. So it says "permissions" but in this case it actually means the _group_.

We probably want to change that to avoid confusion. For these where you are logging in with a _group_, can you please do the following:

1. In [silverstripe/silverstripe-behat-extension](https://github.com/silverstripe/silverstripe-behat-extension/)'s `LoginContext` class, add a new `public function iAmLoggedInWithGroup($group)` method:
    ```php
    /**
     * Creates a member in a specific group.
     * Example: Given I am logged in with the "Content Author" group
     *
     * @Given /^I am logged in with the "([^"]*)" group$/
     * @param string $permCode
     */
    public function iAmLoggedInWithGroup($group)
    {
        $email = "{$group}@example.org";
        $password = 'Secret!123';
        // Get or create group
        $group = Group::get()->filter('Title', $group)->first();
        if (!$group) {
            throw new LogicException("Group $group does not exist");
        }
        // Use the logic from generateMemberWithPermission to add a new member to the group
        // You may want to do some refactoring to avoid duplicated code
        $this->stepILogInWith($email, $password);
    }
    ```
1. Instead of naming the group "AUTHOR group" name it something descriptive like "elemental block editor"
1. Use the new `And I am logged in with the "elemental block editor" group` syntax

tests/Behat/features/add-block-element-to-data-object.feature
GuySartorelli: NVM I understand this now.

sabina-talipova: `Given the "group" "EDITOR group" has permissions "Access to 'Pages' section" and "ELEMENTAL_EDIT"`

How I see we create special group with restricted permissions "Access to 'Pages' section" and "ELEMENTAL_EDIT".
"ELEMENTAL_EDIT" permissions I defined in class ElementalBehatTestAdmin.

tests/Behat/features/element-editor.feature
sabina-talipova: Do you suggest it for all test cases where we use this `Given I am logged in with "*" permissions` approach?

GuySartorelli: Only if it's actually referring to a group.
E.g. `Given I am logged in with "ADMIN" permissions` is actually referring to a permission, so that's fine and should be left as is.

sabina-talipova: If deal only in sentence meaning, I can create new method only for this
``` php   
 /**
     * Creates and login as a member of a group with the correct permissions.
     * Example: Given I am logged in as a memeber of "ADMIN" group
     *
     * @Given /^I am logged in as memeber of "([^"]*)" group$/
     * @param string $permCode
     */
    public function iAmLoggedInAsMemeberOfGroup($permCode)
    {
        $this->iAmLoggedInWithPermissions($permCode);
    }
    
```

GuySartorelli: That solution would require the group to be prefixed with " group" and then you would have to reference it without that - e.g:
```
Given the "group" "elemental block editor group" has permissions "Access to 'Pages' section" and "ELEMENTAL_DATAOBJECT_EDIT"
And I am logged in with the "elemental block editor" group
```
That's not ideal. It's a bit confusing that you need to create the group with the word "group" at the end, but you don't actually use the word "group" a the end of the group name when logging in with it.

sabina-talipova: Actually, it's more confusing, since a user can log in as member, not a group or a role. 
Or a user can log in with the permission that belong to the certain group, not with the group. 

`And I am logged in with "AUTHOR" group permissions`
`And I am logged in as the member with "AUTHOR group" permissions`
`And I am logged in as the member of the "AUTHOR group"`

GuySartorelli: I'm not sure I follow what you mean... in either case I think the logic in the method snippet I pasted above is correct. I'll leave it up to you to decide the exact syntax for the sentence structure - so long as it's clear that you're intended to log into a user which is in a specific group which you have already set up (as opposed to logging in with specific permissions).

tests/Behat/features/add-block-element-to-data-object.feature
```diff
@@ -5,7 +5,8 @@ Feature: Add elements in the CMS DataObject
   So that I can use multiple elements in a Dataobject
 
   Background:
-    Given I am logged in with "ADMIN" permissions
+    Given the "group" "EDITOR group" has permissions "Access to 'Pages' section" and "ELEMENTAL_DATAOBJECT_EDIT"
```
GuySartorelli:
```suggestion
    Given the "group" "EDITOR" has permissions "Access to 'Pages' section" and "ELEMENTAL_DATAOBJECT_EDIT"
```
There should be no need to add "group" to the end of the group name. Obviously the rest of these will need to change as well.

GuySartorelli (changes requested):

tests/Behat/features/add-block-element-to-data-object.feature
```diff
@@ -5,7 +5,8 @@ Feature: Add elements in the CMS DataObject
   So that I can use multiple elements in a Dataobject
 
   Background:
-    Given I am logged in with "ADMIN" permissions
+    Given the "group" "EDITOR" has permissions "Access to 'Pages' section" and "ELEMENTAL_DATAOBJECT_EDIT"
+    And I am logged in as a memeber of "EDITOR" group
```
GuySartorelli:
```suggestion
    And I am logged in as a member of "EDITOR" group
```
Sorry, I should have noticed this typo last time.

GuySartorelli (changes requested):

GuySartorelli (approved):

Issue #1007, opened 2022-09-29
Update documentation for CMS 5 version


GuySartorelli:
The docs in the `5` branch haven't been updated and still reference CMS 4 stuff all over the place, including links to CMS 4 docs.
There are also some `yml` hints missing for code blocks - may as well add those at the same time.

## Acceptance criteria
- [x] Links to docs link to the CMS 5 docs
- [x] There are no references to CMS 4 except where needed to refer to the past
- [x] Code blocks which are for a specific programming language have that language in the code block hint.

## PRs
- https://github.com/silverstripe/silverstripe-elemental/pull/1008

maxime-rainville:
I moved this to peer review since all the ACs appear to have been met.

GuySartorelli:
Okay. I can't remember what state the PR is in though so whoever reviews it should make sure they actively search for outdated docs.

PR #1008, opened 2022-09-29
DOC Update documentation for CMS 5 version of elemental

GuySartorelli:
## Parent issue
- https://github.com/silverstripe/silverstripe-elemental/issues/1007


emteknetnz (changes requested):
There's a reference to `https://github.com/silverstripe/silverstripe-framework/blob/4/src/Forms/FormField.php` in README.md to update

GuySartorelli:
Good catch. Fixed that to point at API docs. Also fixed a similar link that was pointing at https://github.com/silverstripe/silverstripe-elemental-bannerblock/blob/3/src/Block/BannerBlock.php to point at the new `3` branch instead (no api docs for that repo)

emteknetnz (approved):

Issue #950, opened 2021-11-24
Invariant package JS warning on page publish


emteknetnz:
![image](https://user-images.githubusercontent.com/4809037/143320613-6e052607-cc20-41e3-911a-ce35c88d0144.png)

I don't think this is having an end user impact?

Link goes to here https://github.com/apollographql/invariant-packages

Likely a side-effect from https://github.com/silverstripe/silverstripe-elemental/pull/949, or one of the other upgrade JS pull-requests that happened recently

## Acceptance criteria
- [ ] The console warning is squash.
- [ ] If the connection drops, you don't end up with an infinite loop.

## PRs
- https://github.com/silverstripe/silverstripe-elemental/pull/1015

GuySartorelli:
Closing in favour of https://github.com/silverstripe/silverstripe-elemental/issues/1158

PR #1015, opened 2022-10-18
FIX Don't cause invariant violations

GuySartorelli:
## Parent issue
- fixes https://github.com/silverstripe/silverstripe-elemental/issues/950


client/src/legacy/ElementEditor/entwine.js
```diff
@@ -10,7 +10,18 @@ import { destroy } from 'redux-form';
...
-const resetStores = () => {
+const resetStores = (numAttempts = 0) => {
+  // If there are queries waiting to be performed, resetting the cache store will cancel them.
+  // Make sure all queries are complete before resetting stores.
+  // Only try 3 times - after that clearing the store is more likely to cause noticable issues
+  // than resolve them.
+  if (window.ss.apolloClient.queryManager.fetchQueryRejectFns.size > 0 && numAttempts < 3) {
+    setTimeout(() => {
+      resetStores(numAttempts + 1);
+    }, 200);
```
sabina-talipova, line 21:
```suggestion
    }, 500);
```

sabina-talipova: Could we increase to half a second. With 200ms I still see Error in browser console.

sabina-talipova (changes requested):

client/src/legacy/ElementEditor/entwine.js
GuySartorelli: Do you see the error _every time_ with this set to 200ms?

It's already trying 200ms three times - as the comment notes, if it takes longer than that (i.e. > 600ms total) then there's a higher chance of causing issues than resolving them by doing this, because by that time the user is likely interacting with the UI again.
Ultimately the console error doesn't seem to result in any _actual_ problems being caused, so if it still shows up in the console sometimes it's not the end of the world.

sabina-talipova: I have this error in console after 3rd attempt when I save or publish page and ElementalArea is refreshing. I tested it a few times with different time value and it fails on 200, 300 and 400ms, but not on 500ms.
If you say, that increasing time could be a cause of another issue, so we can add this code, but it doesn't solve this issue.

GuySartorelli: I don't think increasing the timestep is an option for the reasons I've mentioned.
If this is _always_ failing for you with the 200ms timestep then this PR is not a valid improvement and should be closed, and the ticket put back in the backlog.

sabina-talipova: I'll close it before we can find a better solution.

sabina-talipova:
CLOSE. Error is still in browser console.

PR #1018, opened 2022-11-02
API Stop using deprecated API

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10542

### DIFF
--- a/tests/ElementControllerTest.php
+++ b/tests/ElementControllerTest.php
@@ -13,8 +13,6 @@ class ElementControllerTest extends FunctionalTest
 {
     protected static $fixture_file = 'ElementControllerTest.yml';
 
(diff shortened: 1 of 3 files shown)


tests/ElementControllerTest.php
```diff
@@ -13,8 +13,6 @@ class ElementControllerTest extends FunctionalTest
 {
     protected static $fixture_file = 'ElementControllerTest.yml';
 
-    protected static $use_draft_site = true;
```
emteknetnz, line 16: Removing because FuntionalTest.use_draft_site is deprecated. This never actually needed to be set to true for the unit test to pass

sabina-talipova (approved):

Issue #1009, opened 2022-09-29
Apply TopPage extensions by default in CMS 5


GuySartorelli:
[The docs about the `TopPage` extensions](https://github.com/silverstripe/silverstripe-elemental/blob/4/docs/en/advanced_setup.md#top-page-reference-performance-enhancement) say these will be applied by default in CMS 5. We should do that.

## Acceptance Criteria
- [x] ~~The extensions that say they're applied by default _are_ applied by default~~
- [ ] This doesn't cause any obvious regressions
- [ ] This is mentioned in the changelog
- [ ] Top Page Extension is folded back in the base element

## PRs
- https://github.com/silverstripe/silverstripe-elemental/pull/1010
- https://github.com/silverstripe/silverstripe-elemental/pull/1022
- https://github.com/silverstripe/developer-docs/pull/94
- https://github.com/silverstripe/developer-docs/pull/113

emteknetnz:
@sabina-talipova You'll need to do an additional PR to the CMS 4 version to add `@deprecated` and `Deprecation::notice()` (probably wrapped in `Deprecation::withNoReplacement()) to any classes + methods being removed

sabina-talipova:
@emteknetnz could you check this failed CI, please.
https://github.com/silverstripe/silverstripe-cms/actions/runs/3467505777/jobs/5809980321

emteknetnz:
All linked PRs have been merged

PR #1019, opened 2022-11-03
API Integrate TopPage ext functionality in to the ElementArea etc.

sabina-talipova:
### Description
Integrate TopPage ext functionality in to the ElementArea and BaseElement

### Parent Issue
- https://github.com/silverstripe/silverstripe-elemental/pull/1009

### DIFF
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -5,7 +5,9 @@ namespace DNADesign\Elemental\Models;
 use DNADesign\Elemental\Controllers\ElementController;
 use DNADesign\Elemental\Forms\TextCheckboxGroupField;
 use DNADesign\Elemental\ORM\FieldType\DBObjectType;
+use DNADesign\Elemental\TopPage\TopPageTrait;
 use Exception;
+use Page;
 use SilverStripe\CMS\Controllers\CMSPageEditController;
 use SilverStripe\CMS\Model\SiteTree;
 use SilverStripe\Control\Controller;
@@ -50,6 +52,9 @@ use SilverStripe\ORM\CMSPreviewable;
  */
 class BaseElement extends DataObject implements CMSPreviewable
 {
+
+    use TopPageTrait;
+
     /**
      * Override this on your custom elements to specify a CSS icon class
      *
@@ -74,7 +79,8 @@ class BaseElement extends DataObject implements CMSPreviewable
     ];
 
     private static $has_one = [
-        'Parent' => ElementalArea::class
+        'Parent' => ElementalArea::class,
+        'TopPage' => Page::class,
     ];
 
     private static $extensions = [
@@ -93,6 +99,7 @@ class BaseElement extends DataObject implements CMSPreviewable
 
     private static $indexes = [
         'Sort' => true,
+        'TopPageID' => true,
     ];
 
     private static $versioned_gridfield_extensions = true;
@@ -197,6 +204,15 @@ class BaseElement extends DataObject implements CMSPreviewable
      * @var string
      */
     protected $anchor = null;
+
+    /**
+     * Global flag which indicates that automatic page determination is enabled or not
+     * If this is set to a page ID it will be used instead of trying to determine the top page
+     *
+     * @see TopPageTrait::withFixedTopPage()
+     * @var int
+     */
+    private $fixedTopPageID = 0;
 
     /**
      * Basic permissions, defaults to page perms where possible.
@@ -339,6 +355,7 @@ class BaseElement extends DataObject implements CMSPreviewable
             // Remove relationship fields
             $fields->removeByName('ParentID');
             $fields->removeByName('Sort');
+            $fields->removeByName('TopPageID');
 
             // Remove link and file tracking tabs
             $fields->removeByName(['LinkTracking', 'FileTracking']);
--- a/src/Models/ElementalArea.php
+++ b/src/Models/ElementalArea.php
@@ -3,6 +3,8 @@
 namespace DNADesign\Elemental\Models;
 
 use DNADesign\Elemental\Extensions\ElementalAreasExtension;
+use DNADesign\Elemental\TopPage\TopPageTrait;
+use Page;
 use SilverStripe\Core\ClassInfo;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Dev\TestOnly;
@@ -25,12 +27,18 @@ use SilverStripe\View\ViewableData;
  */
 class ElementalArea extends DataObject
 {
+    use TopPageTrait;
+
     private static $db = [
         'OwnerClassName' => 'Varchar(255)',
     ];
 
     private static $has_many = [
         'Elements' => BaseElement::class,
+    ];
+
+    private static $has_one = [
+        'TopPage' => Page::class,
     ];
 
     private static $extensions = [
@@ -53,6 +61,10 @@ class ElementalArea extends DataObject
         'Title' => 'Title',
     ];
 
+    private static $indexes = [
+        'TopPageID' => true,
+    ];
+
     private static $table_name = 'ElementalArea';
 
     /**
@@ -70,6 +82,15 @@ class ElementalArea extends DataObject
      * @var array
      */
     protected $cacheData = [];
+
+    /**
+     * Global flag which indicates that automatic page determination is enabled or not
+     * If this is set to a page ID it will be used instead of trying to determine the top page
+     *
+     * @see TopPageTrait::withFixedTopPage()
+     * @var int
+     */
+    private $fixedTopPageID = 0;
 
     /**
      * @return array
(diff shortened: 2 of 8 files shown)


src/Models/BaseElement.php
```diff
@@ -50,6 +52,9 @@
  */
 class BaseElement extends DataObject implements CMSPreviewable
 {
+
```
emteknetnz, line 55:
```suggestion
```

src/Models/BaseElement.php
```diff
@@ -5,7 +5,9 @@
 use DNADesign\Elemental\Controllers\ElementController;
 use DNADesign\Elemental\Forms\TextCheckboxGroupField;
 use DNADesign\Elemental\ORM\FieldType\DBObjectType;
+use DNADesign\Elemental\TopPage\TopPageTrait;
```
emteknetnz, line 8:
```suggestion
use DNADesign\Elemental\TopPage\TopPageTrait;
```
Can we please no longer use the `TopPage` dir, it's always bugged me that it's literally the only things that uses a 'feature' based dir structure.  Instead could you please delete the TopPage dir and instead move thing things inside dir to match the rest our conventions e.g. create a 'Traits' dir and move TopPageTrait there instead.  You'll need to update namespaces accordingly.

src/Models/BaseElement.php
```diff
@@ -74,7 +79,8 @@ class BaseElement extends DataObject implements CMSPreviewable
     ];
 
     private static $has_one = [
-        'Parent' => ElementalArea::class
+        'Parent' => ElementalArea::class,
+        'TopPage' => Page::class,
```
emteknetnz, line 83:
```suggestion
        'TopPage' => SiteTree::class,
```

src/Models/BaseElement.php
```diff
@@ -5,7 +5,9 @@
 use DNADesign\Elemental\Controllers\ElementController;
 use DNADesign\Elemental\Forms\TextCheckboxGroupField;
 use DNADesign\Elemental\ORM\FieldType\DBObjectType;
+use DNADesign\Elemental\TopPage\TopPageTrait;
 use Exception;
+use Page;
```
emteknetnz, line 10:
```suggestion
use SiteTree;
```
I don't like using the `Page` class here as it lives at the Project level and it can have literally anything on it.  We should be using the `SiteTree` class here instead

src/Models/BaseElement.php
```diff
@@ -198,6 +205,15 @@ class BaseElement extends DataObject implements CMSPreviewable
...
     protected $anchor = null;
 
+    /**
+     * Global flag which indicates that automatic page determination is enabled or not
+     * If this is set to a page ID it will be used instead of trying to determine the top page
+     *
+     * @see TopPageTrait::withFixedTopPage()
+     * @var int
+     */
+    private $fixedTopPageID = 0;
```
emteknetnz, line 215:
```suggestion
    private int $fixedTopPageID = 0;
```

src/Models/BaseElement.php
```diff
@@ -198,6 +205,15 @@ class BaseElement extends DataObject implements CMSPreviewable
      */
     protected $anchor = null;
 
+    /**
+     * Global flag which indicates that automatic page determination is enabled or not
+     * If this is set to a page ID it will be used instead of trying to determine the top page
+     *
+     * @see TopPageTrait::withFixedTopPage()
+     * @var int
```
emteknetnz, line 213:
```suggestion
```

src/Models/ElementalArea.php
```diff
@@ -3,6 +3,8 @@
 namespace DNADesign\Elemental\Models;
 
 use DNADesign\Elemental\Extensions\ElementalAreasExtension;
+use DNADesign\Elemental\TopPage\TopPageTrait;
+use Page;
```
emteknetnz, line 7:
```suggestion
use SiteTree;
```

src/Models/ElementalArea.php
```diff
@@ -33,6 +37,10 @@ class ElementalArea extends DataObject
         'Elements' => BaseElement::class,
     ];
 
+    private static $has_one = [
+        'TopPage' => Page::class,
```
emteknetnz, line 41:
```suggestion
        'TopPage' => SiteTree::class,
```

src/Models/ElementalArea.php
```diff
@@ -71,6 +83,15 @@ class ElementalArea extends DataObject
...
     protected $cacheData = [];
 
+    /**
+     * Global flag which indicates that automatic page determination is enabled or not
+     * If this is set to a page ID it will be used instead of trying to determine the top page
+     *
+     * @see TopPageTrait::withFixedTopPage()
+     * @var int
+     */
+    private $fixedTopPageID = 0;
```
emteknetnz, line 93:
```suggestion
     */
    private int $fixedTopPageID = 0;
```

src/TopPage/FluentExtension.php
```diff
@@ -34,7 +36,7 @@ class FluentExtension extends DataExtension
      */
     protected function assignTopPage(Page $page): void
     {
-        parent::assignTopPage($page);
+        TopPageTrait::assignTopPage($page);
```
emteknetnz, line 39: Calling a static method on a trait will cause a deprecation warning in php8.1
https://www.php.net/manual/en/migration81.deprecated.php#migration81.deprecated.core.static-trait

You'll need to update this to call something like `ElementalArea::assignTopPage($page)`

src/TopPage/FluentExtension.php
```diff
@@ -44,7 +46,7 @@ protected function assignTopPage(Page $page): void
      */
     protected function clearTopPage(): void
     {
-        parent::clearTopPage();
+        TopPageTrait::clearTopPage();
```
emteknetnz, line 49: Update to call non-trait

src/TopPage/FluentExtension.php
```diff
@@ -68,6 +70,6 @@ protected function saveChanges(array $extraData = []): void
         $owner = $this->owner;
         $extraData['"TopPageLocale"'] = $owner->TopPageLocale;
 
-        parent::saveChanges($extraData);
+        TopPageTrait::saveChanges($extraData);
```
emteknetnz, line 73: Update to call non-trait

src/TopPage/FluentExtension.php
```diff
@@ -54,7 +56,7 @@ protected function clearTopPage(): void
      */
     protected function assignFixedTopPage(): void
     {
-        parent::assignFixedTopPage();
+        TopPageTrait::assignFixedTopPage();
```
emteknetnz, line 59: Update to call non-trait

src/TopPage/TestState.php
```diff
@@ -9,7 +9,6 @@ class TestState implements BaseState
 {
     public function setUp(SapphireTest $test): void
```
emteknetnz, line 10: Delete this method

src/TopPage/TestState.php
```diff
@@ -18,23 +17,9 @@ public function tearDown(SapphireTest $test): void
 
     public function setUpOnce($class): void
```
emteknetnz, line 18: Delete this method

src/TopPage/TestState.php
```diff
@@ -18,23 +17,9 @@ public function tearDown(SapphireTest $test): void
...
     {
-        $this->disableTopPageUpdate();
     }
 
     public function tearDownOnce($class): void
```
emteknetnz, line 22: Delete this method

src/Models/ElementalArea.php
emteknetnz, line 41: The fact that we need to add the DB stuff on the implementing models now, makes me think we shouldn't be using a trait here and instead we should just be using extensions.  Extensions are basically the Silverstripe+ version of traits.

emteknetnz (changes requested):
I'm thinking we should be using extensions here not traits due to the fact extensions support $db and other model config out of the box.  

If we do decide to stick with traits (I'm really not sure why we would) then we also need to update the corresponding developer-docs PR

src/Models/BaseElement.php
michalkleiner: I think there was a wider discussion started somewhere whether `SiteTree` and `Page` should become one again (and be extended via extensions or projects' own `BasePage` sort of thing) because of modules requiring/using `Page` class which is defined in the project space.

Until that is decided I think either approach works fine as this is not the only instance where that would need changing. Or are we (semi-/officially) suggesting that everywhere a relation to a page is needed within the SS modules' area it should use SiteTree class to be universal? A lot of things would not function without the `Page` class existing anyway.

src/Models/ElementalArea.php
michalkleiner: I agree, I'd prefer to see this as a data extension.

src/TopPage/FluentExtension.php
michalkleiner: If we went with an extension this could be an instance method using `$this->owner->assignTopPage($page)`

sabina-talipova:
Close because made a decision to come back with Extension solution.

PR #1023, opened 2022-11-15
API Update deprecations

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10544

### DIFF
--- a/src/Extensions/ElementalAreaUsedOnTableExtension.php
+++ b/src/Extensions/ElementalAreaUsedOnTableExtension.php
@@ -36,6 +36,9 @@ class ElementalAreaUsedOnTableExtension extends DataExtension
      */
     public function updateUsage(ArrayList &$usage, DataObject &$record)
     {
+        Deprecation::withNoReplacement(function () {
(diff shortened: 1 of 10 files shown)


src/Extensions/ElementalAreaUsedOnTableExtension.php
```diff
@@ -36,6 +36,9 @@ public function updateUsageExcludedClasses(array &$excludedClasses)
      */
     public function updateUsage(ArrayList &$usage, DataObject &$record)
     {
+        Deprecation::withNoReplacement(function () {
```
GuySartorelli, line 39: This method call requires us to bump the framework dependency

GuySartorelli (changes requested):

src/Extensions/ElementalAreaUsedOnTableExtension.php
emteknetnz: Updated

GuySartorelli (approved):

PR #1029, opened 2022-12-08
FIX Cast possibly null link to string

emteknetnz:
Issue https://github.com/silverstripeltd/product-issues/issues/642

Fixes

`ERROR [Emergency]: Uncaught TypeError: SilverStripe\Control\Director::absoluteURL(): Argument #1 ($url) must be of type string, null given, called in /home/runner/work/silverstripe-graphql/silverstripe-graphql/vendor/dnadesign/silverstripe-elemental/src/Models/BaseElement.php on line 940`
https://github.com/silverstripe/silverstripe-graphql/actions/runs/3645001687/jobs/6154761063#step:10:1226

Exception happens during `dev/build` - [api change](https://github.com/silverstripe/silverstripe-framework/commit/53c0147f112a011f8cbdfb964a92dc4eb81a8b7f#diff-983f05f9a107b9bf66a1ed4230e524be82118dd71f4a795529f8794b762c8b47R408) on `Director::absoluteUrl()` means null is no longer accepted

### DIFF
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -374,9 +374,9 @@ class BaseElement extends DataObject implements CMSPreviewable
                 ->setTitle(_t(__CLASS__ . '.MainTabLabel', 'Content'));
 
             $fields->addFieldsToTab('Root.Main', [
-                HiddenField::create('AbsoluteLink', false, Director::absoluteURL($this->PreviewLink())),
-                HiddenField::create('LiveLink', false, Director::absoluteURL($this->Link())),
-                HiddenField::create('StageLink', false, Director::absoluteURL($this->PreviewLink())),
+                HiddenField::create('AbsoluteLink', false, Director::absoluteURL((string) $this->PreviewLink())),
+                HiddenField::create('LiveLink', false, Director::absoluteURL((string) $this->Link())),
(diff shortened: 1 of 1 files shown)


GuySartorelli:
Still failing ci?

GuySartorelli:
Was that API change documented? I can't find it in the changelog.

emteknetnz:
Yes changed - https://github.com/silverstripe/silverstripe-framework/pull/10594/files#diff-983f05f9a107b9bf66a1ed4230e524be82118dd71f4a795529f8794b762c8b47R408

Got changed because signature was strongly typed due to a deprecated param

Not documented - plan is to do all the API changes in one big go - https://github.com/silverstripeltd/product-issues/issues/627

GuySartorelli:
Okay. I wasn't aware of that (I thought that was only going to be for the deleted API) so it's probably worth mentioning that either in standup or in slack so the whole team is across that approach. But that's fine.

GuySartorelli (approved):
See https://github.com/silverstripe/silverstripe-cms/pull/2806#discussion_r1046613902 for a discussion on this approach. Strongly typing the return type of the given methods is out of scope for CMS 5.

PR #1032, opened 2023-01-13
DEP Upgrade front-end build stack

GuySartorelli:
See https://github.com/silverstripe/silverstripe-admin/pull/1389 for a wider discussion on most of the changes

## Parent issue
- https://github.com/silverstripeltd/product-issues/issues/644


client/src/components/ElementActions/ArchiveAction.js
```diff
@@ -30,7 +30,9 @@ const ArchiveAction = (MenuComponent) => (props) => {
     if (handleArchiveBlock && window.confirm(archiveMessage)) {
       handleArchiveBlock(id).then(() => {
         const preview = window.jQuery('.cms-preview');
-        preview.entwine('ss.preview')._loadUrl(preview.find('iframe').attr('src'));
+        if (preview && typeof preview.entwine === 'function') {
+          preview.entwine('ss.preview')._loadUrl(preview.find('iframe').attr('src'));
+        }
```
GuySartorelli, line 35: Required for testing purposes, where we don't actually have jquery with entwine in the DOM.

client/src/components/ElementActions/SaveAction.js
```diff
@@ -1,4 +1,11 @@
-/* global window */
+/* global jest, describe, it, expect, window */
+
+jest.mock('isomorphic-fetch', () =>
+  () => Promise.resolve({
+    json: () => ({}),
+  })
+);
```
GuySartorelli, line 7: Taken from https://github.com/silverstripe/silverstripe-admin/blob/856955c2a83be9d80d2257f1300082207827242c/client/src/components/TreeDropdownField/tests/TreeDropdownField-test.js#L3-L7

babel.config.json
```diff
@@ -0,0 +1,6 @@
+{
```
GuySartorelli, line 1: See https://github.com/silverstripe/silverstripe-asset-admin/pull/1320#discussion_r1068695104 for explanation of this file

client/src/components/ElementActions/tests/PublishAction-test.js
```diff
@@ -1,6 +1,12 @@
 /* eslint-disable import/no-extraneous-dependencies */
 /* global jest, describe, it, expect, window */
 
+jest.mock('isomorphic-fetch', () =>
+  () => Promise.resolve({
+    json: () => ({}),
+  })
+);
+
```
GuySartorelli, line 9: Taken from [TreeDropdownField-test.js](https://github.com/silverstripe/silverstripe-admin/blob/856955c2a83be9d80d2257f1300082207827242c/client/src/components/TreeDropdownField/tests/TreeDropdownField-test.js#L3-L7) to avoid `ReferenceError: TextEncoder is not defined` when running jest tests. See [this stackoverflow question](https://stackoverflow.com/questions/68468203/why-am-i-getting-textencoder-is-not-defined-in-jest) for more details about this error and its causes.

GuySartorelli:
Failing unit test is unrelated.

tests/Behat/features/edit-block-element.feature
```diff
@@ -68,7 +68,7 @@ Feature: Edit elements in the CMS
       And I fill in "Charlie's Block" for "Title" for block 2
     When I press the "Save" button in the actions menu for block 2
       And I wait 1 second
-      Then I should see a "Saved 'Charlie's Block' successfully" notice
+      Then I should see a "Saved 'Charlie's Block' successfully" success toast
```
GuySartorelli, line 71: The "I should see some notice" fixture was removed, as it had previously been deprecated. See [silverstripe/silverstripe-framework#10594](https://github.com/silverstripe/silverstripe-framework/pull/10594/files#diff-b5ac2578edc2e46f01f689ceab76303bb7c8897d79e80962a1d6d65e18db7595L79-L87)

tests/Behat/Context/FixtureContext.php
```diff
@@ -4,14 +4,11 @@
 use DNADesign\Elemental\Models\BaseElement;
 use DNADesign\Elemental\Models\ElementalArea;
 use DNADesign\Elemental\Models\ElementContent;
-use SilverStripe\BehatExtension\Context\FixtureContext as BaseFixtureContext;
+use SilverStripe\CMS\Tests\Behaviour\FixtureContext as BaseFixtureContext;
```
GuySartorelli, line 7: Necessary to include the anchor dropdown fixture method - we can't just add the `SilverStripe\CMS\Tests\Behaviour\FixtureContext` class to the fixtures in `behat.yml` because it has the same superclass as this one, which causes behat to fail since it doesn't know whether to call methods from this fixture class or the CMS one.
Using the CMS fixture class as the superclass here resolves that problem.

tests/Behat/features/element-editor.feature
```diff
@@ -20,11 +20,6 @@ Feature: View types of elements in an area on a page
       And I should see "Bob's Block" as the title for block 2
       And I should see "Some content II" as the summary for block 2
 
-    Examples:
-      | group  |
-      | ADMIN  |
-      | AUTHOR |
```
GuySartorelli, line 26: The scenario outline wasn't actually using these variables. See [the docs on scenario outlines](https://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines)
There's no reason to use admin here anyway so it makes more sense to just remove the examples and change the outline to a normal scenario.

tests/Behat/features/element-editor.feature
```diff
@@ -121,13 +116,25 @@ Feature: View types of elements in an area on a page
     And I wait for 1 seconds
     Then I should see a ".tox-tinymce" element
 
-    # More actions menu can be accessed with keyboard actions
+    # More actions menu can be accessed and hidden with keyboard actions
+    # Doing so does not collapse the block as a whole
     When I press the "tab" key globally
     Then I should not see "Duplicate"
+    And I should see a ".tox-tinymce" element
```
GuySartorelli, line 123: There was an issue where the block expansion was being affected by this menu button

tests/Behat/features/element-editor.feature
```diff
@@ -121,13 +116,25 @@ Feature: View types of elements in an area on a page
...
+    # open menu with space and close with escape
     When I press the "space" key globally
     Then I should see "Duplicate"
+    And I should see a ".tox-tinymce" element
+    When I press the "escape" key globally
```
GuySartorelli, line 128: An update to bootstrap and/or reactstrap has changed the behaviour to give focus to the first element in the dropdown when opening with the keyboard, so we need to use escape to close it rather than hitting space/enter again.

GuySartorelli:
Behat passes locally - I'll set up a full kitchen sink run (link in the main issue) with all of the PRs running all of the behat to show whether this actually passes with all of the changes applied.

emteknetnz (approved):

PR #1035, opened 2023-01-19
MNT Broken builds

emteknetnz:
Issue https://github.com/silverstripeltd/product-issues/issues/659

Fixes:
- `Error: Call to undefined method GraphQL\Type\Definition\FieldDefinition::create()`
- `TypeError: DNADesign\Elemental\TopPage\SiteTreeExtension::onBeforeDuplicate(): Argument https://github.com/silverstripeltd/product-issues/issues/1 ($original) must be of type Page, SilverStripe\CMS\Model\SiteTree given`

The Page/SiteTree type error only occurs in recipe-kitchen-sink, not installer. I'm not sure why this is, though I don't there's any downsite to loosening the Page type the more standard SiteTree class which is allowed since this is for a new major version

The endtoend tests are failing because the elemental editor in the page edit form is not rendering at all - there are the following javascript errors which I'm going to assume are going to be fixed in [this issue](https://github.com/silverstripeltd/product-issues/issues/644)
image

### DIFF
--- a/src/TopPage/DataExtension.php
+++ b/src/TopPage/DataExtension.php
@@ -4,7 +4,7 @@ namespace DNADesign\Elemental\TopPage;
 
 use DNADesign\Elemental\Models\BaseElement;
 use DNADesign\Elemental\Models\ElementalArea;
-use Page;
+use SilverStripe\CMS\Model\SiteTree;
 use SilverStripe\Core\ClassInfo;
 use SilverStripe\ORM\DataExtension as BaseDataExtension;
 use SilverStripe\ORM\DataObject;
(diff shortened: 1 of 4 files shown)


src/TopPage/DataExtension.php
```diff
@@ -31,7 +31,7 @@ class DataExtension extends BaseDataExtension
      * @var array
      */
     private static $has_one = [
-        'TopPage' => Page::class,
+        'TopPage' => SiteTree::class,
```
maxime-rainville, line 34: I'm a bit concerned what will happen when you migrate a page from CMS4 to CMS5 and this relation changes.

I'll create some dummy content block on a CMS4 site and manually hack this relation to reference SiteTree to validate how the Data is migrated.

maxime-rainville: Tested and the value retains its property. Even if it didn't it probably wouldn't be the end of the world since this is meant to act as a sort of cache any way.

maxime-rainville (approved):
LGTM

Issue #1038, opened 2023-02-08
CMS 4.12: Preview broken for nested element (ElementList) in elemental 4.10


davejtoews:
Previews for [Elemental List module](https://github.com/dnadesign/silverstripe-elemental-list) appear to be broken using Elemental 4.10.

When attempting to browse to a child element with the editor in split mode, the CMS redirects back to the page editor. Everything works fine in edit mode so it seems like the preview window is causing this somehow. If I successfully browse to the child element editor in Edit mode, and then switch to split or preview mode, the redirect happens then too. 

https://user-images.githubusercontent.com/9664035/217678835-00aed074-5fb8-4aad-85b7-ce69fcfbfe84.mov

The preview URL for a top-level element looks like:
/new-content-page/?ElementalPreview=1988683919&stage=Stage&CMSPreview=1#e3

Preview url for a nested element looks like:
/new-content-page/?stage=Stage&CMSPreview=1#e3?ElementalPreview=1821505376#e4

But it get's redirected back to the page preview:
/new-content-page/?stage=Stage&CMSPreview=1

The nested Element has two sets of ?query and #anchor parameters in the URL, which seems broken. The problem was likely introduced in this commit: https://github.com/silverstripe/silverstripe-elemental/commit/24f6cde44a223f124b8188fe81189cb3331bdbe5

Thanks @GuySartorelli for helping me identify this.

## PRs
- https://github.com/silverstripe/silverstripe-elemental/pull/1039

GuySartorelli:
Fixed in [4.10.1](https://github.com/silverstripe/silverstripe-elemental/releases/tag/4.10.1)

PR #1039, opened 2023-02-09
FIX Respect existing query strings and anchors in preview link

GuySartorelli:
If the object owning the block has its own query string and/or anchor in the preview URL, it causes problems (see issue)

## parent issue
- https://github.com/silverstripe/silverstripe-elemental/issues/1038

### DIFF
--- a/tests/BaseElementTest.php
+++ b/tests/BaseElementTest.php
@@ -14,6 +14,7 @@ use DNADesign\Elemental\Tests\Src\TestElementDataObject;
 use DNADesign\Elemental\Tests\Src\TestDataObjectWithCMSEditLink;
 use DNADesign\Elemental\Tests\Src\TestMultipleHtmlFieldsElement;
 use DNADesign\Elemental\Tests\Src\TestPage;
+use DNADesign\Elemental\Tests\Src\TestPreviewableDataObjectWithLink;
 use Page;
 use ReflectionClass;
 use SilverStripe\Control\Director;
@@ -48,6 +49,7 @@ class BaseElementTest extends FunctionalTest
         TestDataObjectWithCMSEditLink::class,
         TestElementDataObject::class,
         TestMultipleHtmlFieldsElement::class,
+        TestPreviewableDataObjectWithLink::class,
     ];
 
     public function testSimpleClassName()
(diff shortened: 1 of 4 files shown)


tests/BaseElementTest.php
```diff
@@ -463,6 +465,12 @@ public function previewLinksProvider()
                 'elementDataObject4',
                 'base-link',
             ],
+            // Element in DataObject WITH PreviewLink AND Link which has its own query string and anchor link
+            [
+                TestElement::class,
+                'elementDataObject5',
+                'base-link?something=value&somethingelse=value2',
```
GuySartorelli, line 472: It'll keep the existing query string but replace the anchor

tests/BaseElementTest.php
```diff
@@ -473,12 +481,16 @@ public function testPreviewLink(string $class, string $elementIdentifier, ?strin
...
+        $previewLink = $element->PreviewLink();
 
         if ($link) {
-            $regex = '/^' . preg_quote($link . '?ElementalPreview=', '/') .'\d*#' . $element->getAnchor() . '$/';
-            $this->assertTrue((bool)preg_match($regex, $element->PreviewLink()));
+            $regex = '/^' . preg_quote($link, '/') . '[?&]' . preg_quote('ElementalPreview=', '/') .'\d*#' . $element->getAnchor() . '$/';
+            $this->assertMatchesRegularExpression($regex, $previewLink);
+            // Doesn't try to blindly append query string and anchor - but instead merges intelligently with whatever's already there
+            $this->assertSame(1, substr_count($previewLink, '?'));
+            $this->assertSame(1, substr_count($previewLink, '#'));
```
GuySartorelli: Replaced `assertTrue` with the more correct `assertMatchesRegularExpression` here and modified the regex to account for the new test scenario.

Also counting to make sure we only ever have one query string and one anchor link

davejtoews:
This seems to have resolved the issues we were having on #1038. Thank you!

emteknetnz (approved):
It fine and I've tested it works, though I was also unable to replicate the original issue without this patch.

@GuySartorelli can you please confirm if we still want this merged?

GuySartorelli:
Yes we want it - but you shouldn't merge if you can't reproduce the issue.
I'm assuming you were looking at the preview of a non-inline-editable block which is itself inside a [list block](https://github.com/dnadesign/silverstripe-elemental-list)? You need `$MetaTags` in your base model (normally page) template to see the redirection, but you can just check the preview iframe src as well.

emteknetnz:
Ah right, yup can replicate by adding a list block to a list block

With this patch, I can now add list blocks to list blocks the split view works properly

Issue #1041, opened 2023-03-01
TopPageExtension adds an extra field to content block


maxime-rainville:
When the TopPageExtension is applied to content blocks, it inadvertently adds an editable field to every content block.

This field is not meant to be directly editable/viewable by content authors. 

This affects CMS4 and CMS5, but is especially bad in CMS5 because the extension is now on by default.

![image](https://user-images.githubusercontent.com/1168676/222027667-20259332-ebff-4cf4-b7ef-0d84d60b70e8.png)

## Acceptance criteria
- "Top page" field is not displayed on the front end in both CMS4 and CMS5

## Related issue
- https://github.com/silverstripe/silverstripe-elemental/issues/1009

## PRs
- https://github.com/silverstripe/silverstripe-elemental/pull/1042

GuySartorelli:
PR merged. The "and CMS 5" part will be resolved when it gets merged up.

PR #1042, opened 2023-03-02
FIX Remove TopPageID from editable fields

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-elemental/issues/1041

### DIFF
--- a/src/TopPage/DataExtension.php
+++ b/src/TopPage/DataExtension.php
@@ -12,6 +12,7 @@ use SilverStripe\ORM\Queries\SQLUpdate;
 use SilverStripe\ORM\ValidationException;
 use SilverStripe\Versioned\Versioned;
 use SilverStripe\View\ViewableData;
+use SilverStripe\Forms\FieldList;
 
 /**
  * Class DataExtension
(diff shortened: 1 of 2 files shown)


GuySartorelli (changes requested):
The same treatment is needed in `FluentExtension` to hide the `TopPageLocale` field.

src/TopPage/DataExtension.php
```diff
@@ -255,6 +255,11 @@ public function withFixedTopPage(int $topPageID, callable $callback)
         }
     }
 
+    public function updateCMSFields($fields)
```
GuySartorelli:
```suggestion
    public function updateCMSFields(FieldList $fields)
```
Needs to match the superclass's method signature
https://github.com/silverstripe/silverstripe-framework/blob/66561ccb49e196bd652b1fc2112c5aeda062354c/src/ORM/DataExtension.php#LL218C36-L218C55

GuySartorelli (changes requested):

src/TopPage/DataExtension.php
emteknetnz: Updated

emteknetnz:
Have updated FluentExtension - good spot

GuySartorelli (approved):

PR #1044, opened 2023-03-06
Prevent mapStateToProps re-rendering on formstate changes

blueo:
any formstate change, eg from another FormLoader component will trigger the ElementalEditor to re-render because this function creates a new object for every run. This change assumes form state does not change after it is created. If it needs to, it could be that we need to use another way to pass down that difference.

An example of the re render can be seen here: https://github.com/silverstripe/silverstripe-linkfield/pull/56 where including a field with its own FormBuilderLoader triggers an infinite loop of state updates.


blueo:
I should clarify that if the formstate object is replaced with a new object - then it will trigger a re-render. This would be the desirable/recommended way to make updates IMO.

GuySartorelli:
The prefer-lowest job failure is unrelated to this PR and can be disregarded.

GuySartorelli:
@blueo Targetting the `5` branch will mean this isn't released until `5.1.0` - if you want this to be in the `5.0.0` stable release, please target `5.0`

GuySartorelli:
> This change assumes form state does not change after it is created

I don't know if this is a safe assumption.... the whole point of state is that it _can and does_ change. I'm guessing things like "is the form dirty" (i.e. has changes, which will change any time you... well... make a change) are in the form state.

blueo:
> > This change assumes form state does not change after it is created
> 
> I don't know if this is a safe assumption.... the whole point of state is that it _can and does_ change. I'm guessing things like "is the form dirty" (i.e. has changes, which will change any time you... well... make a change) are in the form state.

yeah just realising this after a bit more testing. I think we'll need to find a way to update the formstate object on a change. 

I guess the schema never changes - but the values etc do...

blueo:
Ok so I misunderstood what was happening there - it creates a new key in `formState` for every element form.
I've added some memoization to that seems to prevent enough re-rendering to stop the infinite loop that occurs when the linkfield is added. Lets hope tests pass this time 🤞

GuySartorelli:
Looks good. Have you checked this new implementation resolves the problem that prompted all this?

Also as per https://github.com/silverstripe/silverstripe-elemental/pull/1044#issuecomment-1457269348 can you please retarget to `5.0` if you want it in the `5.0.0` stable release?

blueo:
> Looks good. Have you checked this new implementation resolves the problem that prompted all this?
> 
> Also as per [#1044 (comment)](https://github.com/silverstripe/silverstripe-elemental/pull/1044#issuecomment-1457269348) can you please retarget to `5.0` if you want it in the `5.0.0` stable release?

haha yes I have tested it with the linkfield module and it is working. I've retargeted it too.

blueo:
rebased to take out some of the 5 branch commits

GuySartorelli (approved):
Looks good, works locally.

PR #1052, opened 2023-04-18
MNT Use React Testing Library

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1419

### DIFF
--- a/client/src/components/ElementActions/tests/AbstractAction-test.js
+++ b/client/src/components/ElementActions/tests/AbstractAction-test.js
@@ -1,62 +1,60 @@
 /* eslint-disable import/no-extraneous-dependencies */
-/* global jest, describe, it, expect */
+/* global jest, test, describe, it, expect */
 
 import React from 'react';
 import AbstractAction from '../AbstractAction';
-import Enzyme, { mount } from 'enzyme';
-import Adapter from 'enzyme-adapter-react-16/build/index';
+import { render, fireEvent } from '@testing-library/react';
 
-Enzyme.configure({ adapter: new Adapter() });
+function makeProps(obj = {}) {
+  return {
+    onClick: () => {},
+    title: 'My abstract action',
+    disabled: false,
+    className: 'foo-bar',
+    toggle: false,
+    ...obj,
+  };
+}
 
-describe('AbstractAction', () => {
-  const clickHandler = jest.fn();
-  let wrapper = null;
+test('AbstractAction renders a DropdownItem', () => {
+  const { container } = render(<AbstractAction {...makeProps()}/>);
+  expect(container.querySelector('.dropdown-item')).not.toBeNull();
+});
 
-  beforeEach(() => {
-    wrapper = mount(
-      <AbstractAction
-        onClick={clickHandler}
-        title="My abstract action"
-        disabled={false}
-        className="foo-bar"
-        toggle={false}
-      />
-    );
-  });
+test('AbstractAction includes the title text', () => {
+  const { container } = render(<AbstractAction {...makeProps()}/>);
+  expect(container.querySelector('.dropdown-item').textContent).toBe('My abstract action');
+});
 
-  it('renders a DropdownItem', () => {
-    expect(wrapper.find('DropdownItem').length).toBe(1);
-  });
+test('AbstractAction delegates clicking to the provided handler', () => {
+  const onClick = jest.fn();
+  const { container } = render(<AbstractAction {...makeProps({ onClick })}/>);
+  fireEvent.click(container.querySelector('.dropdown-item'));
+  expect(onClick).toHaveBeenCalled();
+});
 
-  it('includes the title text', () => {
-    expect(wrapper.text()).toContain('My abstract action');
-  });
+test('AbstractAction adds provided extra classes', () => {
+  const { container } = render(<AbstractAction {...makeProps()}/>);
+  expect(container.querySelector('.dropdown-item').classList.contains('foo-bar')).toBe(true);
+});
 
-  it('delegates clicking to the provided handler', () => {
-    wrapper.find('button').simulate('click');
-    expect(clickHandler).toHaveBeenCalled();
-  });
+test('AbstractAction uses the label prop for the button label if label prop is supplied', () => {
+  const { container } = render(
+    <AbstractAction {...makeProps({
+      title: 'My title',
+      label: 'My label'
+    })}
+    />
+  );
+  expect(container.querySelector('.dropdown-item').textContent).toBe('My label');
+});
 
-  it('adds provided extra classes', () => {
-    expect(wrapper.find('button').hasClass('foo-bar')).toBe(true);
-  });
-
-  it('uses the label prop for the button label if label prop is supplied', () => {
-    const wrapperWithLabel = mount(
-      <AbstractAction
-        title="My title"
-        label="My label"
-      />
-    );
-    expect(wrapperWithLabel.find('button').text()).toBe('My label');
-  });
-
-  it('uses the title prop for the button label if label prop is not supplied', () => {
-    const wrapperWithLabel = mount(
-      <AbstractAction
-        title="My title"
-      />
-    );
-    expect(wrapperWithLabel.find('button').text()).toBe('My title');
-  });
+test('AbstractAction uses the title prop for the button label if label prop is not supplied', () => {
+  const { container } = render(
+    <AbstractAction {...makeProps({
+      title: 'My title'
+    })}
+    />
+  );
+  expect(container.querySelector('.dropdown-item').textContent).toBe('My title');
 });
--- a/client/src/components/ElementActions/tests/ArchiveAction-test.js
+++ b/client/src/components/ElementActions/tests/ArchiveAction-test.js
@@ -1,127 +1,101 @@
 /* eslint-disable import/no-extraneous-dependencies */
-/* global jest, describe, it, expect */
+/* global jest, test, describe, it, expect */
 
 import React from 'react';
 import { Component as ArchiveAction } from '../ArchiveAction';
-import Enzyme, { mount } from 'enzyme';
-import Adapter from 'enzyme-adapter-react-16';
+import { render, fireEvent } from '@testing-library/react';
 
-Enzyme.configure({ adapter: new Adapter() });
+const WrappedComponent = (props) => <div>{props.children}</div>;
+const ActionComponent = ArchiveAction(WrappedComponent);
+const jQuery = jest.fn();
+window.jQuery = jQuery;
 
-describe('ArchiveAction', () => {
-  let wrapper = null;
+function makeProps(obj = {}) {
+  return {
+    title: 'My abstract action',
+    element: {
+      id: 123,
+      isPublished: true,
+      blockSchema: { type: 'Test' }
+    },
+    isPublished: true,
+    actions: {
+      handleArchiveBlock: () => {}
+    },
+    toggle: false,
+    ...obj,
+  };
+}
+
+test('ArchiveAction renders the wrapped component', () => {
+  const { container } = render(<ActionComponent {...makeProps()}/>);
+  expect(container.querySelector('button.element-editor__actions-archive').textContent).toBe('Archive');
+});
+
+test('ArchiveAction does not archive when declining the confirmation', () => {
+  global.confirm = () => false;
   const mockMutation = jest.fn(() => new Promise((resolve) => { resolve(); }));
-  const WrappedComponent = (props) => <div>{props.children}</div>;
-  const ActionComponent = ArchiveAction(WrappedComponent);
-  const jQuery = jest.fn();
-  window.jQuery = jQuery;
+  const { container } = render(<ActionComponent {...makeProps({
+    actions: {
+      handleArchiveBlock: mockMutation
+    }
+  })}
+  />);
+  fireEvent.click(container.querySelector('button.element-editor__actions-archive'));
+  expect(mockMutation).not.toHaveBeenCalled();
+});
 
-  beforeEach(() => {
-    wrapper = mount(
-      <ActionComponent
-        title="My abstract action"
-        element={{
-          id: 123,
-          isPublished: true,
-          blockSchema: { type: 'Test' }
-        }}
-        isPublished
-        actions={{ handleArchiveBlock: mockMutation }}
-        toggle={false}
-      />
-    );
-  });
+test('ArchiveAction archives when accepting the confirmation', () => {
+  global.confirm = () => true;
+  const mockMutation = jest.fn(() => new Promise((resolve) => { resolve(); }));
+  const { container } = render(<ActionComponent {...makeProps({
+    actions: {
+      handleArchiveBlock: mockMutation
+    }
+  })}
+  />);
+  fireEvent.click(container.querySelector('button.element-editor__actions-archive'));
+  expect(mockMutation).toHaveBeenCalled();
+});
 
-  it('renders the wrapped component', () => {
-    expect(wrapper.children().first().type()).toEqual(WrappedComponent);
-  });
+test('ArchiveAction indicates that the block will be sent to archive when it is unpublished', () => {
+  const mockConfirm = jest.fn();
+  global.confirm = mockConfirm;
+  const { container } = render(<ActionComponent {...makeProps({
+    isPublished: false,
+  })}
+  />);
+  fireEvent.click(container.querySelector('button.element-editor__actions-archive'));
+  expect(mockConfirm).toHaveBeenCalledWith('Are you sure you want to send this block to the archive?');
+});
 
-  it('renders a button', () => {
-    expect(wrapper.find('button').length).toBe(1);
-  });
+test('ArchiveAction indicates that the block will be sent to archive when it is published', () => {
+  const mockConfirm = jest.fn();
+  global.confirm = mockConfirm;
+  const { container } = render(<ActionComponent {...makeProps({
+    isPublished: true,
+  })}
+  />);
+  fireEvent.click(container.querySelector('button.element-editor__actions-archive'));
+  expect(mockConfirm).toHaveBeenCalledWith('Warning: This block will be unpublished before being sent to the archive. Are you sure you want to proceed?');
+});
 
-  it('renders the title and class', () => {
-    expect(wrapper.find('button').text()).toContain('Archive');
-    expect(wrapper.find('button').hasClass('element-editor__actions-archive')).toBe(true);
-  });
+test('ArchiveAction is disabled when user doesn\'t have correct permissions', () => {
+  const { container } = render(<ActionComponent {...makeProps({
+    element: {
+      canDelete: false
+    }
+  })}
+  />);
+  expect(container.querySelector('button.element-editor__actions-archive').disabled).toBe(true);
+});
 
-  it('does not archive when declining the confirmation', () => {
-    global.confirm = () => false;
-    wrapper.find('button').simulate('click');
-    expect(mockMutation).not.toHaveBeenCalled();
-  });
-
-  it('archives when accepting the confirmation', () => {
-    global.confirm = () => true;
-    wrapper.find('button').simulate('click');
-    expect(mockMutation).toHaveBeenCalled();
-  });
-
-  it('indicates that the block will be sent to archive', () => {
-    const unpublishedWrapper = mount(
-      <ActionComponent
-        title="My abstract action"
-        element={{
-          id: 123,
-          isPublished: false,
-          blockSchema: { type: 'Test' }
-        }}
-        actions={{ handleArchiveBlock: mockMutation }}
-        toggle={false}
-      />
-    );
-    const mockConfirm = jest.fn();
-    global.confirm = mockConfirm;
-
-    unpublishedWrapper.find('button').simulate('click');
-    expect(mockConfirm).toHaveBeenCalledWith(
-      'Are you sure you want to send this block to the archive?'
-    );
-  });
-
-  it('indicates that the block will be unpublished before archiving', () => {
-    const mockConfirm = jest.fn();
-    global.confirm = mockConfirm;
-
-    wrapper.find('button').simulate('click');
-    expect(mockConfirm).toHaveBeenCalledWith(expect.stringContaining(
-      'Warning: This block will be unpublished'
-    ));
-  });
-
-  it('is disabled when user doesn\'t have correct permissions', () => {
-    const archiveWrapper = mount(
-      <ActionComponent
-        title="My abstract action"
-        element={{
-          ID: 123,
-          IsPublished: false,
-          BlockSchema: { type: 'Test' },
-          canDelete: false
-        }}
-        actions={{ handleArchiveBlock: mockMutation }}
-        toggle={false}
-      />
-    );
-
-    expect(archiveWrapper.find('button').first().prop('disabled')).toBe(true);
-  });
-
-  it('renders a button even when block is broken', () => {
-    wrapper = mount(
-      <ActionComponent
-        title="My unpublish action"
-        element={{
-          id: 123,
-          isPublished: true,
-          blockSchema: { type: 'Test' }
-        }}
-        isPublished
-        type={{ broken: true }}
-        actions={{ handleArchiveBlock: mockMutation }}
-        toggle={false}
-      />
-    );
-    expect(wrapper.find('button').length).toBe(1);
-  });
+test('ArchiveAction renders a button even when block is broken', () => {
+  const { container } = render(<ActionComponent {...makeProps({
+    type: {
+      broken: true
+    }
+  })}
+  />);
+  expect(container.querySelector('button.element-editor__actions-archive')).not.toBeNull();
 });
--- a/client/src/components/ElementActions/tests/DuplicateAction-test.js
+++ b/client/src/components/ElementActions/tests/DuplicateAction-test.js
@@ -1,70 +1,56 @@
 /* eslint-disable import/no-extraneous-dependencies */
-/* global jest, describe, it, expect */
+/* global jest, test, describe, it, expect */
 
 import React from 'react';
 import { Component as DuplicateAction } from '../DuplicateAction';
-import Enzyme, { mount } from 'enzyme';
-import Adapter from 'enzyme-adapter-react-16';
+import { render } from '@testing-library/react';
 
-Enzyme.configure({ adapter: new Adapter() });
+const WrappedComponent = (props) => <div>{props.children}</div>;
+const ActionComponent = DuplicateAction(WrappedComponent);
 
-describe('DuplicateAction', () => {
-  let wrapper = null;
-  const mockMutation = jest.fn(() => new Promise((resolve) => { resolve(); }));
-  const WrappedComponent = (props) => <div>{props.children}</div>;
-  const ActionComponent = DuplicateAction(WrappedComponent);
+function makeProps(obj = {}) {
+  return {
+    title: 'My duplicate action',
+    element: {
+      ID: 123,
+      BlockSchema: { type: 'Test' },
+      canCreate: true,
+    },
+    type: {
+      broken: false
+    },
+    actions: {
+      handlePublishBlock: () => {}
+    },
+    toggle: false,
+    ...obj,
+  };
+}
 
-  beforeEach(() => {
-    wrapper = mount(
-      <ActionComponent
-        title="My duplicate action"
-        element={{
-          ID: 123,
-          BlockSchema: { type: 'Test' },
-          canCreate: true
-        }}
-        type={{ broken: false }}
-        actions={{ handlePublishBlock: mockMutation }}
-        toggle={false}
-      />
-    );
-  });
+test('DuplicateAction renders a button', () => {
+  const { container } = render(<ActionComponent {...makeProps()}/>);
+  expect(container.querySelector('button.element-editor__actions-duplicate')).not.toBeNull();
+});
 
-  it('renders a button', () => {
-    expect(wrapper.find('button').length).toBe(1);
-  });
+test('DuplicateAction is disabled when user doesn\'t have correct permissions', () => {
+  const { container } = render(
+    <ActionComponent {...makeProps({
+      element: {
+        BlockSchema: { type: 'Test' },
+        canCreate: false
+      }
+    })}
+    />
+  );
+  expect(container.querySelector('button.element-editor__actions-duplicate').disabled).toBe(true);
+});
 
-  it('is disabled when user doesn\'t have correct permissions', () => {
-    const duplicateWrapper = mount(
-      <ActionComponent
-        title="My duplicate action"
-        element={{
-          BlockSchema: { type: 'Test' },
-          canCreate: false
-        }}
-        type={{ broken: false }}
-        actions={{ handleDuplicateBlock: mockMutation }}
-        toggle={false}
-      />
-    );
-
-    expect(duplicateWrapper.find('button').first().prop('disabled')).toBe(true);
-  });
-
-  it('does not render a button when block is broken', () => {
-    wrapper = mount(
-      <ActionComponent
-        title="My duplicate action"
-        element={{
-          ID: 123,
-          BlockSchema: { type: 'Test' },
-          canCreate: true
-        }}
-        type={{ broken: true }}
-        actions={{ handleDuplicateBlock: mockMutation }}
-        toggle={false}
-      />
-    );
-    expect(wrapper.find('button').length).toBe(0);
-  });
+test('DuplicateAction does not render a button when block is broken', () => {
+  const { container } = render(
+    <ActionComponent {...makeProps({
+      type: { broken: true }
+    })}
+    />
+  );
+  expect(container.querySelector('button.element-editor__actions-duplicate')).toBeNull();
 });
--- a/client/src/components/ElementActions/tests/PublishAction-test.js
+++ b/client/src/components/ElementActions/tests/PublishAction-test.js
@@ -1,5 +1,8 @@
 /* eslint-disable import/no-extraneous-dependencies */
-/* global jest, describe, it, expect, window */
+/* global jest, test, describe, it, expect, window */
+import React from 'react';
+import { Component as PublishAction } from '../PublishAction';
+import { fireEvent, render } from '@testing-library/react';
 
 jest.mock('isomorphic-fetch', () =>
   () => Promise.resolve({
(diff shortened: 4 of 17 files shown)


client/src/components/ElementActions/tests/SaveAction-test.js
```diff
@@ -1,76 +1,79 @@
...
+//         }}
+//         type={{ broken: true }}
+//         expandable
+//         actions={{ handlePublishBlock: mockMutation }}
+//         toggle={false}
+//       />
+//     );
+//     expect(wrapper.find('button').length).toBe(0);
+//   });
+// });
```
GuySartorelli, line 79: I'm not comfortable having tests that are just commented out. There is clear evidence in other places in the codebase of "todo" comments and commented out code that were meant to be resolved later that just never got resolved. Having commented out tests is the same - we may intend to get back to it, but it's entirely possible if not likely that the task for doing so will just keep being deprioritised until it's forgotten entirely.

@maxime-rainville I'll defer to your judgment on this as mentioned in standup.

package.json
```diff
@@ -35,6 +35,7 @@
   "dependencies": {
     "@apollo/client": "^3.7.1",
     "@popperjs/core": "^2.11.6",
+    "@testing-library/react": "^14.0.0",
```
GuySartorelli: testing library should be in `devDependencies`

client/src/components/ElementActions/tests/AbstractAction-test.js
```diff
@@ -1,62 +1,60 @@
...
+    disabled: false,
+    className: 'foo-bar',
+    toggle: false,
+    ...obj,
+  };
+}
+
+test('AbstractAction renders a DropdownItem', () => {
+  const { container } = render(<AbstractAction {...makeProps()}/>);
+  expect(container.querySelector('.dropdown-item')).not.toBeNull();
```
GuySartorelli, line 21: The original test explicitly checked for 1 item. We should do the same here, if we only expect one item. I think this is important because there's no other way for us to detect duplicate buttons (behat won't flag it, for example)

client/src/components/ElementActions/tests/ArchiveAction-test.js
```diff
@@ -1,127 +1,109 @@
...
+    },
+    toggle: false,
+    ...obj,
+  };
+}
+
+test('ArchiveAction renders the wrapped component', () => {
+  const { container } = render(<ActionComponent {...makeProps()}/>);
+  expect(container.querySelector('button.element-editor__actions-archive').textContent).toBe('Archive');
+});
```
GuySartorelli, line 33: This seems more like a mashup of "renders a button" and "renders the title and class" - what it isn't is "renders the wrapped component" since renders the wrapped component would be checking if the _div_ is there.

I don't think it's worth checking for the wrapped component, but if you are gonna have a test for it, you'll need to add a css class or ID to the div and check that the div is rendered.

The button, class, and title checking tests can be mixed together, but the name of the test needs to make it clear that that's what we're testing for - and it should include the length === 1 check.

client/src/components/ElementActions/tests/ArchiveAction-test.js
```diff
@@ -1,127 +1,109 @@
...
-    expect(mockConfirm).toHaveBeenCalledWith(expect.stringContaining(
-      'Warning: This block will be unpublished'
-    ));
-  });
+test('ArchiveAction indicates that the block will be sent to archive', () => {
+  const mockConfirm = jest.fn();
+  global.confirm = mockConfirm;
+  const mockMutation = jest.fn(() => new Promise((resolve) => { resolve(); }));
+  const { container } = render(<ActionComponent {...makeProps({
+    isPublished: false,
```
GuySartorelli, line 65: In the original test, `isPublished: false` is set on the _element_ prop, but here it's set as a root-level prop. What's the reason for this change?

client/src/components/ElementActions/tests/ArchiveAction-test.js
```diff
@@ -1,127 +1,109 @@
...
-        actions={{ handleArchiveBlock: mockMutation }}
-        toggle={false}
-      />
-    );
+test('ArchiveAction indicates that the block will be sent to archive', () => {
+  const mockConfirm = jest.fn();
+  global.confirm = mockConfirm;
+  const mockMutation = jest.fn(() => new Promise((resolve) => { resolve(); }));
+  const { container } = render(<ActionComponent {...makeProps({
+    isPublished: true,
```
GuySartorelli, line 76:
```suggestion
```
Not needed - that's already in the default props.

client/src/components/ElementActions/tests/ArchiveAction-test.js
```diff
@@ -1,127 +1,109 @@
...
 
-    wrapper.find('button').simulate('click');
-    expect(mockConfirm).toHaveBeenCalledWith(expect.stringContaining(
-      'Warning: This block will be unpublished'
-    ));
-  });
+test('ArchiveAction indicates that the block will be sent to archive', () => {
+  const mockConfirm = jest.fn();
+  global.confirm = mockConfirm;
+  const mockMutation = jest.fn(() => new Promise((resolve) => { resolve(); }));
```
GuySartorelli: Mock mutation isn't referenced in this test, nor in the one below.... should this just be set as a default prop instead?

client/src/components/ElementActions/tests/ArchiveAction-test.js
```diff
@@ -1,127 +1,109 @@
...
-          ID: 123,
-          IsPublished: false,
-          BlockSchema: { type: 'Test' },
-          canDelete: false
-        }}
-        actions={{ handleArchiveBlock: mockMutation }}
-        toggle={false}
-      />
-    );
+test('ArchiveAction indicates that the block will be sent to archive', () => {
```
GuySartorelli: This is the same name as the test above it. The name should make it clear what's different about this test.

client/src/components/ElementActions/tests/ArchiveAction-test.js
GuySartorelli, line 1: The "indicates that the block will be unpublished before archiving" test is missing.

client/src/components/ElementActions/tests/DuplicateAction-test.js
```diff
@@ -1,70 +1,56 @@
...
+  expect(container.querySelector('button.element-editor__actions-duplicate')).not.toBeNull();
+});
 
-    expect(duplicateWrapper.find('button').first().prop('disabled')).toBe(true);
-  });
+test('DuplicateAction renders a button', () => {
+  const { container } = render(
+    <ActionComponent {...makeProps({
+      element: {
+        BlockSchema: { type: 'Test' },
```
GuySartorelli, line 39:
```suggestion
```
That's already in the default props

client/src/components/ElementActions/tests/DuplicateAction-test.js
```diff
@@ -1,70 +1,56 @@
...
+}
+
+test('DuplicateAction renders a button', () => {
+  const { container } = render(<ActionComponent {...makeProps()}/>);
+  expect(container.querySelector('button.element-editor__actions-duplicate')).not.toBeNull();
+});
 
-    expect(duplicateWrapper.find('button').first().prop('disabled')).toBe(true);
-  });
+test('DuplicateAction renders a button', () => {
```
GuySartorelli:
```suggestion
test('DuplicateAction is disabled when user doesn\'t have correct permissions', () => {
```

client/src/components/ElementActions/tests/ArchiveAction-test.js
GuySartorelli: This applies in other test files as well (e.g. for publish action)

client/src/components/ElementActions/tests/PublishAction-test.js
```diff
@@ -1,109 +1,93 @@
...
+  const { container } = render(
+    <ActionComponent {...makeProps({
+      actions: {
+        handlePublishBlock: mockMutation
+      }
+    })}
+    />
+  );
+  fireEvent.click(container.querySelector('button.element-editor__actions-publish'));
+  await new Promise((resolve) => resolve());
```
GuySartorelli, line 55: Please add the comment back in before this promise resolution - it's super unclear what this is for otherwise and looks like it should be removed.

client/src/components/ElementActions/tests/SaveAction-test.js
GuySartorelli:
```suggestion
```
This looks like it is just the old version of the test above.

client/src/components/ElementActions/tests/UnpublishAction-test.js
```diff
@@ -1,95 +1,88 @@
...
-    expect(wrapper.find('button').length).toBe(1);
-  });
-
-  it('renders the title and class', () => {
-    expect(wrapper.find('button').text()).toContain('Unpublish');
-    expect(wrapper.find('button').hasClass('element-editor__actions-unpublish')).toBe(true);
-  });
+function makeProps(obj = {}) {
+  return {
+    title: 'My abstract action',
```
GuySartorelli:
```suggestion
    title: 'My unpublish action',
```
I got confused as to what this was testing for a second because of this. 😅

client/src/components/ElementActions/tests/UnpublishAction-test.js
```diff
@@ -1,95 +1,88 @@
...
+  expect(mockMutation).toHaveBeenCalledWith(123);
+});
 
-    expect(unpublishWrapper.find('button').first().prop('disabled')).toBe(true);
-  });
+test('UnpublishAction is disabled when user doesn\'t have correct permissions', () => {
+  const { container } = render(
+    <ActionComponent {...makeProps({
+      element: {
+        isPublished: true,
```
GuySartorelli, line 69:
```suggestion
```
That's already in the default props

client/src/components/ElementEditor/tests/Content-test.js
```diff
@@ -1,64 +1,50 @@
...
-      expect(wrapper.find(InlineEditFormComponent)).toHaveLength(1);
-    });
+import { render } from '@testing-library/react';
+
+function makeProps(obj = {}) {
+  return {
+    fileUrl: '/ss4/assets/Uploads/c70617f2e4/sample__FillWzEwMCwxMDBd.jpeg',
+    fileTitle: '',
+    content: '',
+    previewExpanded: false,
```
GuySartorelli, line 13: Not a big deal and I won't hold back merging for this - but `previewExpanded` is `true` in more tests than it is `false` so may be worth considering inverting this value.

client/src/components/ElementEditor/tests/Content-test.js
```diff
@@ -1,64 +1,50 @@
...
+  expect(container.querySelector('.element-editor-content')).not.toBeNull();
+  expect(container.querySelector('.test-inline-edit-form')).not.toBeNull();
+  expect(container.querySelector('.test-summary')).toBeNull();
+});
 
-      expect(wrapper.type()).toEqual('div');
-      expect(wrapper.find(SummaryComponent)).toHaveLength(0);
-    });
-  });
+test('Content returns a div when no content or image is provided', () => {
```
GuySartorelli, line 39: This test title doesn't describe what's actually being tested, which makes this hard to review (is the name wrong, or is the test wrong?). Please use a title that's descriptive of what's being tested - or update the test if it's not testing the right thing.

client/src/components/ElementActions/tests/AbstractAction-test.js
GuySartorelli: This also applies with many other assertions throughout this PR

client/src/components/ElementEditor/tests/Element-test.js
```diff
@@ -1,153 +1,114 @@
...
-      );
+test('Element should render the HeaderComponent and the ContentComponent', () => {
+  const { container } = render(<Element {...makeProps()}/>);
+  expect(container.querySelector('.element-editor__element .test-header')).not.toBeNull();
+  expect(container.querySelector('.element-editor__element .test-content')).not.toBeNull();
+});
 
-      expect(wrapper.hasClass('element-editor__element--draft')).toBe(true);
-    });
+test('Element should render null if no ID is given', () => {
```
GuySartorelli, line 48: Based on the name of this test, and the `expect(wrapper.type()).toBeNull();` from the original test, it seems like the whole `<Element>` component should render nothing at all in this scenario (not just not rendering the header and content components). Is there a way to test that?

client/src/components/ElementEditor/tests/Element-test.js
```diff
@@ -1,153 +1,114 @@
...
+  const { container } = render(
+    <Element {...makeProps({
+      element: {
+        ...makeProps().element,
+        isPublished: false
+      }
+    })}
+    />
+  );
+  expect(container.querySelector('.element-editor__element--draft')).not.toBeNull();
```
GuySartorelli, line 85: We should be checking that the original element has this class - so `container.querySelector('.element-editor__element')` classnames should contain `element-editor__element--draft`
(i.e. it's on the main element and not some child element)

GuySartorelli: Same with the published and modified tests below

client/src/components/ElementEditor/tests/ElementEditor-test.js
GuySartorelli, line 1: Seem to be missing two tests:
- "should filter all element types by those allowed for this editor"
- "should retain the order specified by the allowed elements config"

client/src/components/ElementEditor/tests/ElementList-test.js
```diff
@@ -1,116 +1,88 @@
...
-          HoverBarComponent={HoverBar}
-          loading={false}
-          areaId={1}
-          connectDropTarget={connectDropTarget}
-        />
-      );
+test('ElementList renders a loading component', () => {
+  const { container } = render(
+    <ElementList {...makeProps({
+      key: '2',
```
GuySartorelli, line 65: Pretty sure the actual key value doesn't matter, so the default should be fine here. Super minor though, won't hold off merging for this.
```suggestion
```

client/src/components/ElementEditor/tests/ElementList-test.js
```diff
@@ -1,116 +1,88 @@
...
-      expect(wrapper.find(Loading).length).toBe(0);
-      expect(wrapper.find('Add blocks to place your content').length).toBe(0);
-    });
-  });
+test('ElementList renders a placeholder message when no elements are provided as props', () => {
+  const { container } = render(
+    <ElementList {...makeProps({
+      key: '3',
+      blocks: [],
+      loading: false
```
GuySartorelli, line 80:
```suggestion
```
This is already in default props.

client/src/components/ElementEditor/tests/Header-test.js
```diff
@@ -1,307 +1,217 @@
...
-        />
-      );
+test('Header should render the title', () => {
+  const { container } = render(
+    <Header {...makeProps({
+      element: {
+        id: '12',
+        title: 'Sample File Block'
+      }
+    })}
```
GuySartorelli, line 62:
```suggestion
    <Header {...makeProps()}
```
The ID doesn't seem important (you're not checking the ID of the element) and the title is the same as the default.

client/src/components/ElementEditor/tests/Header-test.js
```diff
@@ -1,307 +1,217 @@
...
+    />
+  );
+  expect(container.querySelector('.element-editor-header__title').textContent).toBe('Sample File Block');
+});
 
-      const expandButton = wrapper.find('.element-editor-header__expand');
-      expect(expandButton.length).toBe(0);
-    });
+test('Header should render the title for broken elements', () => {
```
GuySartorelli, line 68:
```suggestion
test('Header should override the title for broken elements', () => {
```
The fact that the title is being overridden (not just using the title of the original element) is important here.

client/src/components/ElementEditor/tests/Header-test.js
```diff
@@ -1,307 +1,217 @@
...
-          connectDragPreview={content => content}
-          onDragEnd={() => {}}
-        />
-      );
+test('Header should contain a Tooltip', async () => {
+  const { container } = render(
+    <Header {...makeProps({
+      element: {
+        id: '13',
+        title: 'Sample File Block',
```
GuySartorelli, line 86:
```suggestion
```
That's the default title in the props.
Also, we can probably avoid changing the ID as well - either set `13` as the default ID (we don't use the ID in any other test except where its value is 13) or just change the `13` in the query selector to `0`.

GuySartorelli: I won't block merging for this, but it seems worth keeping these tests as simple as possible so it's easier to see what's actually important about what's being changed and what's not.

client/src/components/ElementEditor/tests/Header-test.js
```diff
@@ -1,307 +1,217 @@
...
+      }
+    })}
+    />
+  );
+  fireEvent.mouseOver(container.querySelector('#element-icon-13.font-icon-block-file'));
+  const tooltip = await screen.findByText('File');
+  expect(tooltip.getAttribute('role')).toBe('tooltip');
```
GuySartorelli, line 93: This feels a bit backwards - can we find it by its role instead, and then check the text is "File"?
Not a big deal, won't block merging either way.

GuySartorelli: Same with test below - seems more sensible and more explicit to query by role.

GuySartorelli, line 93: This validates whether a tooltip will appear - but the original test (and for the one below) are checking whether a tooltip exists _at all_ - might be worth doing both, i.e. validate that the element exists in the DOM at all, and then check whether it appears on hover.
This will especially make the one below more robust, I think.

client/src/components/ElementEditor/tests/Header-test.js
```diff
@@ -1,307 +1,217 @@
...
+  })}
+  />);
+  expect(container.querySelector('.element-editor-header__expand.font-icon-right-open-big')).not.toBeNull();
+});
 
-      expect(wrapper.text()).toContain('ElementActionsComponent');
-    });
+test('Header should render a down caret button when not expanded', () => {
+  const { container } = render(<Header {...makeProps({
+    expandable: true,
```
GuySartorelli, line 128:
```suggestion
```
That's already the default. Same with other tests below.

client/src/components/ElementEditor/tests/Header-test.js
```diff
@@ -1,307 +1,217 @@
...
-          connectDragPreview={content => content}
-          onDragEnd={() => {}}
-        />
-      );
+test('Header should render a right caret button when not expandable', () => {
+  const { container } = render(<Header {...makeProps({
+    expandable: false
+  })}
+  />);
+  expect(container.querySelector('.element-editor-header__expand.font-icon-right-open-big')).not.toBeNull();
```
GuySartorelli, line 123: With this, and all the other caret tests, it's important to check independently that:
1. only _one_ `.element-editor-header__expand` element exists
2. that element has the relevant class for down, right, etc.


This validates that we don't accidentally have a right caret rendering when a down caret is rendered, which isn't being explicitly tested, but is implicitly tested with the `toBe(1)` assertion on `.element-editor-header__expand`.

client/src/components/ElementEditor/tests/Header-test.js
```diff
@@ -1,307 +1,217 @@
...
+});
 
-      const versionedState = wrapper.find('.element-editor-header__version-state');
-      expect(versionedState.prop('title')).toContain('has unpublished changes');
-      expect(versionedState.hasClass('element-editor-header__version-state--modified')).toBe(true);
-    });
+test('Header should render a versioned state message when the element is not published', () => {
+  const { container } = render(<Header {...makeProps({
+    element: {
+      id: '14',
```
GuySartorelli, line 186:
```suggestion
```
ID doesn't seem to be relevant for this or the below tests.

client/src/components/ElementEditor/tests/Header-test.js
```diff
@@ -1,307 +1,217 @@
...
+    }
+  })}
+  />);
+  expect(container.querySelector('.element-editor-header__version-state--modified').getAttribute('title')).toContain('has unpublished changes');
+});
 
-      expect(wrapper.find('.element-editor-header__version-state').length).toBe(0);
-    });
-  });
+test('Header should render a versioned state message when the element is published', () => {
```
GuySartorelli, line 207:
```suggestion
test('Header should not render a versioned state message when the element is published', () => {
```

client/src/components/ElementEditor/tests/Header-test.js
```diff
@@ -1,307 +1,217 @@
...
+test('Header should render a versioned state message when the element is modified', () => {
+  const { container } = render(<Header {...makeProps({
+    element: {
+      id: '14',
+      isPublished: true,
+      isLiveVersion: false
+    }
+  })}
+  />);
+  expect(container.querySelector('.element-editor-header__version-state--modified').getAttribute('title')).toContain('has unpublished changes');
```
GuySartorelli, line 204:
```suggestion
  expect(container.querySelector('.element-editor-header__version-state.element-editor-header__version-state--modified').getAttribute('title')).toContain('has unpublished changes');
```
The `.element-editor-header__version-state` class is relevant, because that's what we're checking _isn't_ rendered in the below test.
This applies for the above test as well.

client/src/components/ElementEditor/tests/HoverBar-test.js
```diff
@@ -1,46 +1,47 @@
...
+test('HoverBar renders top HoverBarComponent', () => {
+  [
+    { areaId: 1, elementId: 0 },
+    { areaId: 2, elementId: 3 },
+  ].forEach(({ areaId, elementId }) => {
+    const { container } = render(
+      <HoverBar {...makeProps({
+        areaId,
+        elementId,
+        key: elementId
```
GuySartorelli, line 38:
```suggestion
```
Pretty sure the key isn't actually relevant here. I won't block merging for this though - and I also could just be wrong 😅

client/src/components/ElementEditor/tests/HoverBar-test.js
```diff
@@ -1,46 +1,47 @@
...
+      })}
+      />
+    );
+    const hoverBar = container.querySelector('.element-editor__hover-bar');
+    const hoverBarArea = container.querySelector('button.element-editor__hover-bar-area');
+    expect(hoverBar.getAttribute('id')).toBe(`${hoverBarName}_${areaId}_${elementId}`);
+    expect(hoverBarArea.getAttribute('id')).toBe(`${hoverBarAreaName}_${areaId}_${elementId}`);
```
GuySartorelli, line 45: This isn't checking the `target` and `container` props - I understand that props are internal and can't be tested like that anymore, but the `target` at least is presumably being rendered out as some attribute we can test against, and that seems important (i.e. when clicking on the bar, the appropriate element is targetted). The container also seems important, as I think that's what's used to determine which elemental area to add new blocks to if there is more than one in the form.

GuySartorelli: Note that I do recall having a bug related to blocks not being added to the correct elemental area since I've been here, so this does seem worth checking.

client/src/components/ElementEditor/tests/Summary-test.js
```diff
@@ -1,61 +1,45 @@
...
+    fileUrl: '/ss4/assets/Uploads/c70617f2e4/sample__FillWzEwMCwxMDBd.jpeg',
+    fileTitle: 'Sample Image',
+    content: 'Sample content',
+    ...obj
+  };
+}
+
+test('Summary should render an image if the fileUrl prop is provided', () => {
+  const { container } = render(<Summary {...makeProps()}/>);
+  expect(container.querySelector('.element-editor-summary__thumbnail-image')).not.toBeNull();
```
GuySartorelli, line 19: I think it's important here to check for an actual `img` element here, since we want to check an image is actually rendered in the DOM. Bonus points for validating the correct `src` even though the original didn't do that.

client/src/components/ElementEditor/tests/Summary-test.js
```diff
@@ -1,61 +1,45 @@
...
-    const wrapper = shallow(
-      <Summary
-        fileUrl="/ss4/assets/Uploads/c70617f2e4/sample__FillWzEwMCwxMDBd.jpeg"
-        fileTitle=""
-        content=""
-      />
-    );
+test('Summary should render a content summary if the content is provided', () => {
+  const { container } = render(<Summary {...makeProps()}/>);
+  expect(container.querySelector('.element-editor-summary__content')).not.toBeNull();
```
GuySartorelli, line 34: I know the original doesn't do this so I won't hold back merging if you don't make this change, but it would be good to test that the _correct_ content is in the summary.

client/src/components/HistoricElementView/tests/HistoricElementView-test.js
```diff
@@ -1,54 +1,41 @@
...
-    ElementTitle: 'Pretend Element',
-    ElementEditLink: 'http://localhost:8080/'
+function makeProps(obj = {}) {
+  return {
+    data: {
+      ElementID: 1,
+      ElementType: 'Stub',
+      ElementTitle: 'Pretend Element',
+      ElementEditLink: 'http://localhost:8080/',
+      tag: 'div'
```
GuySartorelli, line 14: Do you recall why you added `ElementID` and `tag` which aren't in the original test?
No biggie if not, that's more me being curious than anything.

client/src/components/HistoricElementView/tests/HistoricElementView-test.js
```diff
@@ -1,54 +1,41 @@
...
+const HistoricElementView = ElementalAreaHistoryFactory(FieldGroupStub);
 
-    it('should should have a link in the header', () => {
-      const link = ReactTestUtils.findRenderedDOMComponentWithClass(history, 'elemental-preview__link');
-      expect(link).toBeTruthy();
-      expect(link.href).toBeTruthy();
-      expect(link.textContent).toEqual('Block history');
-    });
-  });
+test('HistoricElementView render ', () => {
```
GuySartorelli, line 34: The space at the end there makes me wonder if you had meant to do some more work on this one (either come up with a more descriptive title or split this back out into the individual well-described tests from the original)?
If not, then this is fine.

GuySartorelli (changes requested):

client/src/components/ElementActions/tests/PublishAction-test.js
emteknetnz: The test will fail if it's removed, that will prevent people from removing it :-)

This is a type of thing I've done loads of times. Best guess as to why I did it is because RTL complained that something wasn't "wrapped in act()" because there was some component state update than happened then a render, or something like that. 

I could put in a note as to why this is done, but it's done it loads of jest tests therefore either I document why it was done like this everywhere, or nowhere.

I think the simple fact that you can delete it, run the test and see the bizarre failure is enough explanation of why we need this code. I myself cannot really articulate the why because thing is really convulated

GuySartorelli: I don't like that it's a bizarre failure, a comment would save future devs some time debugging that - but I take your point that this is in multiple places, so a comment here won't be so useful. Given that context, this is fine as is.

emteknetnz:
@GuySartorelli Rather than respond to all of the feedback individually, I've responded below to some common types of feedback below. This is applicable to all the other PRs because the same patterns of things will show up many times so I'm highlighting them now before all the other peer review is done.

This means I'll be leaving lots of your individual pieces of feedback unanswered

### Using querySelectorAll() + .length `.toHaveLenth(n)` rather than querySelector() + .not.toBeNull();

RTL philosophy is for tests to resemble how users interact with the application, which is somewhat similar to behat. If you have a look at their recommended [priority for querying the rendered dom](https://testing-library.com/docs/queries/about#priority) at the top there are things like `getByRole()` e.g. `getByRole('button', {name: /submit/i})`. I mostly haven't implemented these simply because it was easy to translate the tests to use `.container()` which then allows the use of `.querySelector()` meaning I could reuse the easy css selectors. Also done this way because when I started the process I was brand new to RTL, and this was the easiest method. If this method was listed in the priority then it would probably be ranked #4 below `getByTestId()`. With RTL's philosophy in mind, I don't think we should be using 'all()' style queries and then testing the length, it's not in the spirit of RTL. We also don't test '.length()' in behat. Eventually I'd expect that we'll migrate off the `.container()` approach to a more accessibility first query approach, probably at the same time as we're doing an accessibility first CMS epic.

### Missing assertions e.g. original test tested a prop value, or a state value. The new test doesn't test this.

I made a best effort to try and test what the prop or state was rendered to. In all cases this makes the test more complex because you need to look inside the original component and work out how the prop/state is finally rendered. There is simply a very large amount of tests and I think we need to accept there will be some loss of fidelity due to the fundamentally different way of testing. These tests are honestly quite low value (jest tests catch very few regressions) so I think we're better off just accepting the lower fidelity rather than fussing over things that seem basically unimportant.

### Test title is non-descriptive

I used the original test title most of the time so that it was easier to match up the before and after in peer review time. I think it would probably be easier to open a new card to bulk update the test titles if we think it warrants it, though personally I'm not sure they're important enough to warrant this.

### Things such as "value passed to makeProps() is already in default props", we can could the value of "key", etc

These sorts of changes are so low value IMO I'm not going to spend time fixing them. Normally I wouldn't push back on these and I'd just implement them because it's simply much faster then arguing on github over trivial things, however because there will be hundreds of examples of these in other PRs, I'd much much prefer to have this level of feedback to simply not be provided.

package.json
emteknetnz: Updated

client/src/components/ElementActions/tests/ArchiveAction-test.js
emteknetnz: Removed

client/src/components/ElementActions/tests/ArchiveAction-test.js
emteknetnz: isPublished prop on the component is what determines the confirm message, not what's on props.elements

client/src/components/ElementActions/tests/ArchiveAction-test.js
emteknetnz: Updated

client/src/components/ElementActions/tests/DuplicateAction-test.js
emteknetnz: Updated

client/src/components/ElementActions/tests/UnpublishAction-test.js
emteknetnz: Updated

client/src/components/ElementEditor/tests/Summary-test.js
emteknetnz: Presumably thumbnail-image is an img

GuySartorelli: We can be explicit instead of presuming - this is meant to be a test of what the user experiences, isn't it? It feels important to validate that the user is experiencing an image, and not just some random DOM element.

GuySartorelli:
## Using querySelectorAll() + .length .toHaveLenth(n) rather than querySelector() + .not.toBeNull();

The fact is, `querySelector()` doesn't work the same way that the `getBy()` functions do. The `getBy` versions of the functions will throw an error if more than one item is found. But `querySelector()` does _not_ throw that error, so it does _not do the same thing_ that the old test did, and it _also doesn't do the same thing_ that the `getBy()` functions will do.

Please either change these tests to use `querySelectorAll()` and check for a length of 1, or to use the appropriate `getBy()` functions - or provide a convincing argument why checking for duplicate rendered items isn't worthwhile.

> We also don't test '.length()' in behat.

These aren't behat tests - they're a much lower level set of tests checking smaller "units" of functionality for individual components. We have the opportunity to test things here that we wouldn't test in behat and I don't think we should look at end-to-end tests for a measure of how we write our unit tests.

## Missing assertions e.g. original test tested a prop value, or a state value. The new test doesn't test this.

Please reply to the individual comments for these - this is far too big a blanket, it's impossible for me to know which things I've flagged are covered by this and which you either didn't notice or forgot, etc.

## Test title is non-descriptive

The test titles are important, as they describe what is being tested and therefore give developers a clue as to how the test should behave. If a test starts failing suddenly, the name of the test can be an important clue for what needs to change so that the test passes _while still testing what it is meant to test_.

In the cases where I have flagged the name of a test, these are either scenarios where the new test _does not test the same thing as the old test_, or where the test is new, or where the title is duplicated (two tests with the same title).

In all of the scenarios above, I think it is fair for me to expect a title that is descriptive of what the test is doing, since in these cases it's not just a one-to-one match with the old tests.

## Things such as "value passed to makeProps() is already in default props", we can could the value of "key", etc

I would like to know if these were done intentionally, and if it's desirable to do it this way, since that will help me know in the future if these are intentional patterns that should be followed or if they were simply left as they are because it's not worth the effort to change. But I won't request any more changes or information about this so that we can avoid some ping pong on them.

emteknetnz:
> Using querySelectorAll() + .length .toHaveLenth(n) rather than querySelector() + .not.toBeNull();

At this point there's a vast number of these assertions to fix (there's 2x other issues containing multiple PRs each). I'd rather programmatically fix everything in one go then have to fix everything individually. If you want this done then please raise a new card to do everything in one go after all the PRs are merged.

> Missing assertions e.g. original test tested a prop value, or a state value. The new test doesn't test this.

OK. I'll reply 'Not doing' to the things I'm not doing.

This may seem like I'm just being lazy, though there's over 10,000 lines added so far in all the PRs so I'm genuinely concerned about getting burnt-out here by too many minor changes being asked for in peer review.  

> Test title is non-descriptive

Again, there's far too many to change at this point. When I was transcribing these tests I don't really know what a lot of the tests were really doing other than usually testing a class method.  So I usually just copy paste the old test title and use that.  At this point I really don't know what the tests do as I would transcribe a test and then move on to the next. You probably have just as good an idea what these tests do just by looking at the PR.

If you feel these need to be fixed, please raise a new card and it can be done retrospectively, though I doubt this will every get prioritized because it's simply too low value and because it's simply a long and boring task that I doubt anyone will want to do.

> Things such as "value passed to makeProps() is already in default props .. I would like to know if these were done intentionally"

Maybe? Maybe at the time I thought it would make the tests easier to read? In all instances though stuff like this is of trivial importance and should be ignored.

client/src/components/ElementActions/tests/AbstractAction-test.js
emteknetnz: Not doing

client/src/components/ElementActions/tests/ArchiveAction-test.js
emteknetnz: Covered by "ArchiveAction indicates that the block will be sent to archive when it is published"

client/src/components/ElementActions/tests/ArchiveAction-test.js
emteknetnz: Not doing

client/src/components/ElementEditor/tests/Content-test.js
emteknetnz: Not doing

client/src/components/ElementEditor/tests/Content-test.js
emteknetnz: Not doing

client/src/components/ElementEditor/tests/Element-test.js
emteknetnz: Not doing

client/src/components/ElementEditor/tests/Element-test.js
emteknetnz: Not doing

client/src/components/ElementEditor/tests/HoverBar-test.js
emteknetnz: Not doing

client/src/components/ElementEditor/tests/Summary-test.js
emteknetnz: Not doing

client/src/components/ElementEditor/tests/Summary-test.js
emteknetnz: Not doing

client/src/components/HistoricElementView/tests/HistoricElementView-test.js
emteknetnz: Not doing

client/src/components/ElementEditor/tests/ElementEditor-test.js
GuySartorelli: @emteknetnz Are you doing these ones?

emteknetnz: Added

client/src/components/ElementEditor/tests/Header-test.js
emteknetnz: Not doing

client/src/components/ElementEditor/tests/Header-test.js
emteknetnz: Not doing

client/src/components/ElementEditor/tests/Header-test.js
emteknetnz: Not doing

emteknetnz: Not doing

client/src/components/ElementEditor/tests/Header-test.js
emteknetnz: Not doing

client/src/components/ElementEditor/tests/Header-test.js
emteknetnz: Not doing

client/src/components/ElementEditor/tests/Header-test.js
emteknetnz: Not doing

GuySartorelli:
I've raised https://github.com/silverstripe/.github/issues/43 for the outstanding changes. I'll action it.

GuySartorelli (approved):

client/src/components/ElementEditor/tests/Summary-test.js
GuySartorelli: After looking into this, it turns out the `target` and `container` props eventually get passed down to a reactstrap component, so it's okay to not test them here.

client/src/components/ElementEditor/tests/HoverBar-test.js
GuySartorelli: After looking into this, it turns out the `target` and `container` props eventually get passed down to a reactstrap component, so it's okay to not test them here.

PR #1058, opened 2023-05-03
MNT Apply changes requested from peer review

GuySartorelli:
## Issue
- https://github.com/silverstripe/.github/issues/43

### DIFF
--- a/client/src/components/ElementEditor/tests/Header-test.js
+++ b/client/src/components/ElementEditor/tests/Header-test.js
@@ -49,7 +49,7 @@ function makeProps(obj = {}) {
 
 test('Header should render the icon', () => {
   const { container } = render(<Header {...makeProps()}/>);
-  expect(container.querySelector('i.font-icon-block-file')).not.toBeNull();
+  expect(container.querySelectorAll('i.font-icon-block-file')).toHaveLength(1);
 });
 
 test('Header should render the title', () => {
@@ -65,7 +65,7 @@ test('Header should render the title', () => {
   expect(container.querySelector('.element-editor-header__title').textContent).toBe('Sample File Block');
 });
 
-test('Header should render the title for broken elements', () => {
+test('Header should override the title for broken elements', () => {
   const { container } = render(
     <Header {...makeProps({
       type: {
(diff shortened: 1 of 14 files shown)


client/src/components/ElementEditor/tests/Header-test.js
```diff
@@ -108,19 +108,21 @@ test('Header should not contain a Tooltip for a broken element', async () => {
...
-  // wait for 500 milliseconds for the tooltip to appear (which is should not)
-  // https://testing-library.com/docs/dom-testing-library/api-async/#waitfor
-  await waitFor(() => null, { timeout: 500 });
-  // use "queryBy" because it does not throw an error on fail - https://testing-library.com/docs/queries/about
-  expect(screen.queryByText('File')).toBeNull();
+  // Normally we would use "queryByRole" here because it does not throw an error on fail - but
+  // in this case that provides a false negative result (confirmed by trying a timeout
+  // with queryByRole in the "Header should not contain a Tooltip for a broken element" test
+  // which causes that test to fail)
+  const tooltip = await screen.findByRole('tooltip', null, { onTimeout: () => {} });
```
emteknetnz: Won't this wait for the default 5000 milliseconds rather 500 milliseconds? Means test suite is now way slower

emteknetnz (changes requested):

client/src/components/ElementEditor/tests/Header-test.js
GuySartorelli: This is the only combination of this test and the previous test that gave consistent results. If, for example, I tried the `waitFor(() => null, { timeout: 500 });` and then check `screen.queryByText('File')` in the previous test, it said that element didn't exist. Which meant this test would actually _always_ pass.

emteknetnz: The test is supposed to pass though because there isn't supposed to be a tooltip? i.e. It would fail the previous test if there was a tooltip in the DOM - meaning the test is useful for detecting regressions?

Not sure what what was wrong with the previous test? Happy to change it to fileByRole() though otherwise not happy with the 5000 millisecond delay on the test.

emteknetnz: >  and then check screen.queryByText('File') in the previous test

Oh - previous testing meaning a test where there is supposed to be a tooltip? Meaning the way the test was meant it didn't detect tooltips when they were supposed to be there?

GuySartorelli: Yup. Sorry, I should have said "the test above"

emteknetnz: The updated code in the PR also doesn't seem to work with the test above

This should work:

`await screen.findByRole('tooltip', {}, { timeout: 500, onTimeout: () => null });`

GuySartorelli: Done

emteknetnz (approved):

Issue #1067, opened 2023-06-22
Optimise SiteTree search when elemental is installed


emteknetnz:
I raised this card after getting feedback from a couple of developers about poor performance in the CMS being a pain point, particularly on a large customer website - slack thread (internal) https://silverstripeltd.slack.com/archives/CLXKD9X51/p1614061096035300?thread_ts=1614061096.035300&cid=CLXKD9X51

The site tree search (magnifying glass at the top right of the "pages" seciton in the CMS) is a performance bottleneck, and gets significantly worse with elemental installed. Performance scales badly and get worse the more pages + elements there are

I had a quick look through the code path with elemental installed, specifically in [ElementSiteTreeFilterSearch](https://github.com/silverstripe/silverstripe-elemental/blob/5/src/Controllers/ElementSiteTreeFilterSearch.php) (which copy pastes a lot from [CMSSiteTreeFilter](https://github.com/silverstripe/silverstripe-cms/blob/5/code/Controllers/CMSSiteTreeFilter.php) and overrides it, rather than being an extension)
- There's a lot of PartialMatchFilter on multiple database fields, which translation to a bunch of SQL `WHERE Field LIKE '%search_term%'` on unindexed columns that contain text, as you'd expect, probably can't do much easily without adding in something like elastic specifically for CMS search, which is non-trivial
- More interestingly, there's [this](https://github.com/silverstripe/silverstripe-elemental/blob/5/src/Controllers/ElementSiteTreeFilterSearch.php#L44) seemingly unoptimized code:

```php
        // Get an array of SiteTree record IDs that match the search term in nested element data
        /** @var ArrayList $siteTrees */
        $siteTrees = $query->filterByCallback(function (SiteTree $siteTree) {
            // Filter by elemental PHP
            if (!$siteTree->hasExtension(ElementalPageExtension::class)) {
                return false;
            }

            // Check whether the search term exists in the nested page content
            $pageContent = $siteTree->getElementsForSearch();
            return stripos($pageContent ?? '', $this->params['Term'] ?? '') !== false;
        });

        if ($siteTrees->count()) {
            // Apply the list of IDs as an extra filter
            $this->extraTermFilters['ID:ExactMatch'] = $siteTrees->column('ID');
        }
```

[...]

PR #1069, opened 2023-06-27
ENH Optimise site search

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-elemental/issues/1067

There's a couple of bits to this PR:

A) Utilise eager loading - this gives a small performance gain of ~9%

B) Don't render elemental templates, instead only look at database values on the element (won't include any child DataObjects) - this gave a large performance gain, over 100% improvement

There's likely going to be a "loss in accuracy" (though may also be a gain) in exchange for the massive performance boost - the change in accuracy is a result of:
- more content being searched across, because now fields not on the rendered template will be searched against
- less content being searched across, because any child objects on the element that are normally rendered won't be search against

We'd also need to agree whether to make this an opt-in or an opt-out.  In this PR I've got it as the more conservative opt-in, though I'd prefer opt-out as most people will likely not even realise it's an option available, and I given it's "only" cms page search I don't think any possible loss in accuracy really warrants the fact that page search really is horrible slow without this optimization (I've had people tell me about 10 second response times, and even sites timing out).

Options:
- 1) opt-in in CMS 5.1, stays as opt-in for CMS 6
- 2) opt-in in CMS 5.1, change to opt-out for CMS 6
- 3) opt-out in CMS 5.1 (my recommendation)

### Benchmarks

Running a CMS search for "lorem" against 1,000 pages with 5 blocks each (total of 5,000 blocks) - data generated with https://github.com/emteknetnz/silverstripe-test-data - total of 3 runs - times are in seconds

**render elements - eager load off - (current) - avg 5.81**
5.92
5.63
5.87

**render elements - eager load on - avg 5.35**
5.35
5.21
5.50

**unrendered - eager load off - avg 2.36**
2.37
2.35
2.36

**unrendered - eager load on - avg 2.15**
2.09
2.24
2.13

### DIFF
--- a/docs/en/searching-blocks.md
+++ b/docs/en/searching-blocks.md
@@ -40,3 +40,33 @@ to make it clear in search results where one piece of content ends and another b
 Page:
   search_index_element_delimiter: ' ... '
 ```
+
+## CMS page search
+
+CMS page search will include search results for pages with elements that match the search query.
+
+By default it uses the same method as the search indexing where it will fully render every element that is
+being searched. This is an expensive operation and can cause performance issues if you have a large site with a lot of elements.
+
+To increase performance by a large amount, likely more than doubling it, you can disable the rendering of elements and instead just look at the database values of the elements directly.
+
+```yml
+DNADesign\Elemental\Controllers\ElementSiteTreeFilterSearch:
+  render_elements: false
+```
+
+If `render_elements` is `false`, then all fields that have stored as a Varchar or Text like are searched. Individual fields on blocks can be excluded from the search by adding fields to the `exclude_fields_from_cms_search` array config variable on the element class. e.g.
+
+```yml
+App\MyElement:
+  fields_excluded_from_cms_search:
+    - MyFieldToExclude
+    - AnotherFieldToExclude
+```
+
+If the above is still not performant enough, searching elements for content in CMS page search can be disabled entirely:
+
+```yml
+DNADesign\Elemental\Controllers\ElementSiteTreeFilterSearch:
+  search_for_term_in_content: false
+```
--- a/src/Controllers/ElementSiteTreeFilterSearch.php
+++ b/src/Controllers/ElementSiteTreeFilterSearch.php
@@ -21,6 +21,11 @@ class ElementSiteTreeFilterSearch extends CMSSiteTreeFilter_Search
     private static $search_for_term_in_content = true;
 
     /**
+     * Whether to render elements with templates when doing a CMS SiteTree search
+     */
+    private static bool $render_elements = true;
+
+    /**
      * @var array
      */
     private $extraTermFilters = [];
@@ -47,8 +52,13 @@ class ElementSiteTreeFilterSearch extends CMSSiteTreeFilter_Search
                 return false;
             }
 
-            // Check whether the search term exists in the nested page content
-            $pageContent = $siteTree->getElementsForSearch();
+            if ($this->config()->get('render_elements') === true) {
+                // Check whether the search term exists in the nested page content
+                $pageContent = $siteTree->getElementsForSearch();
+            } else {
+                $pageContent = $siteTree->getContentFromElementsForCmsSearch();
+            }
+
             return stripos($pageContent ?? '', $this->params['Term'] ?? '') !== false;
         });
 
--- a/src/Extensions/ElementalPageExtension.php
+++ b/src/Extensions/ElementalPageExtension.php
@@ -39,6 +39,13 @@ class ElementalPageExtension extends ElementalAreasExtension
     private static $search_index_element_delimiter = ' ';
 
     /**
+     * Used to cache all ElementalArea's prior to eager loading elements
+     *
+     * @internal
+     */
+    private static ?array $elementalAreas = null;
+
+    /**
      * Returns the contents of each ElementalArea has_one's markup for use in Solr or Elastic search indexing
      *
      * @return string
@@ -49,14 +56,17 @@ class ElementalPageExtension extends ElementalAreasExtension
         SSViewer::set_themes(SSViewer::config()->get('themes'));
         try {
             $output = [];
-            $this->loopThroughElements(function (BaseElement $element) use (&$output) {
-                if ($element->getSearchIndexable()) {
-                    $content = $element->getContentForSearchIndex();
-                    if ($content) {
-                        $output[] = $content;
-                    }
+            $elements = $this->getEagerLoadedElements();
+            /** @var BaseElement $element */
+            foreach ($elements as $element) {
+                if (!$element->getSearchIndexable()) {
+                    continue;
                 }
-            });
+                $content = $element->getContentForSearchIndex();
+                if ($content) {
+                    $output[] = $content;
+                }
+            }
         } finally {
             // Reset theme if an exception occurs, if you don't have a
             // try / finally around code that might throw an Exception,
@@ -64,6 +74,28 @@ class ElementalPageExtension extends ElementalAreasExtension
             SSViewer::set_themes($oldThemes);
         }
         return implode($this->owner->config()->get('search_index_element_delimiter') ?? '', $output);
+    }
+
+    /**
+     * Returns the contents of all Elements on the pages ElementalAreas for use in CMS search
+     */
+    public function getContentFromElementsForCmsSearch(): string
+    {
+        $output = [];
+        $elements = $this->getEagerLoadedElements();
+        /** @var BaseElement $element */
+        foreach ($elements as $element) {
+            if (!$element->getSearchIndexable()) {
+                continue;
+            }
+            $content = $element->getContentForCmsSearch();
+            if ($content) {
+                $output[] = $content;
+            }
+        }
+        // Use |%| to delimite different elements rather than space so that you don't
+        // accidentally join results of two elements that are next to each other in a table
+        return implode('|%|', $output);
     }
 
     /**
@@ -96,6 +128,32 @@ class ElementalPageExtension extends ElementalAreasExtension
             }
             $tags = $html->getContent();
         }
+    }
+
+    private function getEagerLoadedElements(): array
+    {
+        $elements = [];
+        if (is_null(self::$elementalAreas)) {
+            self::$elementalAreas = [];
+            foreach (ElementalArea::get()->eagerLoad('Elements') as $elementalArea) {
+                self::$elementalAreas[$elementalArea->ID] = $elementalArea;
+            }
+        }
+        foreach ($this->owner->hasOne() as $relation => $class) {
+            if (!is_a($class, ElementalArea::class, true)) {
+                continue;
+            }
+            $elementalAreaID = $this->owner->{"{$relation}ID"};
+            if ($elementalAreaID && array_key_exists($elementalAreaID, self::$elementalAreas)) {
+                $elementalArea = self::$elementalAreas[$elementalAreaID];
+            } else {
+                $elementalArea = $this->owner->$relation();
+            }
+            foreach ($elementalArea->Elements() as $element) {
+                $elements[] = $element;
+            }
+        }
+        return $elements;
     }
 
     /**
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -33,6 +33,8 @@ use SilverStripe\View\ArrayData;
 use SilverStripe\View\Parsers\URLSegmentFilter;
 use SilverStripe\View\Requirements;
 use SilverStripe\ORM\CMSPreviewable;
+use SilverStripe\Core\Config\Config;
+use SilverStripe\ORM\DataObjectSchema;
 
 /**
  * Class BaseElement
@@ -65,6 +67,14 @@ class BaseElement extends DataObject implements CMSPreviewable
      * @var string
      */
     private static $description = 'Base element class';
+
+    /**
+     * List of fields to exclude from CMS SiteTree seatch
+     * @see ElementSiteTreeFilterSearch::applyDefaultFilters()
+     */
+    private static array $fields_excluded_from_cms_search = [
+        'ExtraClass',
+    ];
 
     private static $db = [
         'Title' => 'Varchar(255)',
@@ -526,6 +536,59 @@ JS
         // Allow projects to update indexable content of third-party elements.
         $this->extend('updateContentForSearchIndex', $content);
         return $content;
+    }
+
+    /**
+     * Provides content for CMS search if ElementSiteTreeFilterSearch.render_elements is false
+     */
+    public function getContentForCmsSearch(): string
+    {
+        $fieldNames = $this->getTextualDatabaseFieldNames();
+        $excludedFieldNames = $this->getFieldNamesExcludedFromCmsSearch();
+        $contents = [];
+        foreach ($fieldNames as $fieldName) {
+            if (in_array($fieldName, $excludedFieldNames)) {
+                continue;
+            }
+            $contents[] = $this->$fieldName;
+        }
+        // Allow projects to update contents of third-party elements.
+        $this->extend('updateContentForCmsSearch', $contents);
+        
+        // Use |#| to delimit different fields rather than space so that you don't
+        // accidentally join results of two columns that are next to each other in a table
+        $content = implode('|#|', array_filter($contents));
+
+        // Strips tags and be sure there's a space between words.
+        $content = trim(strip_tags(str_replace('<', ' <', $content)));
+
+        return $content;
+    }
+
+    /**
+     * Get field names that have a Varchar or Text like type in the database
+     */
+    private function getTextualDatabaseFieldNames(): array
+    {
+        $fieldNames = [];
+        $textualDatabaseFields = DataObject::getSchema()->databaseFields($this);
+        foreach ($textualDatabaseFields as $fieldName => $databaseFieldType) {
+            $lcType = strtolower(strtok($databaseFieldType ?? '', '('));
+            if (str_contains($lcType, 'varchar') || str_contains($lcType, 'text')) {
+                $fieldNames[] = $fieldName;
+            }
+        }
+        return $fieldNames;
+    }
+
+    private function getFieldNamesExcludedFromCmsSearch(): array
+    {
+        return [
+            // `fixed_fields` contains ['ID', 'ClassName', 'LastEdited', 'Created'] and possibly more
+            ...array_keys(static::config()->get('fixed_fields')),
+            // manually excluded fields
+            ...static::config()->get('fields_excluded_from_cms_search')
+        ];
     }
 
     /**
--- a/tests/BaseElementTest.php
+++ b/tests/BaseElementTest.php
@@ -495,4 +495,12 @@ class BaseElementTest extends FunctionalTest
             $this->assertSame($link, $previewLink);
         }
     }
+
+    public function testGetContentForCmsSearch()
+    {
+        $element = $this->objFromFixture(ElementContent::class, 'content1');
+        $this->assertSame('Test Content', $element->getContentForCmsSearch());
+        $element = $this->objFromFixture(TestElement::class, 'elementDataObject3');
+        $this->assertSame('Hello Test|#|Element 3', $element->getContentForCmsSearch());
+    }
 }
--- a/tests/Controllers/ElementSiteTreeFilterSearchTest.php
+++ b/tests/Controllers/ElementSiteTreeFilterSearchTest.php
@@ -6,6 +6,13 @@ use DNADesign\Elemental\Extensions\ElementalPageExtension;
 use DNADesign\Elemental\Tests\Src\TestPage;
 use SilverStripe\CMS\Controllers\CMSSiteTreeFilter_Search;
 use SilverStripe\Dev\SapphireTest;
+use DNADesign\Elemental\Controllers\ElementSiteTreeFilterSearch;
+use ReflectionMethod;
+use SilverStripe\ORM\DataObject;
+use SilverStripe\Core\Config\Config;
+use SilverStripe\ORM\DataObjectSchema;
+use DNADesign\Elemental\Models\ElementContent;
+use DNADesign\Elemental\Tests\Src\TestElementContentExtension;
 
 class ElementSiteTreeFilterSearchTest extends SapphireTest
 {
@@ -15,6 +22,9 @@ class ElementSiteTreeFilterSearchTest extends SapphireTest
         TestPage::class => [
             ElementalPageExtension::class,
         ],
+        ElementContent::class => [
+            TestElementContentExtension::class,
+        ]
     ];
 
     protected static $extra_dataobjects = [
@@ -50,6 +60,78 @@ class ElementSiteTreeFilterSearchTest extends SapphireTest
                 ['Title' => 'Content blocks page'],
                 ['Title' => 'Regular page'],
             ]],
+        ];
+    }
+
+    /**
+     * @dataProvider provideApplyDefaultFilters
+     */
+    public function testApplyDefaultFilters(bool $renderElements, string $term, array $expected): void
+    {
+        // Set protected method visibility - applyDefaultFilters() is essentially an
+        // extension method that's called by silverstripe/cms, so use of reflection here
+        // is pretty much required
+        $method = new ReflectionMethod(ElementSiteTreeFilterSearch::class, 'applyDefaultFilters');
+        Config::modify()->set(ElementSiteTreeFilterSearch::class, 'render_elements', $renderElements);
+        $filterSearch = new ElementSiteTreeFilterSearch(['Term' => $term]);
+        $ret = $method->invoke($filterSearch, DataObject::get(TestPage::class));
+        $this->assertSame($expected, $ret->column('Title'));
+    }
+
+    public function provideApplyDefaultFilters(): array
+    {
+
+        return [
+            'render_elements true - text search' => [
+                'render_elements' => true,
+                'term' => 'This content is rendered',
+                'expected' => ['Content blocks page']
+            ],
+            'render_elements true - unrendered search' => [
+                'render_elements' => true,
+                'term' => 'This field is unrendered',
+                'expected' => []
+            ],
+            'render_elements true - extended search' => [
+                'render_elements' => true,
+                'term' => 'This content is from an extension hook',
+                'expected' => []
+            ],
+            'render_elements true - int search' => [
+                'render_elements' => true,
+                'term' => '456',
+                'expected' => []
+            ],
+            'render_elements true - enum search' => [
+                'render_elements' => true,
+                'term' => 'Sunny',
+                'expected' => []
+            ],
+            'render_elements false - text search' => [
+                'render_elements' => false,
+                'term' => 'This content is rendered',
+                'expected' => ['Content blocks page']
+            ],
+            'render_elements false - unrendered search' => [
+                'render_elements' => false,
+                'term' => 'This field is unrendered',
+                'expected' => ['Content blocks page']
+            ],
+            'render_elements false - extended search' => [
+                'render_elements' => false,
+                'term' => 'This content is from an extension hook',
+                'expected' => ['Content blocks page']
+            ],
+            'render_elements false - int search' => [
+                'render_elements' => false,
+                'term' => '456',
+                'expected' => []
+            ],
+            'render_elements false - enum search' => [
+                'render_elements' => false,
+                'term' => 'Sunny',
+                'expected' => []
+            ],
         ];
     }
 }
--- a/tests/Controllers/ElementSiteTreeFilterSearchTest.yml
+++ b/tests/Controllers/ElementSiteTreeFilterSearchTest.yml
@@ -17,3 +17,12 @@ DNADesign\Elemental\Models\ElementContent:
     Title: Content
     HTML: Specifically blocks page content
     ParentID: =>DNADesign\Elemental\Models\ElementalArea.blocks_page_area
+    MyInt: 123
+    MyEnum: Cloudy
+  blocks_page_unrendered_content:
+    Title: My title
+    HTML: This content is rendered
+    ParentID: =>DNADesign\Elemental\Models\ElementalArea.blocks_page_area
+    UnrenderedField: This field is unrendered
+    MyInt: 456
+    MyEnum: Sunny
--- a/tests/Src/TestElementContentExtension.php
+++ b/tests/Src/TestElementContentExtension.php
@@ -1,0 +1,20 @@
+<?php
+
+namespace DNADesign\Elemental\Tests\Src;
+
+use SilverStripe\ORM\DataExtension;
+use SilverStripe\Dev\TestOnly;
+
+class TestElementContentExtension extends DataExtension implements TestOnly
+{
+    private static $db = [
+        'UnrenderedField' => 'Varchar(255)',
+        'MyInt' => 'Int',
+        'MyEnum' => 'Enum("Sunny, Cloudy", "Sunny")'
+    ];
+
+    public function updateContentForCmsSearch(array &$contents)
+    {
+        $contents[] = 'This content is from an extension hook';
+    }
+}


michalkleiner:
As long as this had an extension point to be able to change what it indexes then it should be ok, I guess. We rely on shorcodes being processed as those could bring in content that is not in any of the db fields directly and should be indexed for the given element, so for me it should support that, too. The list of excluded fields could be a map to iterate over, or two maps where one are fields and other are trailing strings like ID, Hash etc.. so that we don't create lists of comparisons that would be harder to manage. Could be config, too.

docs/en/searching-blocks.md
```diff
@@ -40,3 +40,24 @@ to make it clear in search results where one piece of content ends and another b
 Page:
   search_index_element_delimiter: ' ... '
 ```
+
+## CMS page search
+
+CMS page search will include search results for pages with elements that match the search query.
+
+By default it uses the same method as the search indexing where it will fully render every element that is
+being search. This is an expensive operation and can cause performance issues if you have a large site with a lot of elements.
```
GuySartorelli:
```suggestion
being searched. This is an expensive operation and can cause performance issues if you have a large site with a lot of elements.
```

docs/en/searching-blocks.md
```diff
@@ -40,3 +40,24 @@ to make it clear in search results where one piece of content ends and another b
...
+
+To increase performance by a large amount, likely more than doubling it, you can disable the rendering of elements and instead just look at the database values of the elements instead.
```
GuySartorelli:
```suggestion
To increase performance by a large amount, likely more than doubling it, you can disable the rendering of elements and instead just look at the database values of the elements directly.
```

docs/en/searching-blocks.md
```diff
@@ -40,3 +40,24 @@ to make it clear in search results where one piece of content ends and another b
...
+
+```yml
+DNADesign\Elemental\Controllers\ElementSiteTreeFilterSearch:
+  render_elements_for_cms_search: false
+```
+
+Searching elements for content in CMS page search can be disabled entirely can be disabled:
```
GuySartorelli:
```suggestion
If the above is still not performant enough, searching elements for content in CMS page search can be disabled entirely can be disabled:
```

src/Extensions/ElementalPageExtension.php
```diff
@@ -98,6 +141,32 @@ public function MetaTags(&$tags)
         }
     }
 
+    private function getEagerLoadedElements(): array
```
GuySartorelli, line 133: It probably makes sense to have `loopThroughElements()` use eager loading as well, since that method is only used by this class and is only used for read-only operations. But only if this method stops getting _all_ elemental areas up front.

src/Extensions/ElementalPageExtension.php
```diff
@@ -98,6 +141,32 @@ public function MetaTags(&$tags)
...
+    {
+        $elements = [];
+        if (is_null(self::$elementalAreas)) {
+            self::$elementalAreas = [];
+            foreach (ElementalArea::get()->eagerLoad('Elements') as $elementalArea) {
+                self::$elementalAreas[$elementalArea->ID] = $elementalArea;
+            }
```
GuySartorelli, line 140: I don't understand why we're getting these _all_ up front - why not get the IDs we need for the relations on `$this->owner` based on the logic below, where we _are_ getting the relation IDs for all elemental areas on the record anyway, and then do a `ElementalArea::get()->filter('ID', $ids)->eagerLoad('Elements')`?

src/Extensions/ElementalPageExtension.php
```diff
@@ -66,6 +89,26 @@ public function getElementsForSearch()
...
+    /**
+     * Returns the contents of all Elements on the pages ElementalAreas for use in CMS search
+     */
+    public function getContentFromElementsForCmsSearch(): string
+    {
+        $output = [];
+        $elements = $this->getEagerLoadedElements();
+        /** @var BaseElement $element */
+        foreach ($elements as $element) {
+            // note: not calling $element->getSearchIndexable() here because it's not public facing
```
GuySartorelli: This should still call `$element->getSearchIndexable()` for consistency - that's respected if we _are_ rendering templates, so it should be respected if we're _not_ rendering them as well (unless you've identified that call as a bottleneck as well - in which case we need to clearly document that it's not respected in this scenario)

src/Models/BaseElement.php
```diff
@@ -528,6 +528,49 @@ public function getContentForSearchIndex(): string
...
     }
 
+    /**
+     * Provides content for CMS search
+     */
+    public function getContentForCmsSearch(): string
+    {
+        $contents = [];
+        foreach ($this->config()->get('db') as $fieldName => $fieldType) {
+            // https://docs.silverstripe.org/en/developer_guides/model/data_types_and_casting/
```
GuySartorelli: It's not really clear what this link is here for.... what on that page am I meant to be looking at, and what does it help to clarify for this code?

src/Models/BaseElement.php
```diff
@@ -528,6 +528,49 @@ public function getContentForSearchIndex(): string
...
+                || $fieldName === 'ShowInMenus'
+                || $fieldName === 'ShowInSearch'
+                || $fieldName === 'Sort'
+                || $fieldName === 'HasBrokenFile'
+                || $fieldName === 'HasBrokenLink'
+                || $fieldName === 'ReportClass'
+                || substr($fieldType, -2) === 'ID'
+                || substr($fieldType, -3) === 'Key'
+                || substr($fieldName, -9) === 'ClassName'
+                || substr($fieldType, -4) === 'Hash'
```
GuySartorelli: This is kinda ugly.... maybe something like this instead?

```php
$ignoredFields = array_merge(
    array_keys($this->config()->get('fixed_fields')), // ID, ClassName, LastEdited, Created are all in here,
    array_keys(InheritedPermissionsExtension::config()->get('db')), // CanViewType and CanEditType - more could be added in the future
    // and then the rest of them hardcoded unless you can find a clean way to get them in little bundles like the above two
```

Though I struggle to think of a time where we've had this level of explicit exclusion of fields? It would feel better to me if we used something like `searchable_fields` config, or some other list of _these are the fields we should use_ that can be user-defined.

src/Models/BaseElement.php
```diff
@@ -528,6 +528,49 @@ public function getContentForSearchIndex(): string
...
+            ) {
+                continue;
+            }
+            // TODO: if $fieldType is HTMLText then run shortcode parser on it?
+            // (possibly not? unlikely to really matter from a cms search perspective?)
```
GuySartorelli: I'd see what gridfield filter search does in this regard so our different searches (which appear the same to the user) are consistent

src/Models/BaseElement.php
```diff
@@ -528,6 +528,49 @@ public function getContentForSearchIndex(): string
...
+        }
+        // Use |#| to delimite different fields rather than space so that you don't
+        // accidentally join results of two columns that are next to each other in a table
+        $content = implode('|#|', $contents);
+
+        // Strips tags and be sure there's a space between words.
+        $content = trim(strip_tags(str_replace('<', ' <', $content)));
+
+        // Allow projects to update content of third-party elements.
+        $this->extend('updateContentForCmsSearch', $content);
```
GuySartorelli: I mentioned this above, but it makes a lot more sense to me for the developer to define which explicit fields should be searched against in the first place. Since we're not rendering the template, it doesn't make any sense to me to give developers a string as the way they modify the element's searchable content.

src/Controllers/ElementSiteTreeFilterSearch.php
```diff
@@ -47,8 +52,13 @@ protected function applyDefaultFilters($query)
                 return false;
             }
 
-            // Check whether the search term exists in the nested page content
-            $pageContent = $siteTree->getElementsForSearch();
+            if ($this->config()->get('render_elements_for_cms_search') === true) {
+                // Check whether the search term exists in the nested page content
+                $pageContent = $siteTree->getElementsForSearch();
+            } else {
+                $pageContent = $siteTree->getContentFromElementsForCmsSearch();
```
GuySartorelli, line 59: Given we're just looking at DB fields directly here... is there a way we can craft this as a db query instead of using `stripos`? I suspect that would be faster. Though if that's too complex to do right now we can always merge this PR and look at that as a potential enhancement down the line.

GuySartorelli (changes requested):
Obviously since this is draft I'm mostly just doing a "is this approach okay" review - but I've highlighted a few things that will need to change eventually in any case.

I think this _has_ to be opt in, since we're changing the way developers even define _what_ is searchable for their blocks in the cms, let alone _how_. The general approach seems sensible - though I wonder if there will be projects that could benefit from this new config that won't just go straight to setting `search_for_term_in_content` to false... but in lieu of a way to validate that ahead of time, I see no reason to not go ahead with this.

src/Extensions/ElementalPageExtension.php
emteknetnz: It's essentially doing another level of eager loading by getting ALL ElementalArea's at once. It needs to be this way to get any sort of benefit out of what we're doing,

On most websites there is only going to be a single ElementalArea per Page. ElementalPageExtension is an extension applied to Page's. So, in this context, $ids is simply going to be [<$this->owner->ID>]. This means we're going to have many queries to fetch a single ElementalArea instead of a single one

But it gets worse ...  `eagerLoad('Elements')` will only load elements where the parent (in this case ElementArea) ID is in the DataList ... which in this case in only a single ElementalArea, so for my test case it'll only fetch 5 Elements.  So there will subsequently be a large number of queries to the Element table too.

src/Models/BaseElement.php
emteknetnz: Was just a reminder to myself, will get deleted later

src/Models/BaseElement.php
emteknetnz: I'm not sure how viable it is to use `$searchable_fields` here since the prior "config" was simply whatever was on the rendered template. 

i.e. the generic out-of-the-box `ElementContent` doesn't have `$searchable_fields` https://github.com/silverstripe/silverstripe-elemental/blob/5/src/Models/ElementContent.php

I guess we could just update `ElementContent` `$searchable_fields` in this PR.  Though it's really going to impact the uptake of this feature.  Lets say a project has 40 existing different content blocks, some of which are in 3rd party modules.  This means they'd need to update all of the `$searchable_fields` on all of the content blocks just to get CMS search performance to be faster ... I feel like most people would just put this into the too hard basket.

Also means that whenever a content block gets a new content fields added to it, the developer would need to remember to also add these to `$searchable_fields` just to retain CMS search functionality.  Right now this isn't really an issue because presumably new content fields would also be added to the elements template

`$searchable_fields` is primarily for GridField search right?  In that scenario it makes intuitive sense for a develop to update `$searchable_fields` as they need it because they want an additional column to be able to search for in a gridfield.  For CMS search it will be such an afterthought it just seems to easy to forgot, so I think if this is configurable it needs to be an opt-out style config, rather than an opt-in like `$searchable_fields` is

src/Models/BaseElement.php
emteknetnz: That could well make sense, though if we use  a string it allows projects to do useful things like this:

```php
    protected function updateContentForCmsSearch(string &$content)
    {
        $content .= 'ヽ(⌐■_■)ノ♪♬';
    }
```

src/Models/BaseElement.php
GuySartorelli: > Unfortunately we cannot use `$searchable_fields` here since the prior "config" was simply whatever was on the rendered template

I don't see how that's a blocker, since this configuration is to explicitly _not_ do that?

> generic out-of-the-box `ElementContent` doesn't have `$searchable_fields`

We can add that, you know. IMO we should just add it to all supported modules which define their own blocks, and from there its up to the developers enabling this opt-in feature to ensure the blocks they're using have the appropriate config.

src/Extensions/ElementalPageExtension.php
GuySartorelli: Nevermind, this isn't a sensible suggestion as per https://github.com/silverstripe/silverstripe-elemental/pull/1069#discussion_r1244616454

src/Controllers/ElementSiteTreeFilterSearch.php
emteknetnz: Yeah ...... technically.  Though it would involve working out what all of the fields to query are ahead of time, and then putting all of those into a big `$filter` and putting `:PartialMatch` on all the things. PartialMatch translates to SQL `WHERE <field> LIKE '%<searchterm>%'`

Feels much more like a future ENH to me rather than a do now thing. It may be faster and should def use less PHP + memory, though more database

src/Models/BaseElement.php
emteknetnz: > I don't see how that's a blocker

My point is that fields being rendered in a template currently acts as a quasi config of sorts.  i.e. If you want a field to appear in search, then add it to the template.  No need to add it to proper config.

> We can add that, you know. IMO we should just add it to all supported modules which define their own blocks

Yeah this bits easy.  However we officially support only a few blocks.  

Vast, vast bulk of content blocks out in the wild are defined at the project level, usually as the result of a designer doing UX design for a new website build.   I feel like forcing devs to define to this config which is literally only for CMS page search is adding too much friction and will lead to a worse CMS page search experience since it'll be so easy to forget to do / people won't even know they're supposed to do it - for instance it's pretty normal for junior devs to be tasked with creating new content blocks on a project.

> up to the developers enabling this opt-in feature

I can agee with this being opt-in for CMS 5 at least, though ideally I'd like this to be opt-out for CMS 6 (or simply just get rid of template rendering for cms search as an option) because the performance boost is just so high, and CMS page search is probably the single worst performing part of the CMS

src/Models/BaseElement.php
michalkleiner: I think we need the extension points because we need a way how to remove things as well, not just how to add them.

GuySartorelli: Is removing the content when it's already in this string okay though? If I was developing with this I'd much rather have control over which _fields_ are included, knowing that fields A, B, and C are relevant but fields D and E aren't.

I guess it doesn't hurt to have both ways of controlling it, but this _does_ make it so we can't later optimise this option further by having it do the filter in the DB directly as discussed in https://github.com/silverstripe/silverstripe-elemental/pull/1069#discussion_r1244608200 (at least not without waiting for a major release)

michalkleiner: Yeah, my gut feel is to have the extension and allow modifications on the string. Could we want e.g. different logic based on subsites? Or other stuff where a single config might not be flexible enough?

docs/en/searching-blocks.md
emteknetnz: UPdated

docs/en/searching-blocks.md
emteknetnz: Updated

docs/en/searching-blocks.md
emteknetnz: Updated

src/Extensions/ElementalPageExtension.php
emteknetnz: Added in call to `$element->getSearchIndexable()`

src/Models/BaseElement.php
emteknetnz: I might be wrong, though I couldn't find any reference to shortcode parsing for gridfield search, which I believe uses [SearchContext](https://github.com/silverstripe/silverstripe-framework/blob/5/src/ORM/Search/SearchContext.php)

To double check I did try putting a breakpoint in EmbedShortcodeProvider::handle_shortcode(), made a SiteTreeAdmin and tried doing a search. It didn't trigger the breakpoint, so seems like shortcodes in search aren't a thing.

GuySartorelli: Sweet, in that case probably just remove the todo and call it done, IMO

src/Models/BaseElement.php
emteknetnz: > I'd much rather have control over which fields are included

As I've mentioned previously, there already is `quasi-config` in that the "config of fields that are included" is currently "whatever fields are included in the template".  Obviously we can't use that here.

I'm pretty against the idea of opt-in config for the fields you want to include. I want to actually get adoption on this and I view "having to define every single field on every single element" as a pretty big annoyance.  Also it'll degrade over time because any time a new developer adds a new content block (very standard for junior devs) then it's super easy to forget / not know that you need to add the new field to the CMS SiteTree search thing as well.

I think if we're going to having configurable fields, best we can do here is create a configurable list of fields to NOT include

emteknetnz: I've added config array `BaseElement.exclude_fields_from_cms_search` to allow for opt-ing out fields

src/Models/BaseElement.php

docs/en/searching-blocks.md
```diff
@@ -40,3 +40,33 @@ to make it clear in search results where one piece of content ends and another b
...
+being searched. This is an expensive operation and can cause performance issues if you have a large site with a lot of elements.
+
+To increase performance by a large amount, likely more than doubling it, you can disable the rendering of elements and instead just look at the database values of the elements directly.
+
+```yml
+DNADesign\Elemental\Controllers\ElementSiteTreeFilterSearch:
+  render_elements: false
+```
+
+If `render_elements` is to `false`, then all fields that are not generic fields such as keys or something like `CanViewType` are searched. Individiual fields on blocks can be excluded from the search by adding fields to the `exclude_fields_from_cms_search` array config variable on the element class. e.g.
```
GuySartorelli:
```suggestion
If `render_elements` is `false`, then all fields that are not generic fields such as keys or something like `CanViewType` are searched. Individual fields on blocks can be excluded from the search by adding fields to the `exclude_fields_from_cms_search` array config variable on the element class. e.g.
```

GuySartorelli: > "all fields that are not generic fields such as keys or something like `CanViewType`"

It's not really clear what is meant by "generic fields" here...
If we add a new `getFieldsExcludedFromCMSSearch()` or similarly named method (which I strongly think we should as mentioned in another comment), we could just say here "all fields that are not returned from the `getFieldsExcludedFromCMSSearch()` method".
If we ignore or only include fields of certain types as well, then we will need to mention that here too.

src/Controllers/ElementSiteTreeFilterSearch.php
```diff
@@ -20,6 +20,11 @@ class ElementSiteTreeFilterSearch extends CMSSiteTreeFilter_Search
      */
     private static $search_for_term_in_content = true;
 
+    /**
+     * @todo Change to false for CMS6?
```
GuySartorelli: Please remove this comment - if you think this is an appropriate course of action create a new issue for it.

src/Controllers/ElementSiteTreeFilterSearch.php
```diff
@@ -20,6 +20,11 @@ class ElementSiteTreeFilterSearch extends CMSSiteTreeFilter_Search
...
+     */
+    private static bool $render_elements = false;
```
GuySartorelli: Please add an appropriate description of this configuration property in a PHPDoc comment. We don't do this enough - but it's useful for anyone looking at the API docs site.

src/Models/BaseElement.php
GuySartorelli: Lets generate this list in a cleaner way then, in a new protected method, e.g:
```php
protected function getFieldsExcludedFromCMSSearch(): array
{
    return [
        ...array_keys(static::config()->get('fixed_fields')), // ID, ClassName, LastEdited, Created are all in here,
        ...array_keys(InheritedPermissionsExtension::config()->get('db')), // CanViewType and CanEditType - more could be added in the future
        ...array_keys(Versioned::config()->get('db')), // The "Versioned" field name is here but more could be added in the future
        'ShowInMenus', // though for this and `ShowInSearch` it could be better to instead skip any Boolean fields
        'ShowInSearch',
        'Sort', // though it may be better to just skip all Int fields
        //...etc any other fields we want to explicitly exclude
        ...static::config()->get('exclude_fields_from_cms_search')
    ];
}
```
What's more, we should only be adding the fields from `Versioned` and `InheritedPermissionsExtension` if those extensions are applied, so we need a condition on `static::has_extension()` for those two.

I don't think we should include any of the substr ones there. I'm guessing "ends with ID" and "ends with ClassName" were for has_one, but those aren't in the db config array itself so we don't need to account for those. For the others, if someone has a key or hash field they can exclude it themselves. It seems unlikely that a hash field will match any search string anyway and the one extra field isn't likely to slow things by much. And "*Key" could be anything, including something that is relevant for search.

As mentioned in comments in the example above, it's probably a good idea for `getContentForCmsSearch()` to just outright ignore `DBBoolean`, `DBInt`, and a few other field types (and their subclasses) - or, perhaps better yet, to only look at fields which are `DBString` or some subclass of it.

src/Models/BaseElement.php
GuySartorelli: I've decided I'm okay with searching across all db fields with configurable exceptions - and it doesn't hurt to have this extension point, so that's fine too.

But I _would_ ideally also like a way to add extra fields which are part of relations, since it is _very_ common to have a block which gets its content almost exclusively from relations (e.g. [carousels](https://github.com/dnadesign/silverstripe-elemental-carousel), [nested blocks](https://github.com/dnadesign/silverstripe-elemental-list), [decision trees](https://github.com/dnadesign/silverstripe-elemental-decisiontree), [accordions](https://github.com/dynamic/silverstripe-elemental-accordion), etc etc).

@michalkleiner I'd like your input on that side as well since you've expressed interest in the configuration aspects of this PR - would you find it useful to have the ability to add relation fields into the list of searchable fields?

GuySartorelli (changes requested):
I'll give this a more thorough review once we've resolved the current conversations about the approach - but it sounds like we're getting to a consensus now.

src/Models/BaseElement.php
michalkleiner: Yes, with the "index the rendered template" approach that content is included, so the change here shouldn't make the content scope smaller, or should provide an alternative way how to achieve the same. Alt tags for linked images, nested carousel content etc.

The extension point is good-enough affordance for that, so if we go with "all content db fields less configured exclusion fields plus content provided/removed via the extension point", then that'd work for us.

If we could support some form of wildcard mechanism for the exclusions that'd be dope, e.g. `Video*` would exclude all fields starting with 'Video' (`VideoLink`, `VideoTitle`...), '*' would exclude all and let the extension handle what's needed specifically on project level.

src/Models/BaseElement.php
emteknetnz: Have updated to only include Text or Varchar like fields, this greatly limits the amount we need to filter out. We don't  need the InheritedPermissionsExtension or Versioned ones since they only include Enum + Int

docs/en/searching-blocks.md
emteknetnz: Updated

emteknetnz: Changed to `then all fields that have stored as a Varchar or Text like are searched`

src/Controllers/ElementSiteTreeFilterSearch.php
emteknetnz: The question mark is really - should we change this default for CMS6. IMO we should - I mentioned in [earlier](https://github.com/silverstripe/silverstripe-elemental/pull/1069#issue-1776324663) as option B which would be my 2nd choice of Option C which we've deicded against

We don't have a proper mechanism for saying "Do this thing for CMS 6" - part of the issue is we don't have CMS 6 branches (and ideally we won't for a fair while due to their additional overhead).  Simply creating a card at this point isn't good IMO, it'll just get lost in our oversized backlog.

So ... a @todo with a cms6 is probably the best I can think of for now?

src/Controllers/ElementSiteTreeFilterSearch.php
emteknetnz: Updated

src/Models/BaseElement.php
emteknetnz: `Video*` seems almost trivial to do without a wildcard - it's easy to exclude fields on a per element class level with the config in this PR:

```yml
App\MyElement:
  fields_excluded_from_cms_search:
    - MyFieldToExclude
    - AnotherFieldToExclude
```

I think for simplicity sake it better to NOT include a wildcard mechanism, I don't think it's consistent with any other config in Silverstripe

Re adding fields from related classes, that's easy enough to do using the new extension point:

```php
<?php

// extension is applied to any Element class in the project
class MyElementExtension extends DataExtension
{
    protected function updateContentForCmsSearch(string &$content): void
    {
        $content .= $this->owner->MyRelation()->MyField;
    }
}

```

GuySartorelli: > Re adding fields from related classes, that's easy enough to do using the new extension point

It is a bit awkward to have to update a string when you just want to add a field - and in your example the delimiter is missed out which means your newly added field will be appended directly to the end of the previous field which I think is going to be a common mistake. That said, this PR has had about enough time in the oven and this would be an easy enhancement to make in the future if people want it, so I'm okay leaving the extensibility as is for now.

src/Controllers/ElementSiteTreeFilterSearch.php
GuySartorelli: An issue is better than a todo every time. It's not uncommon for us to have an issue that's open for discussion (RFC for example). We can't easily add discussion to a todo comment, and frankly we're not likely to look at it when we do CMS 6 stuff so we'll be seeing this comment well into CMS 8 in all likelyhood. But an issue with "CMS 6: " in the title has a high chance of being looked at as we're likely to search for issues that mention CMS 6 in them when we start looking at what to include. Please remove the todo.

src/Models/BaseElement.php
```diff
@@ -528,6 +536,85 @@ public function getContentForSearchIndex(): string
...
+        return $textualDatabaseFields;
+    }
+
+    private function getFieldsExcludedFromCmsSearch(): array
+    {
+        return [
+            // `fixed_fields` contains ['ID', 'ClassName', 'LastEdited', 'Created'] and possibly more
+            ...array_keys(static::config()->get('fixed_fields')),
+            // generic field names that are not useful for search
+            'ExtraClass',
```
```suggestion
```
Put this in the config instead - it's declared in this class so that's the place that makes most sense.

src/Models/BaseElement.php
```diff
@@ -66,6 +68,12 @@ class BaseElement extends DataObject implements CMSPreviewable
      */
     private static $description = 'Base element class';
 
+    /**
+     * List of fields to exclude from CMS SiteTree seatch
+     * @see ElementSiteTreeFilterSearch::applyDefaultFilters()
+     */
+    private static array $fields_excluded_from_cms_search = [];
```
```suggestion
    private static array $fields_excluded_from_cms_search = [
        'ExtraClass',
    ];
```
as per https://github.com/silverstripe/silverstripe-elemental/pull/1069#discussion_r1251298926

src/Models/BaseElement.php
```diff
@@ -528,6 +536,85 @@ public function getContentForSearchIndex(): string
...
+        // Allow projects to update content of third-party elements.
+        $content .= $delimiter;
+        $this->extend('updateContentForCmsSearch', $content);
+
+        // Remove delimiters that aren't required
+        $content = $this->removeExtraDelimiters($content, $delimiter);
+
+        return $content;
+    }
+    /**
```
```suggestion
    }

    /**
```

src/Models/BaseElement.php
```diff
@@ -528,6 +536,85 @@ public function getContentForSearchIndex(): string
...
+     * Get fields that have a Varchar or Text like type in the database
+     */
+    private function getTextualDatabaseFields(): array
+    {
+        $textualDatabaseFields = singleton(DataObjectSchema::class)->databaseFields($this);
```
```suggestion
        $textualDatabaseFields = DataObject::getSchema()->databaseFields($this);
```
This is the standard way to get that singleton

src/Models/BaseElement.php
```diff
@@ -528,6 +536,85 @@ public function getContentForSearchIndex(): string
...
+        foreach ($textualDatabaseFields as $fieldName => $databaseFieldType) {
+            $lcType = strtolower($databaseFieldType);
+            if (!str_contains($lcType, 'varchar') && !str_contains($lcType, 'text')) {
+                unset($textualDatabaseFields[$fieldName]);
+            }
+        }
+        return $textualDatabaseFields;
```
GuySartorelli: An alternative to this is to explicitly check if the field is a `DBString` or subclass of it instead of a naive check against the name of the field spec, e.g (with variable names that make more sense to me):
```suggestion
        $dbFields = DataObject::getSchema()->databaseFields($this);
        foreach ($dbFields as $fieldName => $fieldSpec) {
            $fieldClass = strtok($fieldSpec ?? '', '(');
            if ($fieldClass !== 'ForeignKey' && $fieldClass !== 'PrimaryKey' && singleton($fieldClass) instanceof DBString) {
                unset($dbFields[$fieldName]);
            }
        }
        return $dbFields;
```

Note that this logic is copied from `DataObjectSchema::cacheDatabaseFields()` - so if you're worried about having so many singleton calls there's definitely room to extract that out the db field class in DataObjectSchema as a separately cached datapoint. This could also be useful for projects and other areas of core code where we're interested in specifically the _type_ of a field - though understandable if that's considered out of scope for now.

But even if you don't do any of the above (which is optional - you end up with very similar, though not identical results e.g. the above includes enum), you should make this change:
```suggestion
        $textualDatabaseFields = singleton(DataObjectSchema::class)->databaseFields($this);
        foreach ($textualDatabaseFields as $fieldName => $databaseFieldType) {
            $lcType = strtolower(strtok($databaseFieldType ?? '', '('););
            if (!str_contains($lcType, 'varchar') && !str_contains($lcType, 'text')) {
                unset($textualDatabaseFields[$fieldName]);
            }
        }
        return $textualDatabaseFields;
```
Otherwise you might end up including fields you don't expect if "varchar" or "text" appear in the arguments of a field type that would otherwise not be included.

src/Models/BaseElement.php
```diff
@@ -528,6 +536,85 @@ public function getContentForSearchIndex(): string
...
+
+        // Strips tags and be sure there's a space between words.
+        $content = trim(strip_tags(str_replace('<', ' <', $content)));
+
+        // Allow projects to update content of third-party elements.
+        $content .= $delimiter;
+        $this->extend('updateContentForCmsSearch', $content);
+
+        // Remove delimiters that aren't required
+        $content = $this->removeExtraDelimiters($content, $delimiter);
```
GuySartorelli: This step seems like unnecessary additional complexity. Does this improve the accuracy or performance of the search in a meaningful way? If not I propose we just remove it.

src/Models/BaseElement.php
```diff
@@ -528,6 +536,85 @@ public function getContentForSearchIndex(): string
         return $content;
     }
 
+    /**
+     * Provides content for CMS search
```
```suggestion
     * Provides content for CMS search if ElementSiteTreeFilterSearch.render_elements is false
```

src/Models/BaseElement.php
```diff
@@ -528,6 +536,85 @@ public function getContentForSearchIndex(): string
...
+    public function getContentForCmsSearch(): string
+    {
+        $textualDatabaseFields = $this->getTextualDatabaseFields();
+        $excludedFields = $this->getFieldsExcludedFromCmsSearch();
+        $contents = [];
+        foreach ($this->config()->get('db') as $fieldName => $fieldType) {
+            if (!in_array($fieldType, $textualDatabaseFields) || in_array($fieldName, $excludedFields)) {
+                continue;
+            }
+            $contents[] = $this->$fieldName;
```
```suggestion
        $content = [];
        foreach ($this->config()->get('db') as $fieldName => $fieldType) {
            if (!in_array($fieldType, $textualDatabaseFields) || in_array($fieldName, $excludedFields)) {
                continue;
            }
            $content[] = $this->$fieldName;
```
This way we reuse the same variable below instead of still holding on to this array of the same content, which we don't need beyond that point.

src/Models/BaseElement.php
```diff
@@ -528,6 +536,85 @@ public function getContentForSearchIndex(): string
...
+        // accidentally join results of two columns that are next to each other in a table
+        $delimiter = '|#|';
+        $content = implode($delimiter, $contents);
+
+        // Strips tags and be sure there's a space between words.
+        $content = trim(strip_tags(str_replace('<', ' <', $content)));
+
+        // Allow projects to update content of third-party elements.
+        $content .= $delimiter;
+        $this->extend('updateContentForCmsSearch', $content);
```
```suggestion
        $content = trim(strip_tags(str_replace('<', ' <', $content))) . $delimiter;

        // Allow projects to update content of third-party elements.
        $this->extend('updateContentForCmsSearch', $content);
```
This just looks a little tidier

src/Models/BaseElement.php
```diff
@@ -528,6 +536,85 @@ public function getContentForSearchIndex(): string
...
+     */
+    public function getContentForCmsSearch(): string
+    {
+        $textualDatabaseFields = $this->getTextualDatabaseFields();
+        $excludedFields = $this->getFieldsExcludedFromCmsSearch();
+        $contents = [];
+        foreach ($this->config()->get('db') as $fieldName => $fieldType) {
+            if (!in_array($fieldType, $textualDatabaseFields) || in_array($fieldName, $excludedFields)) {
+                continue;
+            }
```
```suggestion
        $textualDatabaseFields = $this->getTextualDatabaseFields();
        $excludedFields = $this->getFieldsExcludedFromCmsSearch();
        $contents = [];
        foreach ($textualDatabaseFields as $fieldName) {
            if (in_array($fieldName, $excludedFields)) {
                continue;
            }
```
We already get the relevant field names in `getTextualDatabaseFields()` so there's no need to go through the db config a second time, unless I'm missing something?

Note depending on https://github.com/silverstripe/silverstripe-elemental/pull/1069#discussion_r1251324764 this may need an `array_keys()` call

src/Models/BaseElement.php
GuySartorelli: Might be better to rearrange so that the resultant array is just an array of field _names_ - that makes it so we can directly loop through them, since this should be all of the relevant fields we want to try include (we could do `array_keys()` but that's O(n) so not as ideal)

tests/BaseElementTest.php
GuySartorelli, line 1: Please add a test for `getTextualDatabaseFields()`

GuySartorelli (changes requested):

src/Models/BaseElement.php
emteknetnz: I do not like variable reuse, especially since $contents is an array while $content is a string

src/Models/BaseElement.php
emteknetnz: I think that's less clear.

The idea of adding the $delimiter on its own line is to make it clearer that it's for the benefit of the extension hook

GuySartorelli: That wasn't at all clear to me - the comment will need to reflect that if that's the intention.

src/Models/BaseElement.php
GuySartorelli: Not a big deal either way so fine as is

src/Models/BaseElement.php
emteknetnz: I don't want to use `instanceof DBString` because that includes Enum's which are really just fancy integers rather than unstructured text content.  Projects could also do something weird with a subclass of DBString, so seems sensible to me to just check the database type

Makes sense to remove the (...) bit though, have updated

src/Controllers/ElementSiteTreeFilterSearch.php
emteknetnz: Made issue https://github.com/silverstripe/silverstripe-elemental/issues/1072 and removed comment

src/Models/BaseElement.php
emteknetnz: Updated

src/Models/BaseElement.php
emteknetnz: Seems correct, updated

src/Models/BaseElement.php

src/Models/BaseElement.php
emteknetnz: Updated

src/Models/BaseElement.php
emteknetnz: Updated

src/Models/BaseElement.php
emteknetnz: Updated

src/Models/BaseElement.php
emteknetnz: Updated

tests/BaseElementTest.php
emteknetnz: It's a private so ideally you shouldn't test, instead you should be aiming to test public API - there's lots of references to "don't test private methods" on the interweb e.g. https://softwareengineering.stackexchange.com/a/380296.

This method is partially covered by the test `ElementSiteTreeFilterSearchTest::testApplyDeafultFilters()` below since it is a code path as part of that test

src/Models/BaseElement.php
emteknetnz: Updated

tests/BaseElementTest.php
GuySartorelli: I think it'll be pretty ugly trying to test it properly through public API but if that's how you want to do it, please more completely test the functionality of `getTextualDatabaseFields()` through the public methods.

src/Models/BaseElement.php
michalkleiner: Also projects can still munch the content together from several fields without the delimiter, should we pass it in as a second param and add a comment (for folks who search where the hook is used) to mention it's recommended to delimit separate pieces of content with the provided delimiter?

GuySartorelli: That sounds like a sensible idea - it'll make it much less likely for people to forget (or not even know about) delimiting content.

emteknetnz: Actually given that probably makes sense to pass in $contents array rather than $contents string to the extension hook and then delimit so that projects don't need to worry about it

Have updated

tests/BaseElementTest.php
emteknetnz: Have added tests for int + enum that do not return search results. We're already testing for positive matches on a varchar and htmltext field so I think we're good enough

docs/en/searching-blocks.md
```diff
@@ -40,3 +40,33 @@ to make it clear in search results where one piece of content ends and another b
...
+If `render_elements` is `false`, then all fields that have stored as a Varchar or Text like are searched. Individual fields on blocks can be excluded from the search by adding fields to the `exclude_fields_from_cms_search` array config variable on the element class. e.g.
+
+```yml
+App\MyElement:
+  fields_excluded_from_cms_search:
+    - MyFieldToExclude
+    - AnotherFieldToExclude
+```
+
+If the above is still not performant enough, searching elements for content in CMS page search can be disabled entirely can be disabled:
```
michalkleiner:
```suggestion
If the above is still not performant enough, searching elements for content in CMS page search can be disabled entirely:
```

src/Models/BaseElement.php
```diff
@@ -528,6 +538,59 @@ public function getContentForSearchIndex(): string
...
+            $contents[] = $this->$fieldName;
+        }
+        // Allow projects to update contents of third-party elements.
+        $this->extend('updateContentForCmsSearch', $contents);
+        
+        // Use |#| to delimit different fields rather than space so that you don't
+        // accidentally join results of two columns that are next to each other in a table
+        $content = implode('|#|', array_filter($contents));
```
michalkleiner, line 560: Should we use `|%|` or `|#|` consistently everywhere? And is this already a good enough example that it would make sense to have it either as a config or at least as a class constant so it's consistent across usages?

docs/en/searching-blocks.md
emteknetnz: Updated

src/Models/BaseElement.php
emteknetnz: It really makes no difference, I just used different delimeter in different places so that it's easier to debug where things are coming from

GuySartorelli: Yeah I don't think it matters so much _what_ the delimiter is - all that matters is that there _is_ a delimiter, and that the delimiter isn't something likely to be included as a search term.

GuySartorelli (approved):
Looks good, works well locally.

There's a subtle side-effect where when rendering templates, if the title of the block _isn't_ displayed (which is the default for most blocks), it won't be included in search - but with the new feature, it _will_, which is actually a slight improvement to search fidelity 😅

PR #1070, opened 2023-06-29
MNT ESLint issues

sabina-talipova:
## Description
Turn on some of the ESLint.
## Parent issue
-  https://github.com/silverstripe/silverstripe-admin/issues/1433


client/src/components/ElementEditor/AddNewButton.js
```diff
@@ -17,9 +17,9 @@ class AddNewButton extends Component {
   }
 
   toggle() {
-    this.setState({
-      popoverOpen: !this.state.popoverOpen
-    });
+    this.setState((prevState) => ({
```
emteknetnz, line 20: Do we need to make this change to get eslint to pass? I'd prefer not to make changes like this in a linting PR

client/src/components/ElementEditor/Element.js
```diff
@@ -187,9 +187,9 @@ class Element extends Component {
     }
 
     if (type.inlineEditable && !loadingError) {
-      this.setState({
-        previewExpanded: !this.state.previewExpanded
-      });
+      this.setState((prevState) => ({
```
emteknetnz, line 190: As above

client/src/components/ElementEditor/Header.js
```diff
@@ -79,11 +78,10 @@ class Header extends Component {
     );
   }
 
-
   toggle() {
-    this.setState({
-      tooltipOpen: !this.state.tooltipOpen
-    });
+    this.setState((prevState) => ({
```
emteknetnz, line 82: As above

client/src/components/ElementEditor/HoverBar.js
```diff
@@ -63,9 +63,9 @@ class HoverBar extends Component {
   }
 
   toggle() {
-    this.setState({
-      popoverOpen: !this.state.popoverOpen
-    });
+    this.setState((prevState) => ({
```
emteknetnz, line 66: As above

emteknetnz (changes requested):

client/src/components/ElementEditor/HoverBar.js
sabina-talipova: This is the rule [react/no-access-state-in-setstate](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-access-state-in-setstate.md).
```txt
Usage of this.state inside setState calls might result
in errors when two state calls are called in batch and thus referencing 
old state and not the current state.
``` 
This problem describes [there](https://legacy.reactjs.org/docs/react-component.html#setstate)
```txt
setState() does not always immediately update the component.
It may batch or defer the update until later. This makes reading this.state right after
calling setState() a potential pitfall. Instead, use componentDidUpdate or a setState callback
(setState(updater, callback)), either of which are guaranteed to fire after the update has been applied. 
If you need to set the state based on the previous state, read about the updater argument below.
```

client/src/components/ElementEditor/Header.js
sabina-talipova: See [this comment](https://github.com/silverstripe/silverstripe-elemental/pull/1070#discussion_r1253833874)

client/src/components/ElementEditor/AddNewButton.js
sabina-talipova: See [this comment](https://github.com/silverstripe/silverstripe-elemental/pull/1070#discussion_r1253833874)

client/src/components/ElementEditor/Element.js
sabina-talipova: See [this comment](https://github.com/silverstripe/silverstripe-elemental/pull/1070#discussion_r1253833874)

emteknetnz (changes requested):
Please change the commit and PR prefix from FIX to MNT

emteknetnz (approved):

PR #1073, opened 2023-07-04
MNT Fix broken unit test when subsites is installed

emteknetnz:
Issue https://github.com/silverstripe/recipe-kitchen-sink/issues/44

Regression started happening because of this PR https://github.com/silverstripe/silverstripe-subsites/pull/527

### DIFF
--- a/tests/BaseElementTest.php
+++ b/tests/BaseElementTest.php
@@ -482,9 +482,17 @@ class BaseElementTest extends FunctionalTest
         /** @var BaseElement $element */
         $element = $this->objFromFixture($class, $elementIdentifier);
         $previewLink = $element->PreviewLink();
+        // Remove SubsiteID from preview link in case subsites is installed
+        $rx = '/[?|&]SubsiteID=[0-9]+/';
+        if (preg_match($rx, $previewLink)) {
+            $previewLink = preg_replace($rx, '', $previewLink);
+        }
 
         if ($link) {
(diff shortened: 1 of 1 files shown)


tests/BaseElementTest.php
```diff
@@ -482,17 +482,31 @@ public function testPreviewLink(string $class, string $elementIdentifier, ?strin
...
             $this->assertSame(1, substr_count($previewLink, '#'));
         } else {
-            $this->assertSame($link, $previewLink);
+            if ('/' === $previewLink) {
+                // $previewLink will be suffixed with forward slash when subsites is installed
+                // See comment higher up in this method
+                $this->assertSame('/', $previewLink);
+            } else {
+                $this->assertSame($link, $previewLink);
+            }
```
GuySartorelli: You're effectively doing `$this->assertSame($previewLink, $previewLink);`.... let's just strip off the suffix instead.

```suggestion
            $this->assertEmpty(rtrim($previewLink, '/'));
```

It'll either be an empty string or a null - which for preview link are equivalent anyway. And in this assertion `$link` is always going to be null according to the provided values, or at the very least falsy if we change the provided values. So this should be safe.

GuySartorelli (changes requested):

tests/BaseElementTest.php
emteknetnz: Updated

GuySartorelli (approved):

PR #1074, opened 2023-07-06
MNT Handle null variable

emteknetnz:
Issue https://github.com/silverstripe/recipe-kitchen-sink/issues/44

### DIFF
--- a/tests/BaseElementTest.php
+++ b/tests/BaseElementTest.php
@@ -481,7 +481,7 @@ class BaseElementTest extends FunctionalTest
     {
         /** @var BaseElement $element */
         $element = $this->objFromFixture($class, $elementIdentifier);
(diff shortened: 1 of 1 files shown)


tests/BaseElementTest.php
```diff
@@ -484,7 +484,7 @@ public function testPreviewLink(string $class, string $elementIdentifier, ?strin
         $previewLink = $element->PreviewLink();
```
michalkleiner: I'd do 
`$previewLink = $element->PreviewLink() ?? '';`
and then we don't need to worry about it for the whole method.

emteknetnz: Updated

michalkleiner (approved):
MOG

PR #1082, opened 2023-08-17
fix: getPage returning the wrong page

wilr:
If you programaticaly update the ParentID of an element, getPage() still returns the outdated page.

### DIFF
--- a/tests/BaseElementTest.php
+++ b/tests/BaseElementTest.php
@@ -511,4 +511,17 @@ class BaseElementTest extends FunctionalTest
         $element = $this->objFromFixture(TestElement::class, 'elementDataObject3');
         $this->assertSame('Hello Test|#|Element 3', $element->getContentForCmsSearch());
     }
+
+
+    public function testGetPage()
+    {
+        $element = $this->objFromFixture(ElementContent::class, 'content1');
+
+        $this->assertStringContainsString($element->getPage()->Title, 'Test Elemental');
+
+        $newArea = $this->objFromFixture(ElementalArea::class, 'area52');
+        $element->ParentID = $newArea->ID;
+
+        $this->assertStringContainsString($element->getPage()->Title, 'Page with one elements');
+    }
 }
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -692,21 +692,28 @@ JS
 
     /**
      * Despite the name of the method, getPage can return any type of DataObject
+     *
      * @return null|DataObject
      * @throws \Psr\Container\NotFoundExceptionInterface
      * @throws \SilverStripe\ORM\ValidationException
      */
     public function getPage()
     {
-        // Allow for repeated calls to be cached
         if (isset($this->cacheData['page'])) {
-            return $this->cacheData['page'];
+            if (isset($this->cacheData['parent_id']) && $this->cacheData['parent_id'] === $this->ParentID) {
+                return $this->cacheData['page'];
+            }
         }
 
         $class = DataObject::getSchema()->hasOneComponent($this, 'Parent');
         $area = ($this->ParentID) ? DataObject::get_by_id($class, $this->ParentID) : null;
+
         if ($area instanceof ElementalArea && $area->exists()) {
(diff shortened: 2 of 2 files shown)


GuySartorelli (changes requested):
Changing the key used in `cacheData` could break customisations in subclasses which fetch or set data into `cacheData`, since it's not a private property.

Is there a sensible way to bust the cache instead? e.g. in `onBeforeWrite()`, or - if you want to call `getPage()` before writing for some reason, check if `$this->isChanged('PageID')` and don't use the cached data if it has been changed?

This also seems like a good candidate for a unit test to avoid regressions.

wilr:
@GuySartorelli Updated to use  isChanged() now and added a unit test.

GuySartorelli (approved):
Ideally the unit test would also test the caching functionality - but given that functionality was already there it wouldn't be reasonable to ask for this PR to introduce that test coverage. I'm okay with this as is.

Looks good, will merge when CI goes green. Thanks for submitting this.

tests/BaseElementTest.php
```diff
@@ -511,4 +511,18 @@ public function testGetContentForCmsSearch()
...
+    {
+        $element = $this->objFromFixture(ElementContent::class, 'content1');
+
+        $this->assertStringContainsString($element->getPage()->Title(), 'Test Elemental');
+
+        $updatePage = $this->objFromFixture(TestPage::class, 'page_with_elements');
+
+        $element->ParentID = $updatePage->ID;
+
+        $this->assertStringContainsString($element->getPage()->Title(), 'Page with elements');
```
GuySartorelli:
```suggestion
        $this->assertStringContainsString($element->getPage()->Title, 'Test Elemental');

        $updatePage = $this->objFromFixture(TestPage::class, 'page_with_elements');

        $element->ParentID = $updatePage->ID;

        $this->assertStringContainsString($element->getPage()->Title, 'Page with elements');
```

GuySartorelli (changes requested):
CI is unhappy - please fix the test.

GuySartorelli:
CI is still unhappy here. Can you please resolve that?

GuySartorelli (commented):
There are, again, a _lot_ of unrelated changes in this PR related to formatting.
I am going to ask you, again, that in future pull requests you either don't commit those changes, or you include them in a _separate_ commit. I'm going to start asking you to make those changes before merging PRs if you continue to ignore that request. You can leave it as is for this one though.

> Updated to use isChanged() now and added a unit test.

This doesn't seem to be the case... did you change it again since then? If so, can you please explain why? Using `isChanged()` seems like the perfect solution here instead of adding more info into cacheData.

wilr:
Hi @GuySartorelli isChanged() doesn't work as it was cleared when calling onBeforeWrite() so I reverted back (but to a new value to keep your original comment in mind). What we were seeing

```
echo $element->getPage()->Title // 'Contact us';
$element->ParentID = 1;
$element->write();
echo $element->getPage()->Title // 'Contact us';
```

As $element->isChanged('ParentID') == false. We could clear the cache data onBeforeWrite but that seemed like leaking things outside of that method. 

Sorry! Editor being all auto-magic fix. I'll turn it off

GuySartorelli:
Thanks for that context. In that case, I think this is good as is. Once CI has gone green on this latest push I'll approve and merge.

> Sorry! Editor being all auto-magic fix. I'll turn it off

You can keep it on if you like, so long as any changes it makes are in a separate commit (e.g. save the page without making your changes, and if there's a change caused by your IDE, commit that result first and _then_ make your PR changes) - though I could see that workflow being a pain to remember to do, especially if it differs from your regular workflow.

GuySartorelli (approved):
Looks great, thanks!

rasstislav:
Hi @GuySartorelli. Hi @wilr.

This issue also exists v4. Is it possible to backport this patch commit to this version as well?

Thanks

GuySartorelli:
@rasstislav Absolutely.
The easiest way to do that would be to just open a new PR introducing this change to the `4.11` branch.

rasstislav:
@GuySartorelli ready, thanks :)

PR #1085, opened 2023-08-21
FIX: Save repeated database queries to fetch elemental area name (fixes #928)

### LINKED ISSUE #928: Repeated database queries from BaseElement::getAreaRelationName()
During element rendering, `forTemplate()` calls `getRenderTemplates()`, which in turn calls `getAreaRelationName()` to help build a list of applicable templates (presumably so that elements can be rendered differently based on which area they belong to).

`BaseElement::getAreaRelationName()` calls `$this->Parent()`, which runs through a typical flow for fetching a has_one component. However, these calls aren’t cached so this results in an additional database query for each element as it’s rendered:

https://github.com/silverstripe/silverstripe-elemental/blob/f626840fd6d1d4267e49014eaa55ae94bc490b85/src/Models/BaseElement.php#L758-L760

I’d suggest we replace this with something like:
```php
$class = DataObject::getSchema()->hasOneComponent($this, 'Parent');
$area = ($this->ParentID) ? DataObject::get_by_id($class, $this->ParentID) : null;
$area = $area ?: ElementalArea::singleton();
```

lozcalver:
Issue: #928

### DIFF
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -894,16 +894,19 @@ JS
         $result = 'ElementalArea';
 
         if ($page) {
-            $has_one = $page->config()->get('has_one');
-            $area = $this->Parent();
+            $class = DataObject::getSchema()->hasOneComponent($this, 'Parent');
+            $area = $this->ParentID ? DataObject::get_by_id($class, $this->ParentID) : null;
 
-            foreach ($has_one as $relationName => $relationClass) {
-                if ($page instanceof BaseElement && $relationName === 'Parent') {
-                    continue;
-                }
-                if ($relationClass === $area->ClassName && $page->{$relationName}()->ID === $area->ID) {
-                    $result = $relationName;
-                    break;
+            if ($area) {
+                $has_one = $page->config()->get('has_one');
+                foreach ($has_one as $relationName => $relationClass) {
+                    if ($page instanceof BaseElement && $relationName === 'Parent') {
+                        continue;
+                    }
+                    if ($relationClass === $area->ClassName && $page->{$relationName}()->ID === $area->ID) {
+                        $result = $relationName;
+                        break;
+                    }
                 }
             }
         }


src/Models/BaseElement.php
```diff
@@ -895,7 +895,9 @@ public function getAreaRelationName()
 
         if ($page) {
             $has_one = $page->config()->get('has_one');
-            $area = $this->Parent();
+            $class = DataObject::getSchema()->hasOneComponent($this, 'Parent');
+            $area = ($this->ParentID) ? DataObject::get_by_id($class, $this->ParentID) : null;
+            $area = $area ?: ElementalArea::singleton();
```
GuySartorelli: What's the idea behind getting a singleton here? This feels like a bit of a hack tbh. The get_by_id part is sensible but I don't understand why you're getting a singleton if that fails.

Assuming it is sensible, we should use the null coallescing asignment operator instead of `?:` since this shouldn't be falsy unless it's explicitly null.
```suggestion
            $area ??= ElementalArea::singleton();
```

GuySartorelli (changes requested):

src/Models/BaseElement.php
lozcalver: > What's the idea behind getting a singleton here? This feels like a bit of a hack tbh. The get_by_id part is sensible but I don't understand why you're getting a singleton if that fails.

Just matching the `$this->Parent()` call - `DataObject::getComponent()` will return an empty record if the record doesn‘t exist. I suppose we could remove that and wrap the code that follows in an `if ($area)` check instead?

lozcalver: I’ve replaced the singleton logic with an `if ($area)` statement

GuySartorelli: Thanks! I think that's more straightforward and more consistent with similar logic elsewhere in core and supported modules.

GuySartorelli (approved):
Looks good, works well locally.

Issue #1093, opened 2023-08-30
Elements and ElementalAreas duplicated when restoring page from archived state


AljosaB:
## Setup
Elemental + Fluent as described in [advanced documentation](https://github.com/silverstripe/silverstripe-elemental/blob/5/docs/en/advanced_setup.md#elemental-with-fluent-setup)

## Steps to Reproduce
- create a page, add any element and remember element's ID
- archive page
- restore page
- check element's (new) ID

## Issue
```php
public function onBeforeWrite()
{
    parent::onBeforeWrite();

    if (!$this->isDraftedInLocale() && $this->isInDB()) {
        $elementalArea = $this->ElementalArea();

        $elementalAreaNew = $elementalArea->duplicate();
        $this->ElementalAreaID = $elementalAreaNew->ID;
    }

    return;
}
```

When page is being restored the code above gets triggered and duplicates elemental area and it's elements (cascade_duplicates on elemental area). This is not a desired state in my opinion as it can potentially break some things. For example you could have an ElementVirtual linking to element and relying on it's ID to stay the same.

## PRs
- https://github.com/silverstripe/silverstripe-elemental/pull/1094

GuySartorelli:
PR merged. Thanks for sorting that out.

PR #1094, opened 2023-08-30
Fix not needed duplication when restoring a page from archived state

AljosaB:
Fix for https://github.com/silverstripe/silverstripe-elemental/issues/1093

### DIFF
--- a/docs/en/advanced_setup.md
+++ b/docs/en/advanced_setup.md
@@ -201,7 +201,7 @@ class BlockPage extends Page
     {
         parent::onBeforeWrite();
 
-        if (!$this->isDraftedInLocale() && $this->isInDB()) {
+        if (!$this->isDraftedInLocale() && $this->isInDB() && !$this->isArchived()) {
             $elementalArea = $this->ElementalArea();
 
             $elementalAreaNew = $elementalArea->duplicate();


GuySartorelli:
The issue describes a perceived issue with the current behaviour and doesn't mention fluent. Your change is to documentation, specifically for the "Elemental with Fluent setup" section....

Can you please describe a bit more how this change to documentation fixes the issue you're linking to?

AljosaB:
The issue does mention fluent, it's literally the second word under the setup section :) Also the provided [link](https://github.com/silverstripe/silverstripe-elemental/blob/5/docs/en/advanced_setup.md#elemental-with-fluent-setup) is pointing to the fluent part of documentation and quoted part of code is from the same section :)

The intention is to improve the suggested code in the documentation when using elemental with fluent. The issue mentioned (I'm the author of both) is not describing a malfunction / bug in elemental's code, but in the documentation's code.

GuySartorelli:
> The issue does mention fluent, it's literally the second word under the setup section

My apologies, I missed that. It's not mentioned after that so I guess I was focusing on the main description, and the issue didn't make it very clear that this was only a documentation problem.

> I'm the author of both

Lol yes, thank you, I can see that.

Please update your commit message to start with the `DOC` prefix as per our [contribution documentation](https://docs.silverstripe.org/en/5/contributing/code/#commit-messages) so that it doesn't trigger an automated patch release.

AljosaB:
Indeed, it wasn't very clear it's only about the docs.

Commit message fixed :+1:

GuySartorelli (approved):
Seems okay. Thanks for submitting it. I'll merge when it goes green

Issue #1099, opened 2023-09-18
BUG: handling of linked element in virtual element failed


rasstislav:
Hello,

I have [ElementForm](https://github.com/dnadesign/silverstripe-elemental-userforms/blob/3.3.3/src/Model/ElementForm.php)  as  [LinkedElement in ElementVirtual](https://github.com/dnadesign/silverstripe-elemental-virtual/blob/44635640e73b0f076618ac8135f738a8f3a4efe3/src/Model/ElementVirtual.php#L21) and after submitting the form I got an error:

`Fatal error: Element $id not found for this page in /var/www/html/vendor/dnadesign/silverstripe-elemental/src/Extensions/ElementalContentControllerExtension.php on line 62`.

## PRs
- https://github.com/silverstripe/silverstripe-elemental/pull/1100

GuySartorelli:
Thanks for pointing this out.
Based on the additional context provided in your pull request, it sounds like this is something that should and can be resolved in the virtual module instead. I'll close this issue, but if you find that this is something that needs to be handled here instead please provide additional context explaining why.

PR #1100, opened 2023-09-18
feat: added support for handle linked element in virtual element

rasstislav:
issue #1099


wilr:
@rasstislav Is this fixed via https://github.com/dnadesign/silverstripe-elemental-virtual/commit/467885dea8cb3867f3839aa286898b7b1dd13cd5?

Ideally this module does not know or get coupled to the virtual one.

rasstislav:
Hi @wilr. Wow, maybe yes, but it's not available for version 1.5 (elemental 4). Is it possible to add this fix to an older version as well?

GuySartorelli:
Thank you for contributing but I'm going to close this PR - as wilr points out, this module shouldn't have any knowledge or coupling to the virtual elements module so this change is fundamentally the wrong approach. Ideally this will be fixed in the other module, and it sounds like backporting an existing commit in that module will achieve that.

PR #1103, opened 2023-09-25
FIX Element::getPage returning the wrong page

rasstislav:
CMS 4 port of #1082


GuySartorelli (approved):
Looks good, will merge on green

PR #1107, opened 2023-10-11
fix form submission, when form block is inside another block

sukhwinder-somar:
Creating new PR, related to the https://github.com/dnadesign/silverstripe-elemental-userforms/pull/78

## Issue
- https://github.com/dnadesign/silverstripe-elemental-userforms/issues/27

### DIFF
--- a/src/Extensions/ElementalContentControllerExtension.php
+++ b/src/Extensions/ElementalContentControllerExtension.php
@@ -35,9 +35,7 @@ class ElementalContentControllerExtension extends Extension
         }
 
         foreach ($elementalAreaRelations as $elementalAreaRelation) {
-            $element = $elementOwner->$elementalAreaRelation()->Elements()
-                ->filter('ID', $id)
-                ->First();
+            $element = $this->findElement($elementOwner->{$elementalAreaRelation}()->Elements(), $id);
 
             if ($element) {
                 return $element->getController();
@@ -46,5 +44,36 @@ class ElementalContentControllerExtension extends Extension
 
         user_error('Element $id not found for this page', E_USER_ERROR);
         return false;
+    }
+
+    private function findElement($elements, $id)
+    {
+        $element = $elements->filter('ID', $id)->First();
+
+        if ($element) {
+            return $element;
+        }
+
+        foreach ($elements as $el) {
+            if (!$el->hasMethod('Elements')) {
+                continue;
+            }
+
+            $subElementAreaRelations = $el->getElementalRelations();
+
+            if (!$subElementAreaRelations) {
+                continue;
+            }
+
+            foreach ($subElementAreaRelations as $subElementalAreaRelation) {
+                $element = $this->findElement($el->{$subElementalAreaRelation}()->Elements(), $id);
+
+                if ($element) {
+                    return $element;
+                }
+            }
+        }
+
+        return null;
     }
 }
--- a/src/Extensions/ElementalContentControllerExtension.php
+++ b/src/Extensions/ElementalContentControllerExtension.php
@@ -2,9 +2,10 @@
 
 namespace DNADesign\Elemental\Extensions;
 
+use SilverStripe\Core\Extension;
+use SilverStripe\ORM\DataObject;
 use DNADesign\Elemental\Models\ElementalArea;
 use DNADesign\Elemental\Extensions\ElementalAreasExtension;
-use SilverStripe\Core\Extension;
 
 class ElementalContentControllerExtension extends Extension
 {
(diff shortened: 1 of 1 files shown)


GuySartorelli (changes requested):
Please change to this instead:

```php
public function handleElement()
{
    $id = $this->owner->getRequest()->param('ID');

    if (!$id) {
        user_error('No element ID supplied', E_USER_ERROR);
        return false;
    }

    /** @var DataObject $elementOwner */
    $elementOwner = $this->owner->data();

    if (!$elementOwner->hasExtension(ElementalAreasExtension::class)) {
        user_error(get_class($elementOwner) . ' doesnt have the ElementalAreasExtension applied', E_USER_ERROR);
        return false;
    }

    $elementalAreaRelations = $elementOwner->getElementalRelations();
    if (!$elementalAreaRelations) {
        user_error(get_class($elementOwner) . ' has no ElementalArea relationships', E_USER_ERROR);
        return false;
    }

    $element = $this->findElement($elementalAreaRelations, $elementOwner, $id);

    if ($element) {
        return $element->getController();
    }

    user_error('Element $id not found for this page', E_USER_ERROR);
    return false;
}

private function findElement(iterable $elementalAreaRelations, DataObject $owner, $id): ?DataObject
{
    foreach ($elementalAreaRelations as $elementalAreaRelation) {
        $elements = $owner->$elementalAreaRelation()->Elements();
        $found = $elements->filter('ID', $id)->First();

        if ($found) {
            return $found;
        }

        /** @var BaseElement $element */
        foreach ($elements as $element)             {
            if (!$element->hasExtension(ElementalAreasExtension::class)) {
                continue;
            }

            /** @var BaseElement&ElementalAreasExtension $element */
            $found = $this->findElement($element->getElementalRelations(), $element, $id);

            if ($found) {
                return $found;
            }
        }
    }

    return null;
}
```

This is similar to your approach, but with the following key differences:
- Adds an additional protection against `$elementOwner` not having the `ElementalAreasExtension` applied (in the case of custom nested elements)
- Checks for use of the `ElementalAreasExtension` instead of the `Elements` method, which allows for the relation to have any arbitrary name instead of just the one that the elemental list module uses
- Flips the logic of `findElement()` somewhat, so we have no `foreach` loop in the `handleElement()` anymore, delegating all of the logic to `findElement()`.
- Adds appropriate type hinting to the new method

I've tested this in conjunction with https://github.com/dnadesign/silverstripe-elemental-userforms/pull/78 and together they resolve the bug, so after you've made these changes I'll be happy to merge them both.

sukhwinder-somar:
sweet, I have updated the code with your changes, but I didn't test it so I'm hoping it will work as you have already tested it....

sukhwinder-somar:
sweet, I have updated the code with your changes, but I didn't tested it so I'm hoping it will work as you have already tested it....

GuySartorelli:
Please test the changes. I will merge once you have confirmed they work on your end.
The whole point of a pull request is that there are at least two people involved.

sukhwinder-somar:
Sure, then I will update you tomorrow or tonight after testing...

sukhwinder-somar:
Hi,

I have tested this and can confirm the above is working for me on First and Second Level of page.

While testing I had first and second level forms on same page and I found out If I was submitting one of them, then both of them was showing confirmation messages, which I feel looks weird, so thought I should tell you....

thanks,

GuySartorelli:
> While testing I had first and second level forms on same page and I found out If I was submitting one of them, then both of them was showing confirmation messages, which I feel looks weird, so thought I should tell you....

That's a known issue, captured in https://github.com/dnadesign/silverstripe-elemental-userforms/issues/50 - it was fixed recently so if you update your dependency it shouldn't happen anymore.

GuySartorelli (approved):
Looks good to me, thanks for fixing this.

GuySartorelli:
Ahh, sorry, I just realised this is targetting the wrong branch.
Please retarget it to the `4.11` branch. You may need to [reset your commits](https://gist.github.com/GuySartorelli/74627843aeb75e12b777b1f44cb22d2f) afterward.

sukhwinder-somar:
All good, thanks for all support 😊 🙏

GuySartorelli (changes requested):
Oops, looks like when you reset the commits you accidentally reverted the changes from https://github.com/silverstripe/silverstripe-elemental/pull/1107#pullrequestreview-1672672600 as well!

sukhwinder-somar:
apologies again, I forgot to cherry pick both commits, but have added it now 😅

GuySartorelli (approved):

Issue #1104, opened 2023-09-28
Upload field multi selection auto detection breaking with inline editor


mfendeksilverstripe:
CMS version: `4.13`
Elemental version `4.9.0`

`UploadField::getIsMultiUpload()` has auto-detection capability - it can guess if multiple fields are allowed or not based on the related DB record.

To reproduce this, create a `has_one` relation to `File` on a new elemental block, and add an `UploadField` without explicitly setting `isMultiUpload`. It will accept multiple files.

<details>
<summary>More specific details</summary>

```php
public function getIsMultiUpload()
{
    if (isset($this->multiUpload)) {
        return $this->multiUpload;
    }
    // Guess from record
    $record = $this->getRecord();
    $name = $this->getName(); // <------- this breaks

    // Disabled for has_one components
    if ($record && DataObject::getSchema()->hasOneComponent(get_class($record), $name)) {
        return false;
    }
    return true;
}
```

Using the upload field within a content block that uses inline editing will break this feature unfortunately. This is caused by `EditFormFactory::namespaceFields()` which changes the field names.

```php
protected function namespaceFields(FieldList $fields, array $context)
{
    $elementID = $context['Record']->ID;

    foreach ($fields->dataFields() as $field) {
        $namespacedName = sprintf(self::FIELD_NAMESPACE_TEMPLATE ?? '', $elementID, $field->getName());
        $field->setName($namespacedName);// <------ field name change
    }
}
```

This will change for example `Image` to `PageElements_153_Image`. This will break the auto-detection feature of the `UploadField` as it can no longer recognise the field by name, so it falls back to default.
</details>

<details>
<summary>Workaround</summary>

This is my workaround - sublcass upload field and override the `getIsMultiUpload()` method.

```php
/**
 * Can't use directly @see EditFormFactory::FIELD_NAMESPACE_TEMPLATE
 */
private const NAMESPACE_PREFIX = 'PageElements_';

/**
 * Copied from @see UploadField::getIsMultiUpload()
 * Fixed the issue when this field is used in the elemental inline editor context
 * The field names are prefixed which breaks the auto-detection feature of this field
 * Issue caused by @see EditFormFactory::namespaceFields()
 * GitHub issue raised: https://github.com/silverstripe/silverstripe-elemental/issues/1104
 *
 * @return bool
 */
public function getIsMultiUpload(): bool
{
    if ($this->multiUpload !== null) {
        // Explicit settings
        return (bool) $this->multiUpload;
[...]

PR #1108, opened 2023-10-11
FIX Apply multiupload auto-detection for upload fields

GuySartorelli:
UploadField checks hasone components based on the field name to see if it can be a multiUpload field or not. We need to trigger that before changing the name of the field.

## Issue
- https://github.com/silverstripe/silverstripe-elemental/issues/1104

### DIFF
--- a/src/Forms/EditFormFactory.php
+++ b/src/Forms/EditFormFactory.php
@@ -2,6 +2,7 @@
 
 namespace DNADesign\Elemental\Forms;
 
+use SilverStripe\AssetAdmin\Forms\UploadField;
 use SilverStripe\Control\RequestHandler;
 use SilverStripe\Core\Config\Configurable;
 use SilverStripe\Forms\DefaultFormFactory;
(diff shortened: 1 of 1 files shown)


emteknetnz (approved):
Well that was easy :-)

Tested locally, works good

emteknetnz:
In case anyone's interested here's the code I used to validate the before and fater

```php
<?php

use SilverStripe\Assets\File;
use SilverStripe\AssetAdmin\Forms\UploadField;

class MyElement extends BaseElement
{
    private static array $has_one = [
        'MyFile' => File::class
    ];

    public function getType()
    {
        return 'My block';
    }

    {
        $fields->addFieldToTab('Root.Main', new UploadField('MyFile', 'MyFile'));
        return $fields;
    }
}

```

emteknetnz:
In case anyone's interested here's the code I used to validate the before and after

```php
<?php

use SilverStripe\Assets\File;
use SilverStripe\AssetAdmin\Forms\UploadField;

class MyElement extends BaseElement
{
    private static array $has_one = [
        'MyFile' => File::class
    ];

    public function getType()
    {
        return 'My block';
    }

    {
        $fields->addFieldToTab('Root.Main', new UploadField('MyFile', 'MyFile'));
        return $fields;
    }
}

```

Issue #1068, opened 2023-06-25
SPIKE Come up with an approach to validate inline blocks


maxime-rainville:
We have a gnarly card to implement validation for inline blocks.

## TODO
- Document the various ways in which inline blocks are saved.
- Review possible a ways to validate form schemas ond other potential form schema improvements that could help with Elemental
- Identify potential approaches to implement validation for inline blocks. Ideally those could be implemented in a minor.

## Timebox
2 days (with potential for extension if need be)

## Parent card 
https://github.com/silverstripe/silverstripe-elemental/issues/329

## PRs
- https://github.com/silverstripe/silverstripe-elemental/pull/1113

emteknetnz:
tl;dr - IMO there is only one "correct" way to handle inline validation which is to utilise the existing FormSchema.php + FormBuilder.js functionality to do everything in relation to inline XHR saves and validation.

IMO there aren't any good viable alternatives to doing this, so I haven't provide any

What's happening right now is we use the existing functionality for rendering the inline form, but not for validating/saving the data in it.

I've got a [draft PR](https://github.com/silverstripe/silverstripe-elemental/pull/1113) that is the net result of my investigation. This was very slow and painful trying to work out what's currently happening at trying to move it toward the state it should have always been in. As usual everything is hard with incomplete implementations, things are undocumented and there's lots of jumping around the place.

Going forward I think we should simply continue with the draft PR and work on that until it's all wired together.  This may require some work in silverstripe/admin FormBuilder.js as well. We could also consider if we want a more [centralised FormSchema API controller](https://github.com/silverstripe/silverstripe-admin/issues/1589), though the work here could and possibly should happen independently of that for now and we just migrate to that later.

The following is a bit of a dump of various things:

**GETing elemental data:**
- Frontend uses FormBuilderLoader.js to request a FormSchema
- GET FormSchema request is made to /admin/elemental-area/schema/elementForm/3
- There is a controllers ElementalAreaController.php extends CMSMain (which extends LeftAndMain) which has a $url_segment = 'elemental-area'
- Routing goes to LeftAndMain::schema() with params $FormName = elementForm and $ItemID = 3
- In LeftAndMain::schema() .. $formMethod = "get{$formName}" .. $form = $this->{$formMethod}($ItemID); which translates to ElementalAreaController::getElementForm($elementID): Form;
[...]

Issue #329, opened 2018-08-21
Implement validation method for inline editable form


brynwhyman:
Validation messaging should be handled within the custom form React component as opposed to at a page level.

There are currently two ways to submit content for validation - you can save the page, or you can save an individual block.

## Acceptance criteria
- In-line editable content block are validated with form schema as outline in the SPIKE.
- ~~Validation occurs both on page save and on individual block save.~~ (new issue created)
- API for enforcing validation of content block is inline with the one to validate pages.
- ~~Validation works both for inline block and "not inline" block.~~ (new issue created)
- Official elemental doc provides clear guidance on how to validate elemental block.

## Note
- Need to create a follow up UX card for handling block that are broken and general display of validation errors.
  - @GuySartorelli has some good questions about [what the UI should do with validation errors](https://github.com/silverstripe/silverstripe-elemental/issues/329#issuecomment-1586339153).
- Can we constrained this card to validating a block that is currently being edited?
- Related card: https://github.com/silverstripe/silverstripe-framework/issues/11115

## SPIKE
There was a spike to investigate this (which includes a POC PR):
- https://github.com/silverstripe/silverstripe-elemental/issues/1068

## New issues created
- https://github.com/silverstripe/silverstripe-elemental/issues/1154
- https://github.com/silverstripe/silverstripe-elemental/issues/1155

## PRs
- https://github.com/silverstripe/silverstripe-elemental/pull/1150
- https://github.com/silverstripe/silverstripe-admin/pull/1685
- https://github.com/silverstripe/developer-docs/pull/466
- https://github.com/silverstripe/developer-docs/pull/467
- https://github.com/silverstripe/silverstripe-frameworktest/pull/169
- https://github.com/silverstripe/silverstripe-elemental/pull/1169

## Multi-PR CI run
- https://github.com/emteknetnz/silverstripe-installer/actions/runs/8610397572

clarkepaul:
We have some initial designs for the notification placement but we haven't gone through each scenario to figure out what the notifications should say in each case. Please refer to the Style Guide for guidance https://projects.invisionapp.com/dsm/silver-stripe/silver-stripe

ScopeyNZ:
Note at this point we'll ideally have completed #327 which implements a custom React form component. With this (and the messages that are a part of FormBuilderLoader) we should be able to expose the messages we get back and ship them up to the parent component.

chrispenny:
Hi team,

[...]

PR #1113, opened 2023-10-16
SPIKE Inline validation

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-elemental/issues/1068

DO NOT MERGE

This PR is only for investigative purposes


GuySartorelli:
There's so much going on here it's hard to see what's doing what. Can you please add a few notes in the PR description explaining what all this PR is doing?

emteknetnz:
I don't think this PR is worth peer reviewing, it's not in a good state and it's not complete. It's not meant to be merged, it's kind of just a dumping ground for code I used to do investigation

GuySartorelli:
Is there anything in this PR that's useful for understanding and/or evaluating the high level approach you're recommending in the issue? Currently the issue is more of an info dump than a "this is the recommended approach" so it's a bit hard to follow from that angle.

GuySartorelli:
Closing this for now - we may end up copying some bits from this when we implement this validation in https://github.com/silverstripe/silverstripe-elemental/issues/329 but it likely won't use this exact PR.

PR #1121, opened 2023-11-28
PATCH: removing superfluous (and breaking) Controller::join_links

sunnysideup:
The Controller::join_links function removed the ? in the Preview Link.


GuySartorelli:
Join links shouldn't break query strings. Please raise a pr to fix join_links in framework instead, as that's where the bug actually is.

Issue #1122, opened 2023-12-06
Infinite rendering loop when rendering a FormBuilder as part of an ElementAction


emteknetnz:
When rendering a FormBuilder component as part of an ElementAction, ElementEditor will go into an infinite loop of re-rendering ElementEditor

This started happening as part of this change https://github.com/silverstripe/silverstripe-elemental/pull/1044

## PR
- https://github.com/silverstripe/silverstripe-elemental/pull/1123

GuySartorelli:
Merged

PR #1123, opened 2023-12-06
FIX Handle empty element formstate

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-elemental/issues/1122


GuySartorelli (approved):
LGTM - matches what we discussed about this.

PR #1133, opened 2024-01-18
FIX Null submit handler for element component

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-linkfield/issues/122


client/src/components/ElementEditor/InlineEditForm.js
```diff
@@ -53,7 +53,12 @@ class InlineEditForm extends PureComponent {
       schemaUrl,
       identifier: 'element',
       refetchSchemaOnMount: !formHasState,
-      onLoadingError: this.handleLoadingError
+      onLoadingError: this.handleLoadingError,
+      // This null submit handler is used to prevent the FormBuilder form from submitting when any nested
+      // Formbuilder forms, such as LinkField v4+, are themselves submitted
+      // This Formbuilder form is not submitted in the traditional sense i.e. there is no rendered submit button
+      // instead the inline save button will do its own submission in SaveAction.js
```
michalkleiner: a minor tweak for consistency
```suggestion
      // This null submit handler is used to prevent the FormBuilder form from submitting when any nested
      // FormBuilder forms, such as LinkField v4+, are themselves submitted
      // This FormBuilder form is not submitted in the traditional sense i.e. there is no rendered submit button
      // instead the inline save button will do its own submission in SaveAction.js
```

GuySartorelli (changes requested):
I now get this error instead when creating or editing a link inside an elemental block:

> Uncaught Error: Promise was not returned for submitting

client/src/components/ElementEditor/InlineEditForm.js
emteknetnz: Updated

GuySartorelli (approved):
Works great

PR #1148, opened 2024-02-21
SPIKE Inline validation

emteknetnz:

Do not review or merge this

This is just for an investigation

This is the same branch that was used on https://github.com/silverstripe/silverstripe-elemental/pull/1113

Note https://github.com/silverstripe/silverstripe-elemental/pull/1150 is a more up to date version of this - this PR may still be useful because it contains code for doing a remote submit that there's still a chance we may use

### DIFF
--- a/client/src/components/ElementActions/SaveAction.js
+++ b/client/src/components/ElementActions/SaveAction.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useContext } from 'react';
 import { compose } from 'redux';
 import { connect } from 'react-redux';
 import AbstractAction from 'components/ElementActions/AbstractAction';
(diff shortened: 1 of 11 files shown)


client/src/components/ElementActions/SaveAction.js
```diff
@@ -34,6 +40,14 @@ const SaveAction = (MenuComponent) => (props) => {
       { type: type.title }
     );
 
+    // https://redux-form.com/8.3.0/examples/remotesubmit/
+    submitForm();
```
emteknetnz, line 44: There was a callback below this to show toast notifications, however I don't see us getting this to work because redux-forms 'submit()' function used for triggering a remote submit isn't thenable nor does it allow for adding a callback or have any other method that allows us to do stuff 'onAfterSuccess / onAfterFailure'

Make me inlined to say that remote submit isn't feasible and we need to put the save/submit button actually on the form

PR #1150, opened 2024-02-26
NEW Validation of fields when inline saving

emteknetnz:

Requires https://github.com/silverstripe/silverstripe-admin/pull/1685 to be functional

Requires https://github.com/silverstripe/silverstripe-frameworktest/pull/169 for behat to pass

Use the following test block

```php
<?php

use SilverStripe\Forms\CompositeValidator;
use SilverStripe\Forms\RequiredFields;
use SilverStripe\Forms\UrlField;

{
    private static $db = [
        'MyField' => 'Varchar',
        'MyHTML' => 'HTMLText',
        'MyUrl' => 'Varchar',
    ];

    private static $table_name = 'MyBlock';

    private static $singular_name = 'My Block';

    private static $plural_name = 'My Blocks';

    private static $description = 'This is my block';

    private static $icon = 'font-icon-block-content';

    public function getType()
    {
        return 'My Block';
    }

    public function validate()
    {
        $result = parent::validate();
        if ($this->Title == 'x') {
            $result->addFieldError('Title', 'Title cannot be x', 'validation');
        }
        if ($this->MyField == 'x') {
            $result->addFieldError('MyField', 'MyField cannot be x', 'validation');
        }
        $c = trim(strip_tags($this->MyHTML ?? ''));
        if ($c == 'x') {
            $result->addFieldError('MyHTML', 'MyHTML cannot be x', 'validation');
        }
        if ($this->MyField == 'z' && $c == 'z') {
            $result->addError('This is a general error message');
        }
        return $result;
    }

    public function getCMSCompositeValidator(): CompositeValidator
    {
        return CompositeValidator::create([RequiredFields::create(['Title'])]);
    }

    {
        $fields->dataFieldByName('MyHTML')->setRows(5);
        $fields->removeByName('MyUrl');
        // URLfield contains its own validate() methods
        $fields->addFieldToTab('Root.Main', UrlField::create('MyUrl', 'My URL'));
        return $fields;
    }
}
```


src/Controllers/ElementalAreaController.php
```diff
@@ -28,18 +31,25 @@ class ElementalAreaController extends CMSMain
...
+        // Copying LinkField
+        'elementForm/$ItemID' => 'elementForm',
         // API access points with structured data
         'POST api/saveForm/$ID' => 'apiSaveForm',
         '$FormName/field/$FieldName' => 'formAction',
     ];
 
     private static $allowed_actions = [
         'elementForm',
-        'schema',
```
emteknetnz, line 38: This is already an allowed action in LeftAndMain

client/src/components/ElementActions/SaveAction.js
```diff
@@ -1,128 +1,56 @@
...
+  const saveElement = () => {
+    submitForm();
+    onAfterSave();
   };
 
   const newProps = {
     title: i18n._t('ElementSaveAction.SAVE', 'Save'),
     className: 'element-editor__actions-save',
     onClick: handleClick,
+    toggle: props.toggle,
```
emteknetnz, line 30: toggle is required to prevent an error on SaveAction-test.js jest test

Have it here is consistent with other action components e.g. PublishAction.js

src/Controllers/ElementalAreaController.php
```diff
@@ -216,4 +207,130 @@ public static function removeNamespacesFromFields(array $data, $elementID)
...
+     * @param HTTPRequest $request
+     * @return HTTPResponse|null JSON encoded string or null if an exception is thrown
+     * @throws HTTPResponse_Exception
+     *
+     * @deprecated 5.3.0 Send a POST request to elementForm/$ItemID instead
+     */
+    public function apiSaveForm(HTTPRequest $request)
+    {
+        Deprecation::notice('5.3.0', 'Send a POST request to elementForm/$ItemID instead');
+        throw new HTTPResponse_Exception('This endpoint should not be used');
```
emteknetnz, line 196: There's no single reason to call this now that we're using FormSchema to save data.  Previously this method handled POSTed form data

If for some reason we wanted to keep this working (which I recommend we do not), then at some point a while back this code did work, though possibly in no longer does

```php
$data = $request->postVars();
$id = $request->param('ID');
$form = $this->getElementForm($id);
if (!$form) {
    $this->jsonError(400);
}
// If the following fails validation, a FormSchema response will be returned to the frontend
// due to a ValidationException being thrown and then caught in core code
$this->save($data, $form);
// Send legacy response
return $this->jsonSuccess(200, [
    'status' => 'success',
    'updated' => $this->elementUpdated,
]);
```

And also we'd need to un-delete any code related to `$elementUpdated` in the PR

client/src/components/TextCheckboxGroupField/TextCheckboxGroupField.js
```diff
@@ -8,10 +8,13 @@ const TextCheckboxGroupField = (props) => {
   // Map out the children and clone to set the "noHolder" prop on them.
   const childrenWithProps = React.Children.toArray(
     React.Children.map(children, (child, index) => {
-      const additionalProps = { noHolder: true };
```
emteknetnz, line 11: Removing noHolder is required to get validation messages showing on the TextField of this composite field

src/Controllers/ElementalAreaController.php
```diff
@@ -45,132 +47,121 @@ public function getClientConfig()
...
-        // Get ID either from posted back value, or url parameter
-        if (!$request) {
-            $this->jsonError(400);
-            return null;
-        }
-        $id = $request->param('ID');
+        $request = $request ?: $this->getRequest();
+        $id = $request->param('ItemID');
+
+        // Respect previous behaviour
```
emteknetnz, line 70: I'm probably being overly backwards compatible here by not changing HTTP status codes in a minor release. Especially considering I've completely disabled apiSaveForm(). Feel free to recommend that we don't respect previous http codes and change it to something better

src/Controllers/ElementalAreaController.php
```diff
@@ -186,7 +177,7 @@ public function formAction(HTTPRequest $request)
         $formName = $request->param('FormName');
 
         // Get the element ID from the form name
-        $id = substr($formName ?? '', strlen(sprintf(self::FORM_NAME_TEMPLATE ?? '', '')));
```
emteknetnz, line 189: This is unrelated refactoring to remove automatically added `??` which will never be required

GuySartorelli:
I have tried this locally and found some problems as outlined below.
I haven't looked at the code yet, but presumably fixing these will require some substantial changes so I'll hold off on that.

Please add behat tests to cover the below scenarios one they're fixed.

## Things that don't work

- When I save the page, form (i.e. `RequiredFields` and the `URLField`) validation for the block is not triggered.
- When saving the block on its own, if there are multiple form-level validation errors (e.g. empty "Title" field and invalid "MyUrl" value), only the first error is displayed
- When I save the page, if there are errors on the block, clicking "save" on the block does nothing. For example:
    1. Set the "Title" and "MyField" values to "x"
    2. Save the page - get validation errors.
    3. Set the value of all fields to "z"
    4. Save the block - there is no visual change, it seems like nothing at all happened.
    5. Set all the fields to valid values
    6. Save the block - still nothing happens.

emteknetnz:
> When I save the page, if there are errors on the block, clicking "save" on the block does nothing

Have now fixed this with latest code push and written a behat test for it

> When I save the page, form (i.e. RequiredFields and the URLField) validation for the block is not triggered.

I noted on this [comment](https://github.com/silverstripe/silverstripe-elemental/issues/329#issuecomment-1982283254) on the original issue that I was reducing the scope of this issue to only focus on inline save and I've [split off a new issue](https://github.com/silverstripe/silverstripe-elemental/issues/1155) to deal with page save because that traverses a different code path and has its own set of issues

> When saving the block on its own, if there are multiple form-level validation errors (e.g. empty "Title" field and invalid "MyUrl" value), only the first error is displayed

That's an issue with Silverstripe validation in general where different types of validation errors result in exception being thrown and caught in different places without running all of the validation types being processed. If we want to tackle then it should be done in different card.

client/src/components/ElementActions/SaveAction.js
```diff
@@ -1,128 +1,56 @@
...
-              'ElementSaveAction.ERROR_NOTIFICATION',
-              'Error saving \'{title}\''
-            ),
-            { title: element.Title || noTitle }
-          ),
-          stay: false,
-          type: 'error'
-        });
-      });
+  const saveElement = () => {
```
emteknetnz, line 21: This is kind of unnecessary convoluted in that formHasRendered, doSaveElement, submitForm and onAfterSave() all come in via Context set on Element.js.  We could actually do just about everything within Element.js.

The reason it's like this is to mirror PublishAction.js which does have some GraphQL HOC logic that's not available in Element.js

GuySartorelli:
> > When I save the page, form (i.e. RequiredFields and the URLField) validation for the block is not triggered.
> 
> I noted on this [comment](https://github.com/silverstripe/silverstripe-elemental/issues/329#issuecomment-1982283254) on the original issue that I was reducing the scope of this issue to only focus on inline save and I've [split off a new issue](https://github.com/silverstripe/silverstripe-elemental/issues/1155) to deal with page save because that traverses a different code path and has its own set of issues

I think the acceptance criteria need to be updated then. 

> > When saving the block on its own, if there are multiple form-level validation errors (e.g. empty "Title" field and invalid "MyUrl" value), only the first error is displayed
> 
> That's an issue with Silverstripe validation in general where different types of validation errors result in exception being thrown and caught in different places without running all of the validation types being processed. If we want to tackle then it should be done in different card.

That's not the case. Try it on a page with a URLField and a Required fields validator. Errors for both of those will be included together, because those are both managed through the form validator. 

You're thinking of form validation Vs the `validate` method on the model i think - and I agree those should remain separate for the purposes of this pr.

emteknetnz:
> I think the acceptance criteria need to be updated then.

Have updated ACs

> That's not the case. Try it on a page with a URLField and a Required fields validator. Errors for both of those will be included together, because those are both managed through the form validator.

Possibly what's happening in this instance is javascript validation. If you try that particular setup on an inline editable block then what will happen is that only the Title validation will fire because there's javascript validation that'll happen first, there won't be an XHR request fired.  Could you double check that there's actually an XHR request being fired?

GuySartorelli:
You're correct, there is no XHR request if there are validation errors due to required fields - but there are for literally any other form or model validation.

This means for elemental blocks there are three stages to validation:
1. Required fields
2. Any other form validator and field validation
3. `validate()` called on the block itself.

For pages and anything edited in a gridfield, this steps 1 and 2 above are combined.

I suspect this also affects other react forms such as linkfield and asset admin, so if you want to split that off as a separate card I'd be okay with that.

GuySartorelli:
Please also add an installer CI run to show that behat passes when all the required PRs are merged.
Then once I'm happy that the approach generally seems to do what it needs to (which if you split the above off as a separate card, I will be), I'll start reviewing the actual code.

emteknetnz:
Mutli PR run is here:

https://github.com/emteknetnz/silverstripe-installer/actions/runs/8291384194

> I suspect this also affects other react forms such as linkfield and asset admin, so if you want to split that off as a separate card I'd be okay with that.

I'm not sure what you want split of what as a separate card? Javascript validation in FormBuilder forms is doing what it's supposed to do?  If we wanted 1 + 2 to go at the same time on elemental then we'd need be to disable javascript validation on FormBuilder forms which would be a BC step backwards.  I think FormField::getSchemaValidation() is what controls what gets passed to Validator.js on the frontend

GuySartorelli:
Your installer CI run doesn't include elemental tests? That was the main purpose of it, form my perspective.

> I'm not sure what you want split of what as a separate card? Javascript validation in FormBuilder forms is doing what it's supposed to do? If we wanted 1 + 2 to go at the same time on elemental then we'd need be to disable javascript validation on FormBuilder forms which would be a BC step backwards. I think FormField::getSchemaValidation() is what controls what gets passed to Validator.js on the frontend

I'll open a separate card about it.

client/src/components/ElementEditor/Element.js
```diff
@@ -53,6 +73,67 @@ class Element extends Component {
...
+  }
+
+  showSavedElementToast() {
+    const title = this.state.newTitle || this.getNoTitle();
+    const message = i18n.inject(
+      i18n._t('ElementSaveAction.SUCCESS_NOTIFICATION', 'Saved \'{title}\' successfully'),
+      { title }
+    );
+    this.props.actions.toasts.success(message);
+  }
```
GuySartorelli: In the old `SaveAction.js` there was also a `'Error saving \'{title}\''` message when there's an error. Why is that not included?

client/src/components/ElementEditor/Element.js
```diff
@@ -53,6 +73,67 @@ class Element extends Component {
...
+    }
+  }
+
+  componentDidUpdate() {
+    if (this.state.justSavedElement) {
+      this.setState({
+        justSavedElement: false,
+      });
+      if (!this.state.doPublishElement) {
+        this.showSavedElementToast();
```
GuySartorelli: It feels weird doing a toast in `componentDidUpdate()` - not arguing that it should be moved necessarily but at least needs a comment saying why this is happening here instead of e.g. at the end of a promise etc.

Update: See https://github.com/silverstripe/silverstripe-elemental/pull/1150#discussion_r1527616657

client/src/components/ElementEditor/Element.js
```diff
@@ -215,6 +296,81 @@ class Element extends Component {
...
+  }
+
+  handlePublishButtonClick() {
+    this.setState({
+      justClickedPublishButton: true,
+      ensureFormRendered: true,
+    });
+    if (this.props.formDirty) {
+      // Save the element first before publishing, which may trigger validation errors
+      this.props.submitForm();
```
GuySartorelli: Feels like we need to set `doPublishElement` to true here too? Or else won't it just save and not publish?

client/src/components/ElementEditor/Element.js
```diff
@@ -215,6 +296,81 @@ class Element extends Component {
...
+        doPublishElement: true,
+      });
+    }
+  }
+
+  handleAfterSave() {
+    this.setState({
+      doSaveElement: false,
+    });
+  }
```
GuySartorelli, line 319: Wouldn't this be a better place for the post-save toast message? Is there some weird timing issue or something that prevents us doing it here?

client/src/components/ElementEditor/Element.js
```diff
@@ -215,6 +296,81 @@ class Element extends Component {
...
+      justClickedPublishButton: true,
+      ensureFormRendered: true,
+    });
+    if (this.props.formDirty) {
+      // Save the element first before publishing, which may trigger validation errors
+      this.props.submitForm();
+    } else {
+      // Just publish the element straight away without saving first
+      this.setState({
+        doPublishElement: true,
```
GuySartorelli: This feels like a really weird way to use state - are we using state as an event triggering mechanism here? What's the advantage of that over just putting the publish logic here or calling a function here that performs publishing logic?

client/src/components/ElementEditor/InlineEditForm.scss
```diff
@@ -25,4 +25,25 @@
...
     }
   }
+
+  // Render the form off the screen so that it's not visible
+  // Note that using display:nome will mean the form is not rendered at all and cannot be submitted
+  &--rendered-not-visible {
+    position: absolute;
+    top: -9999px;
+    left: -9999px;
+  }
```
GuySartorelli: This might affect screen readers in a weird way - this is usually used for "visually hidden" elements i.e. elements that should not be visible but should be read by a screen reader.

Need to make sure that when doing this the focus isn't on any of the elements inside this section, or use some other mechanism to make sure the screen reader doesn't start reading content inside here.

src/Controllers/ElementalAreaController.php
```diff
@@ -45,132 +47,158 @@ public function getClientConfig()
...
     }
 
     /**
-     * @param HTTPRequest|null $request
+     * Used for both:
+     * - GET requests to get the FormSchema via getElementForm() called from LeftAndMain::schema()
+     * - POST Requests to save the Form. Will be handled by to FormRequestHandler::httpSubmission()
+     * /admin/linkfield/elementForm/<ElementID>
+     *
+     * @param HTTPRequest $request - deprecated 5.3.0 Will be removed without equivalent functionality to replace it
```
GuySartorelli: Why is this being deprecated? The request **is** being used in this method.

src/Controllers/ElementalAreaController.php
```diff
@@ -45,132 +47,158 @@ public function getClientConfig()
...
-        if (!$request) {
-            $this->jsonError(400);
-            return null;
-        }
-        $id = $request->param('ID');
+        $request = $request ?: $this->getRequest();
+        $id = $request->param('ItemID');
+
+        // Respect previous behaviour
+        // This should just be removed later and so that it 404's below when BaseElement::get()->byID($id) fails
```
GuySartorelli, line 71: Please remove this comment. `400` is correct here - it's a bad request, as it doesn't include some of the necessary data. 404 is "you passed in all the necessary data, but we can't find a block with that ID" while 400 is "you didn't even pass in an ID for us to check, this request is bad"

src/Controllers/ElementalAreaController.php
```diff
@@ -45,132 +47,158 @@ public function getClientConfig()
...
     public function elementForm(HTTPRequest $request = null)
     {
-        // Get ID either from posted back value, or url parameter
-        if (!$request) {
-            $this->jsonError(400);
-            return null;
-        }
-        $id = $request->param('ID');
+        $request = $request ?: $this->getRequest();
+        $id = $request->param('ItemID');
```
GuySartorelli: Why did you change `ID` to `ItemID`?

src/Controllers/ElementalAreaController.php
```diff
@@ -45,132 +47,158 @@ public function getClientConfig()
...
      * @return Form|null Returns null if no element exists for the given ID
      */
     public function getElementForm($elementID)
     {
-        $scaffolder = Injector::inst()->get(EditFormFactory::class);
         $element = BaseElement::get()->byID($elementID);
-
+        // This is returning null to match previously behaviour
+        // in future major versions this should be deleted and rely on the 404 in elementForm()
+        // though do note them method is also called by formAction()
```
GuySartorelli:
```suggestion
```
We don't track todo in comments - please create an issue if you think we need to do this in a future major release.

src/Controllers/ElementalAreaController.php
```diff
@@ -45,132 +47,158 @@ public function getClientConfig()
...
+    public function save(array $data, Form $form): HTTPResponse
     {
-        // Validate required input data
-        if (!isset($this->urlParams['ID'])) {
+        $request = $this->getRequest();
+
+        // Check security token for non-view operation
+        if (!SecurityToken::inst()->checkRequest($request)) {
             $this->jsonError(400);
-            return null;
```
GuySartorelli, line 121: I personally like the return here even though it's dead code, since it makes it really clear at a glance "we don't continue executing any code after this".
But since it is *technically* correct to remove it I won't ask you to put it back.

src/Controllers/ElementalAreaController.php
```diff
@@ -45,132 +47,158 @@ public function getClientConfig()
...
         }
 
-        $data = json_decode($request->getBody(), true);
+        // We can assume a valid $id has been passed because it already been validated in elementForm()
+        $id = (int) $request->param('ID');
+        // /** @var BaseElement $element */
```
GuySartorelli:
```suggestion
```

src/Controllers/ElementalAreaController.php
```diff
@@ -45,132 +47,158 @@ public function getClientConfig()
...
+
+        // Check security token for non-view operation
+        if (!SecurityToken::inst()->checkRequest($request)) {
             $this->jsonError(400);
-            return null;
         }
 
-        $data = json_decode($request->getBody(), true);
+        // We can assume a valid $id has been passed because it already been validated in elementForm()
+        $id = (int) $request->param('ID');
```
GuySartorelli, line 119: Wouldn't hurt to do some defensive programming here, but I won't hold it against you if you leave it as is

src/Controllers/ElementalAreaController.php
```diff
@@ -45,132 +47,158 @@ public function getClientConfig()
...
+        // Note that canView() permissions were already checked in elementForm()
+        if (!$element->canEdit()) {
+            $this->jsonError(403);
+        }
+
+        // Ensure data has being POSTed
         if (empty($data)) {
             $this->jsonError(400);
-            return null;
         }
```
GuySartorelli, line 125: Do this before permissions checks - some permissions checks can be pretty unperformant so if we can exit early we should.

src/Controllers/ElementalAreaController.php
```diff
@@ -45,132 +47,158 @@ public function getClientConfig()
...
 
-        // Inject request body as request vars
-        foreach ($data as $key => $value) {
-            $request->offsetSet($key, $value);
+        // Ensure that ID is not passed in the data
+        if (isset($data['ID'])) {
+            $this->jsonError(400);
         }
```
GuySartorelli: Same with this - do it before permission checks

src/Controllers/ElementalAreaController.php
```diff
@@ -195,25 +228,123 @@ public function formAction(HTTPRequest $request)
...
+    ): ValidationResult {
+
+        // Remove the pseudo namespaces that were added by EditFormFactory
+        $dataWithoutNamespaces = $this->removeNamespacesFromFields($data, $element->ID);
+
+        // Create a seperate form because we're going to modify it
+        // if we simply updated $form it will mess up the FormSchema response, even if we `clone $form`
+        $formWithoutNamespaces = $this->createElementForm($element);
+
+        // Remove pseudo namespaces from fields so that something like RequiredFields('Title') works
```
GuySartorelli: What are "pseudo namespaces"?
Why are we worrying about form validators like `RequiredFields` here? This method is for model (and not form) level validation isn't it?

It looks like this method is only called from `save()` which is also the method that calls `write()` on the elements, so seems like that's the place to do it?

src/Controllers/ElementalAreaController.php
```diff
@@ -195,25 +228,123 @@ public function formAction(HTTPRequest $request)
...
+            return $this->getSchemaResponse($schemaId, $form, $errors);
+        });
+
+        return $form;
+    }
+
+    /**
+     * Call model validation taking into account the namespaced fields
+     */
+    private function getElementValidationResult(
```
GuySartorelli: This method feels like an anti-pattern. The model `validate()` method is called on `write()` and a `ValidationException` thrown if there are problems. Shouldn't we just let that process happen, and catch the exception and handle the validation errors from that point?

src/Controllers/ElementalAreaController.php
```diff
@@ -45,132 +47,158 @@ public function getClientConfig()
...
-                // Track changes so we can return to the client
-                $updated = true;
+        // Create and send FormSchema JSON response
+        $schemaID = $form->FormAction();
+        $response = $this->getSchemaResponse($schemaID, $form, $validationResult);
+        return $response;
+    }
+
+    /**
+     * Remove the pseudo namespaces that were added to form fields by the form factory
```
GuySartorelli: What is a "pseudo namespace"? Is that terminology already used in this module to refer to them?

src/Controllers/ElementalAreaController.php
GuySartorelli: Throwing an exception here is effectively the same as removing the method. That said, I don't think there's any valid use case for sending manual POST requests for saving blocks. We're probably safe doing this - though it's stretching semver a little.

src/Forms/EditFormFactory.php
```diff
@@ -55,6 +57,33 @@ protected function getFormFields(RequestHandler $controller = null, $name, $cont
...
+    {
+        $compositeValidator = parent::getFormValidator($controller, $name, $context);
+        if (!$compositeValidator) {
+            return null;
+        }
+        $id = $context['Record']->ID;
+        foreach ($compositeValidator->getValidators() as $validator) {
+            if (!is_a($validator, RequiredFields::class)) {
+                continue;
+            }
```
GuySartorelli:
```suggestion
        foreach ($compositeValidator->getValidatorsByType(RequiredFields::class) as $validator) {
```

src/Forms/EditFormFactory.php
```diff
@@ -55,6 +57,33 @@ protected function getFormFields(RequestHandler $controller = null, $name, $cont
         return $fields;
     }
 
+    protected function getFormValidator(RequestHandler $controller = null, $name, $context = [])
```
GuySartorelli, line 61: If we remove this method, the validation will still work, right? It'll just skip the JS validation and validate it on the backend only?

Seems like that would mean:
1. less complexity
2. better UX when there are validation errors in other form validators
3. only slightly worse (slower) UX when there are only required fields validation errors

If my assumptions above are correct, IMO we should remove this method.

GuySartorelli (changes requested):
Haven't looked at tests yet.

src/Controllers/ElementalAreaController.php
emteknetnz: 404 is correct here because ItemID is in the URL. If it was post data then I agree that 400 is correct.

client/src/components/ElementEditor/Element.js
emteknetnz: Due to legacy reasons, namely the graphql HOC on the PublishAction.js publish button, we need to do the publishing on the button, not within Element.js.  This state is passed as a prop to PublishAction.js via ElementContext.

If I was starting from scratch I'd just do everything in Element.js

Getting rid of GraphQL from elemental would require creating publish + duplicate + archive controller endpoints to replicate what GraphQL is already doing. It's increasing the scope of this card too much IMO

src/Controllers/ElementalAreaController.php
emteknetnz: Leave as is

src/Forms/EditFormFactory.php
emteknetnz: We shouldn't be removing JS validation, it's an existing feature

src/Controllers/ElementalAreaController.php
emteknetnz: I think ID used to work with the old routing because it's a one of the default params from `/$Action/$ID/$OtherID`

$ItemID is correct for both of the code paths mentioned in the docblock comment

src/Controllers/ElementalAreaController.php
emteknetnz: Updated

src/Controllers/ElementalAreaController.php
emteknetnz: Updated

src/Controllers/ElementalAreaController.php
emteknetnz: Updated

src/Controllers/ElementalAreaController.php
emteknetnz: Changed to 'namespace prefixes'

client/src/components/ElementEditor/Element.js
emteknetnz: Have added comment

client/src/components/ElementEditor/Element.js
emteknetnz: Have added comment on https://github.com/silverstripe/silverstripe-elemental/pull/1150/files#r1527640688

tl;dr it's all because we're doing a remote submit using redux-form which doesn't return a promise

src/Controllers/ElementalAreaController.php
emteknetnz: Have removed usable to $request in the method, just using $this->getRequest() only now

I think it used to be relevant with some legacy code, now it's no longer relevant

src/Controllers/ElementalAreaController.php
emteknetnz: Have removed, code is removed in a follow up [CMS 6 PR](https://github.com/creative-commoners/silverstripe-elemental/pull/2)

src/Controllers/ElementalAreaController.php
emteknetnz: Have updated to 'namespace prefixes'

Have updated comment to simply be `Remove namespace prefixes from fields so that validation works` - it's needed for model level validation I believe.

`save()` seems like a bad place to be worrying about removing namespaces, currently it's doing checks that return HTTP status codes and calling this validation method and write(), it's all kind of high level.  Seems far better to have the namespace stuff neatly tucked away in this private method

src/Controllers/ElementalAreaController.php
emteknetnz: See the docblock comment `Call model validation taking into account the namespaced fields`

It's all about the namespace prefixes

src/Forms/EditFormFactory.php
emteknetnz: Updated

client/src/components/ElementEditor/InlineEditForm.scss
emteknetnz: > or use some other mechanism to make sure the screen reader doesn't start reading content inside here

Have gone this route

Have added dynamic [aria-hidden](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden) attribute via Content.js + InlineEditForm.js

emteknetnz:
Have updated CI run linked on main issue description to only include behat for elemental

client/src/components/ElementEditor/Element.js
emteknetnz: We now do a remote-submit of redux-form which doesn't return a promise. Because of this we can't respond for a 400+ http status code which is what the old catch() block used to do when doing a fetch via lib/Backend.  What this means is there's no way for us to see if the request failed

The user experience is bad, there's none, all you get is a failed request in the network panel of dev tools.  However it doesn't really matter IMO because to get the 400+ status codes you pretty much need to be doing something that you shouldn't be doing i.e. crafting custom XHR requests

If you'd like to improve this UX then it should been done as a seperate card because we'd be updating it in silverstripe-admin on one of the FormBuilder javascript files. IMO I don't think this is particularly worthwhile

client/src/components/ElementEditor/Element.js
emteknetnz: No that's inccorect.

Instead we're saving the form first via this.props.submitForm(). That will trigger validation errors via the whole FormSubmission thing server side.  We need to do this before we publish.

The publish button just hits a generic GraphQL endpoint that publishes a dataobject

We're setting justClickedPublishButton: true before this. What this will do is set 

```js
    if (this.state.justClickedPublishButton) {
      this.setState({
        justClickedPublishButton: false,
        doPublishElement: true,
      });
    }
```

In `handleFormSchemaSubmitResponse()` after a save submission without validation errors has occured

client/src/components/ElementEditor/Element.js
GuySartorelli: I challenge the argument that it would be a separate card - this PR is breaking that functionality, so if we want to keep it we should do it as part of this card.

The one question I have is: If we **could** detect that failed request, would it include failures as a result of validation errors?
If so, we need to fix it so we get toasts for validation errors.
If not, I'm happy for that to be left as is.

client/src/components/ElementEditor/Element.js
GuySartorelli: That doesn't explain why the toast can't be here? Is this not after the save event has occurred? The publish toast is in `handleAfterPublish` so I'm confused as to why they look like they should be the same but aren't. Is the publish flow somehow completely different than the save flow?

client/src/components/ElementEditor/Element.js
GuySartorelli: This is really confusing to me. We set `justClickedPublishButton` to true, which then gets set to false so we can set `doPublishElement` to true, which then goes and does another thing.... this whole flow of saving and publishing feels really convoluted.

It's clear from discussions with you that this is done to work around limitations in the existing code.... but I think we need a writeup somewhere that says "when you click save, this is the flow of events that happens" and  "when you click publish, this is the flow of events that happens" - otherwise we'll have to unpick all of this any time we need to do maintenance in this area.

It's also really hard to review this code without something that tells me how it's all wired together - it's too hard to trace it all.

client/src/components/ElementEditor/Element.js
GuySartorelli: Don't bother answering that - instead just refer to https://github.com/silverstripe/silverstripe-elemental/pull/1150#discussion_r1531092733.

src/Controllers/ElementalAreaController.php
GuySartorelli: This method is a URL handler as per `$url_handlers` - therefore it must take and use an `HTTPRequest` argument.

I think it needs to keep being used, and not be deprecated unless we're removing it from `$url_handlers`.

src/Controllers/ElementalAreaController.php
GuySartorelli: If `$id` is **always**  in the URL and there's no way to get here without one, the condition here is misleading at best, since the only time it's hit is when `$id` is `0` explicitly, not when it's missing which is what the condition implies.

If that's all correct, please modify the condition to test explicitly for `$id === 0` and either return a 400 (because an ID of 0 is always going to be bad data) or a 404 (because I feel like you won't agree about the 400 :p)

src/Controllers/ElementalAreaController.php
```diff
@@ -195,25 +224,123 @@ public function formAction(HTTPRequest $request)
...
+     * Call model validation taking into account the namespaced fields
+     */
+    private function getElementValidationResult(
+        Form $form,
+        array $data,
+        BaseElement $element
+    ): ValidationResult {
+
+        // Remove the namespace prefixes that were added by EditFormFactory
+        $dataWithoutNamespaces = $this->removeNamespacesFromFields($data, $element->ID);
```
GuySartorelli:
```suggestion
        $dataWithoutNamespaces = static::removeNamespacesFromFields($data, $element->ID);
```
Static methods should be called statically.

src/Controllers/ElementalAreaController.php
GuySartorelli: > `save()` seems like a bad place to be worrying about removing namespaces, currently it's doing checks that return HTTP status codes and calling this validation method and write(), it's all kind of high level. Seems far better to have the namespace stuff neatly tucked away in this private method

Sorry, the comment about `save()` was meant to be about `getElementValidationResult()`, not about `removeNamespacesFromFields()`.

In the `save()` method, we call `write()`. This results in calling `validate()` and throws a `ValidationException` if there are errors.

Seems to me we should just catch the `ValidationException` at **that** point, handle any namespace changes that are needed in the validation result (yes, by calling a separate method from `save()`), and throw the `ValidationException` again from there.

src/Controllers/ElementalAreaController.php
GuySartorelli: Discussion about this can be continued in https://github.com/silverstripe/silverstripe-elemental/pull/1150#discussion_r1531110573 which ended up being the same thing.

src/Forms/EditFormFactory.php
GuySartorelli: No validation existing in elemental blocks before this, so javascript validation in elemental blocks was not an existing feature - and this is adding extra code to get it working. Which means as I stated above:
1. more complexity
2. worse UX in some scenarios
3. Only avoids a single fast XHR request

I can see you won't budge on this though so it will be part of the card I mentioned in https://github.com/silverstripe/silverstripe-elemental/pull/1150#issuecomment-2002616999

GuySartorelli:
Sorry, I forgot to hit "start review"
There are a few responses to existing items above - I've marked everything as resolved which has been resolved.

The main point is https://github.com/silverstripe/silverstripe-elemental/pull/1150#discussion_r1531092733
It's very hard to follow what's happening when saving/publishing, which is making this PR very hard to review. If it's not possible to make those flows more straight-forward, we need something documented that explains what's going on.

GuySartorelli:
NOTE:
This review cycle was just looking at stuff we had already discussed and continuing those discussions - I didn't look at any of the **new** code e.g. the workaround for the weird publishing timing issue.

emteknetnz:
I spend some time writing docs to describe the flow ... but then I stopped because I'm just literally just describing what the code does here which I think are redundant comments that can easily get out of sync with the main code.  Good comments IMO are just explaining why weird looking code is the way it is.

Things are convoluted for several reasons:
- There's an existng GraphQL HOC readBlocksForAreaQuery around ElementList which fetches data for the entire ElementalArea
- There's an existing GraphQL HOC on PublishAction for a mutation to publish the block. After publish the mutation will call readBlocksForAreaQuery which will refresh the entire ElementalArea
- SaveAction and PublishAction are added via registerTransforms.js, so we need to talk to them via the newly added <ElementContext>
- SaveAction used to just do a regular XHR fetch(), now it triggers a remote submit on redux-form which powers FormBuilder. However there isn't a promise returned when using remote submit
- We use remote-submit because the injector added save (submit) button is not actually on the FormBuilder form - 
- Blocks can be inline saved and published when the block is closed and the FormBuilder form has not rendered. We need the form to be rendered in order to do a remote submit.
- In some scenarios redux-form will set the form to dirty on initialization, specifically during a behat test when blocks aren't published and you inline publish a block without without opening the block. Hence the use of the setTimeout() workaround to set the form state to pristine.

Props are passed between Element.js and PublishButton.js via ElementContext which is defined on Element

Layout of components, hoc's and context
```
<readBlocksForAreaQuery> (added via registerTransforms.js)
  <ElementList>
    <Element>
      <ElementContext>
        <Content>
          <InlineEditForm>
            <FormBuilderLoader>
        // ...
        <SaveAction> (the save button in the context menu)
        <publishBlockMutation> (added via compose() on PublishAction)
            <PublishAction> (the publish button)
```
                
The flow for a the publish button

1. User clicks publish button which is handled by PublishButton::handleClick()
2. This calls PublishButton.props.onPublishButtonClick()
3. This is a reference of Element.handlePublishButtonClick() 
4. This calls Element.setState({justClickedPublishButton: true, ensureFormRendered: true});

Element.state.ensureFormRendered flow:
- Used as a conditional in Content to ensure the InlineEditForm and child FormBuilderLoader are rendered
- There's a callback Element.handleFormInit(), which is all the way to FormBuilderLoader.js, which is triggered when the redux-form has rendered
- This calls Element.setState({formHasRendered: true});

Element.state.justClickedPublishButton flow:
- Element.componentDidUpdate() checks for (this.state.justClickedPublishButton && this.state.formHasRendered)), if true ...
- if this.props.formDirty, then remote-submit, and call Element.setState({doPublishElementAfterSave: true}
- if not this.props.formDirty, call Element.setState({doPublishElement: true})

remote-submit flow:
- ...

(I stopped writing here, I'd just end up replicating all of the code)

So ... yeah it's complicated though I don't know what to do about that.  I've spent way too long getting this fragile thing to work and I'm not going to fundamentally change anything at this point.  If we want to simplyfy this solution I think it requires to removing/replacing things like GraphQL and redux-form, which is too much scope at this point. We're just trying to get validation working.

Would it work if we just limited the peer review to "test it to make sure it works and check that there's nothing absolutely crazy" rather than trying to write/describe easy to read code?

GuySartorelli:
> Would it work if we just limited the peer review to "test it to make sure it works and check that there's nothing absolutely crazy" rather than trying to write/describe easy to read code?

The problem with that is the way state is being used to trigger things to happen (as opposed to literally just "this is the current state") **is** absolutely crazy to my mind... Maybe lets discuss in person on the whiteboard when you're in the office next so I can wrap my head around this.

GuySartorelli:
In the meantime there are still a few requested changes that can be made/responded to:
- https://github.com/silverstripe/silverstripe-elemental/pull/1150#discussion_r1531081672
- https://github.com/silverstripe/silverstripe-elemental/pull/1150#discussion_r1531096166
- https://github.com/silverstripe/silverstripe-elemental/pull/1150#discussion_r1531099328
- https://github.com/silverstripe/silverstripe-elemental/pull/1150#discussion_r1531110573
- https://github.com/silverstripe/silverstripe-elemental/pull/1150#discussion_r1531114758 (just informational - no change to make here unless you want to be kind to my brain :p)
- https://github.com/silverstripe/silverstripe-elemental/pull/1150#discussion_r1531103858

client/src/components/ElementEditor/Element.js
emteknetnz: >  If we could detect that failed request, would it include failures as a result of validation errors?

No, the old toast triggered by the catch() block was for 400-500 http status code failures such as failed permission check (403) (lib/Backend adds some logic so that you can use a catch block for 400-500), not validation errors which come back in a 200. That's possible incorrect (maybe it should be 400), though linkfield is also returning 200 on validation error so should stick with that.

src/Controllers/ElementalAreaController.php
emteknetnz: Oh :-) Updated

src/Controllers/ElementalAreaController.php
emteknetnz: Changed to `$id === 0`

src/Controllers/ElementalAreaController.php
emteknetnz: Updated

I've also just removed some unused code from that method that looked like it was originally doing Form validation, which is unnecessary. I think with was copied from the original spike

src/Forms/EditFormFactory.php
emteknetnz: JS validation is part of FormBuilder/FormSchema, it's not specific to elemental.

For instance try submitting a malformed email address in LinkField v4 email field - there's no XHR request because the validation happens in JS

It's pretty inconsistent which fields have it, for instance the new UrlField (external link field) has no JS validation and will get validated on the server - you'll see an XHR request for that one

src/Controllers/ElementalAreaController.php
emteknetnz: Updated

src/Forms/EditFormFactory.php
GuySartorelli: That's fine, like I said, I can see we won't agree about it so I'll leave it alone in this PR and include it in a separate card.

emteknetnz:
I had a go at removing the "state to control things" logic and only use callbacks, though I ended up reverting most of it because it wasn't making things any easier IMO.

The main area right now where there's state which could be removed in the SaveAction.js, where numerious things are passed in via context which could just be used from the Element.js.  However if we do this then the Save button now doesn't match what's happening in the publish button nor the duplicate button, where the 'publish' and 'duplicate' logic (aka graphql) are contained in their respective js files.  I think it's better to be consistent and keep the buttons roughtly similar.

As to why we're using state as we are, it really comes down to dependencies between doing things i.e. we need to confirm that previous things have completed before proceeding.

For instance, when I click the publish button, I need the certain things to be true:
- The form must have rendered (if you click publish on a block that hasn't been rendered, the form won't be there)
- If the form is dirty, I must submit (save) the form before publishing. If it's pristine, publish straight away.
- If saving, there needs to be no validation errors
- Only now I can publish the block.

So the `justClickedPublishButton` state means the user "has just clicked the publish button", though we still need to go through all the other steps above before we actually hit the publish endpoint.  We use different state variables to denote where we are in the processes.

It's worth nothing that we're also heavily using callbacks which is the normal way to do "do this, then do that".  However I don't think in the case it's viable to only use callbacks because there's a bunch of "if then else" conditions in the flow i.e. has the form rendered yes or no? is the formDirty yes or no?, did validation pass or fail?

client/src/components/ElementEditor/Element.js
```diff
@@ -53,6 +74,72 @@ class Element extends Component {
...
+          doPublishElement: true,
+        });
+      }
+    }
+    if (this.state.doDispatchAddFormChanged) {
+      this.setState({
+        doDispatchAddFormChanged: false,
+      });
+      this.props.dispatchAddFormChanged();
+    }
```
GuySartorelli:
```suggestion
    const newState = {};
    if (this.state.justClickedPublishButton && this.state.formHasRendered) {
      newState.justClickedPublishButton = false;
      if (this.props.formDirty) {
        // Save the element first before publishing, which may trigger validation errors
        this.props.submitForm();
        newState.doPublishElementAfterSave = true;
      } else {
        // Just publish the element straight away without saving first
        newState.doPublishElement = true;
      }
    }
    if (this.state.doDispatchAddFormChanged) {
      newState.doDispatchAddFormChanged = false;
      this.props.dispatchAddFormChanged();
    }
    this.setState(newState);
```
We only need to set state once here (unless those function calls require the new state - but I think the actual **setting** of the state happens in the next event cycle anyway?)

client/src/components/ElementEditor/Element.js
```diff
@@ -215,6 +302,102 @@ class Element extends Component {
...
+    if (this.state.doPublishElementAfterSave) {
+      this.setState({
+        doPublishElementAfterSave: false,
+        doPublishElement: true,
+      });
+    }
+    if (!this.state.doPublishElement && !this.state.doPublishElementAfterSave) {
+      this.showSavedElementToast(title);
+    }
+    this.refetchElementalArea();
```
GuySartorelli:
```suggestion
    const newState = {};
    if (hasValidationErrors) {
      if (this.props.type.inlineEditable) {
        newState.previewExpanded = true;
      }
      // Ensure that formDirty remains truthy
      // Note we need to call this.props.dispatchAddFormChanged() on the next render rather than straight away
      // or it will get unset by code somewhere else, probably redux-form
      newState.doDispatchAddFormChanged = true;
      // Don't accidentally auto publish the element once validation errors are fixed
      if (this.state.doPublishElementAfterSave) {
        newState.doPublishElementAfterSave = false;
      }
      this.setState(newState);
      return;
    }
    // Form is valid
    newState.newTitle = title;
    if (this.state.doPublishElementAfterSave) {
      newState.doPublishElementAfterSave = false;
      newState.doPublishElement = true;
    }
    if (!this.state.doPublishElement && !this.state.doPublishElementAfterSave) {
      this.showSavedElementToast(title);
    }
    this.setState(newState);
    this.refetchElementalArea();
```
As above, I suspect we don't need to be making so many `setState()` calls.

src/Controllers/ElementalAreaController.php
```diff
@@ -45,132 +47,153 @@ public function getClientConfig()
...
      */
     public function elementForm(HTTPRequest $request = null)
     {
-        // Get ID either from posted back value, or url parameter
-        if (!$request) {
+        $request = $request ?: $this->getRequest();
+        $id = $request->param('ItemID');
+
+        // Respect previous behaviour
+        // This should just be removed later and so that it 404's below when BaseElement::get()->byID($id) fails
```
GuySartorelli, line 71:
```suggestion
```
This is no longer valid due to the decision in https://github.com/silverstripe/silverstripe-elemental/pull/1150#discussion_r1531453680

GuySartorelli (dismissed):
## The publish flow
Please move `publishElement()` into the `Element.js` file, or move the whole publish flow into `PublishAction.js`.

Either clicking the publish button should hand over responsibility to `Element.js` via the `onPublishButtonClick()` callback, OR the publish action should be responsible for publishing the block.
It is much too confusing for the publish button to handle control to the element via a callback, only for control to be passed back to it through a context variable - which then *passes control back to the element via another callback*. There's too much back-and-forth going on there.

## The save flow
Same deal here - we should only pass control one way. So either after the click we pass full control to `Element.js`, or we handle all of the save logic and **then** pass control to `Element.js` for any after-save actions.

Passing control back and forth so many times (especially since there's a combination of context, callbacks, and state all involved in controlling where the flow goes) makes it hard to see what's happening which makes review hard but it'll also make maintenance hard.

emteknetnz:
PublishAction uses GraphQL as a weird HOC defined in publishBlockMutation.js which includes `handlePublishBlock`.  Within PublishAction you have access to the prop `handlePublishBlock` via some sort of magic.  I don't know how you're supposed to import publishBlockMutation and get access to `handlePublishBlock` without wrapping all of Element.js in the mutation which I'm not going to do

If I `import publishBlockMutation from 'state/editor/publishBlockMutation';` and then `console.log(publishBlockMutation);` I get `function b(r)` if I `console.log(publishBlockMutation());` I get `e is undefined` and a wall of exception from vendor.js

Even I was able to move the GraphQL query then there's also a secondary concern that even if we did move this, then it's now inconsistent with ArchiveAction.js, UnpublishAction.js and DuplicationAction.js which all follow the GraphQL HOC pattern, so we'd be adding more ways to do the same thing.

Because moving the logic for the publish button doesn't seem feasible, I'd like to keep save button logic the same as well so that SaveAction and PublishAction are consistent

client/src/components/ElementEditor/Element.js
emteknetnz: Updated

client/src/components/ElementEditor/Element.js
emteknetnz: Updated

src/Controllers/ElementalAreaController.php
emteknetnz: I think the comment still makes sense. The logic to return a 400 will be removed in PR to CMS 6 so it returns a 404 in this scenario rather than a 400.

PR is [https://github.com/creative-commoners/silverstripe-elemental/pull/2/files#diff-942115e4b8f6030e4d72ebc2b3a0772ec65e5dfcd08fbd0e677c70d1231daf28L74](here) - note it needs a rebase, the branch it's from is very out of date

GuySartorelli:
I think I need a second opinion on this before I okay it - it feels like the flow is too convoluted and will cause trouble for us down the line, which I've mentioned in more detail in comments above.
@maxime-rainville @silverstripe/core-team does anyone else have opinions about this?

emteknetnz:
In terms of refactoring to make this less convoluted, as mentioned earlier the bottle neck is really the use of the GraphQL HOC on the publish button.  So in order to refactor I can think of a couple of options:

a) have way to use apollo/graphql so you can do an "inline" query rather than as a HOC (I don't know how to do this) - then we could use the mutation from Element.js
b) A possibly easier option (easier in that we don't have 2 different ways to do graphql on top of all the other multiple ways of doing things), is to just not use GraphQL and create a little publish endpoint on ElementalAreaController and send a basic fetch() request from Element.js. That's fairly trivial to implement.

GuySartorelli:
> Another possibly easier option (so we don't have 2 different ways to do graphql on top of all the other multiple ways of doing things), is to just not bother with graphql and create a little publish endpoint on ElementalAreaController and send a basic fetch() request from Element.js. That's fairly trivial.

That sounds preferable to me.
Bonus points (but not a requirement) if it's something that can be handled via formschema so we can reuse that scenario for any formschema-based forms which have versioned records going forward.

emteknetnz:
Publishing has nothing to do with formschema.  This would just be method on a controller that calls `$element->publishRecursive();` and some others things like permission checks and returned the correct https status code

I did a [POC](https://github.com/emteknetnz/silverstripe-elemental/pull/4/files#diff-942115e4b8f6030e4d72ebc2b3a0772ec65e5dfcd08fbd0e677c70d1231daf28R155) a while back of using regular controllers in elemental so I'd largely just copy that

client/src/components/ElementEditor/Element.js
```diff
@@ -1,20 +1,25 @@
...
 import { loadElementFormStateName } from 'state/editor/loadElementFormStateName';
 import { loadElementSchemaValue } from 'state/editor/loadElementSchemaValue';
 import * as TabsActions from 'state/tabs/TabsActions';
 import { DragSource, DropTarget } from 'react-dnd';
 import { getEmptyImage } from 'react-dnd-html5-backend';
 import { elementDragSource, isOverTop } from 'lib/dragHelpers';
+import * as toastsActions from 'state/toasts/ToastsActions';
+import { addFormChanged, removeFormChanged } from 'state/unsavedForms/UnsavedFormsActions';
+
+export const ElementContext = createContext(null);
```
maxime-rainville, line 22: This probably should be in its own file.

client/src/components/ElementEditor/Element.js
```diff
@@ -248,6 +423,19 @@ class Element extends Component {
       this.getVersionedStateClassName()
     );
 
+    // eslint-disable-next-line react/jsx-no-constructed-context-values
+    const providerValue = {
```
maxime-rainville, line 427: I think this is abusing the `useContext` hook. Context normally passes a simple value down the virtual dom. This is passing down values and a bunch of handlers to update things.

Seems like the component is already heavily using redux anyway, so layering context on top of it seems to be adding complexity.

If we're not going to put this in redux, then we should minimally be extremely clear about what each entry will is doing by putting the context in a dedicated file with clear documentation of what each key is doing.

client/src/components/ElementEditor/Element.js
maxime-rainville, line 1: We're already rewriting most of the component ... seems like this would have been a good opportunity to refactor has a functional component.

client/src/components/ElementEditor/Element.js
```diff
@@ -53,6 +74,68 @@ class Element extends Component {
...
+    // Check if formSchema state has already been loaded before opening a block
+    // This can happen if there was a validation error on a block after performing a Page save
+    if (this.props.formStateExists) {
+      this.setState({
+        formHasRendered: true
+      });
+    }
+  }
+
+  componentDidUpdate() {
```
maxime-rainville, line 86: Triggering HTTP requests by updating the state and picking up the state update in `componentDidUpdate` seems incredibly weird and questionable.

client/src/components/ElementEditor/InlineEditForm.scss
```diff
@@ -25,4 +25,25 @@
       max-width: 900px;
     }
   }
+
+  // Render the form off the screen so that it's not visible
+  // Note that using display:none will mean the form is not rendered at all and cannot be submitted
+  &--rendered-not-visible {
```
maxime-rainville, line 31: Why are we rendering the form off the screen?

Won't this cause accessibility problems?

client/src/components/ElementActions/PublishAction.js
```diff
@@ -132,6 +46,19 @@ const PublishAction = (MenuComponent) => (props) => {
     toggle: props.toggle,
   };
 
+  useEffect(() => {
+    if (formHasRendered && doPublishElement) {
```
maxime-rainville, line 50: We are triggering an action based on a state update we are getting from somewhere else. This seems very bad.

client/src/components/ElementEditor/Element.js
```diff
@@ -215,6 +298,96 @@ class Element extends Component {
...
+      this.showSavedElementToast(title);
+    }
+    this.refetchElementalArea();
+    this.setState(newState);
+  }
+
+  refetchElementalArea() {
+    // This will trigger a graphql readOneElementalArea request that will cause this
+    // element to re-render including any updated title and versioned badge
+    window.ss.apolloClient.queryManager.reFetchObservableQueries();
```
maxime-rainville, line 388: Surely there's a better way to refresh a single query than force every other query to re-run?

maxime-rainville (commented):
To the extent that all the other actions on the block (archive, duplicate, unpublish, etc) are still going through GraphQL, I would question the wisdom of moving publication off GraphQL.

I do think the way we write GraphQL request is dumb and unduly complicated. I would be in favour of coming up with a better standard for creating those GraphQL requests.

I think the terraformers have done some customisation to an important project in that space. Might be worth validating with them if that's going to bit them in the ass.

michalkleiner:
> To the extent that all the other actions on the block (archive, duplicate, unpublish, etc) are still going through GraphQL, I would question the wisdom of moving publication off GraphQL.

☝️  

> I do think the way we write GraphQL request is dumb and unduly complicated. I would be in favour of coming up with a better standard for creating those GraphQL requests.

👍

client/src/components/ElementEditor/Element.js
GuySartorelli: Considering how long we've been working on this I think that should be explicitly out of scope.

client/src/components/ElementEditor/InlineEditForm.scss
GuySartorelli: There's an `aria-hidden` (or whatever the actual thing is) to avoid this causing problems for screen-reader users.
It needs to be rendered so the form can submit with the correct values. Otherwise, the formschema form submission has no values to submit.

It's an unfortunate quirk of using formschema in an element that can be collapsed.

emteknetnz:
Based on Max's reponse it seems like the plan is basically

1) Redo the graphql publish query in a non-hoc fashion
2) Move save + publish logic to Element.js
3) Possibly move ElementContext to its own file (though not sure there will be much left after step 2)

Like has Guy has already replied, I'm not a fan of "refactor this to become a functional component" alongside an existing complex PR because that increase in scope makes things FAR harder to peer review.

However to achieve 1) it's a requirement because apollo now uses hooks - https://www.apollographql.com/docs/react/data/mutations

The existing apollo hoc method is deprecated https://www.apollographql.com/docs/react/api/react/hoc/

Given that this PR has been open and worked on for an extremely long time, can I propose that we just merge this as is and then open a new card to refactor this to a functional component and implement 1-3?

client/src/components/ElementEditor/Element.js
emteknetnz: It's the only way I could work out how to do this.

Essentially we need to ensure that several dependent occurred in a sequence sending a publish request, e.g. when publishing
- The form has rendered before sending a request so that validation can occur - it won't have rendered if the element has been toggled open
- If the form is dirty, save the form first to trigger validation. If it's valid then publish.

client/src/components/ElementEditor/InlineEditForm.scss
maxime-rainville: That's good context. This seems like a very important point that should be document in the component using this class rather than the style sheet.

client/src/components/ElementEditor/Element.js
maxime-rainville: Have we tried using Promises?

The way I would approach this is:
1. Define a `waitForFormRender` method. The method would return a Promise.
  * If the form as already rendered, then `Promise.resolve` is return.
  * If the form as not resolved a new promise is returned that will be resolved when `handleFormInit` is called.
2. Update `handlePublishButtonClick` to call `waitForFormRender` and trigger whatever request is needed when the promise resolves.

client/src/components/ElementEditor/Element.js
```diff
@@ -215,6 +298,96 @@ class Element extends Component {
...
+    });
+    // Ensure that formDirty becomes falsey after publishing
+    // We need to call at a later render rather than straight away or redux-form may override this
+    // and set the form state to dirty under certain conditions
+    // setTimeout is a hackish way to do this, though I'm not sure how else we can do this
+    // The core issue is that redux-form will detect changes when a form is hydrated for the first
+    // time under certain conditions, specifically during a behat test when trying to publish a closed
+    // block when presumably the apollo cache is empty (or something like that). This happens late and
+    // there are no hooks/callbacks available after this happens the input onchange handlers are fired
+    setTimeout(() => {
```
maxime-rainville, line 335: If this is waiting for a HTTP request to finish, it's entirely possible it will take more than .5 sec to complete.

If we know what action in redux form is triggering this behaviour, we could add our own logic in our reducer to handle it and trigger our own follow up action.

maxime-rainville (commented):
> Given that this PR has been open and worked on for an extremely long time, can I propose that we just merge this as is and then open a new card to refactor this to a functional component and implement 1-3?

I would be keen to see the timeout, weird state management logic and the ElementContext solve in this card.

Given that GraphQL bits require coming up with new and improve way of doing GraphQL queries, I'm fine with putting that in a different card.

emteknetnz:
> I would be keen to see the timeout, weird state management logic and the ElementContext solve in this card.
> Given that GraphQL bits require coming up with new and improve way of doing GraphQL queries, I'm fine with putting that in a different card.

Most of the weird state management logic + ElementContext goes away if we resolve the GraphQL query.  Hence why I suggested merging this as is and opening a new card.  No point refactoring stuff as part of this peer review only to get rid of it almost straight away.

This card has been in peer review for over a month :-/ Are people OK with my suggestion to just merge this as (I believe it works functionally just fine as is) and do any additional refactoring work in a new card?

maxime-rainville:
I'm really not sure what the value of merging something that introduces a lot of bad construct is if we are going to immediately have a follow up card.

I'll just take all my grievances and convert them to ACs on the new card. You'll end up with something like:
- Simplify the GraphQL publication call.
- Remove abusive context usage on PublishAction.
- Don't trigger action on state change.
- Don't use `setTimeout` in Element.

emteknetnz:
I've created a follow up card for refactoring and put into refinement https://github.com/silverstripe/silverstripe-elemental/issues/1164

client/src/components/ElementActions/PublishAction.js
emteknetnz: Will be addressed in follow up card https://github.com/silverstripe/silverstripe-elemental/issues/1164

client/src/components/ElementEditor/Element.js
emteknetnz: Will be addressed in follow up card https://github.com/silverstripe/silverstripe-elemental/issues/1164

client/src/components/ElementEditor/Element.js
emteknetnz: Will be addressed in follow up card https://github.com/silverstripe/silverstripe-elemental/issues/1164

client/src/components/ElementEditor/Element.js
emteknetnz: Will be addressed in follow up card https://github.com/silverstripe/silverstripe-elemental/issues/1164

client/src/components/ElementEditor/Element.js
emteknetnz: Will be addressed in follow up card https://github.com/silverstripe/silverstripe-elemental/issues/1164

client/src/components/ElementEditor/Element.js
emteknetnz: Will be addressed in follow up card https://github.com/silverstripe/silverstripe-elemental/issues/1164

client/src/components/ElementEditor/Element.js
emteknetnz: Will be addressed in follow up card https://github.com/silverstripe/silverstripe-elemental/issues/1164

client/src/components/ElementEditor/InlineEditForm.scss
emteknetnz: Have put a note about this in Content.js which is where this css class is added

maxime-rainville (commented):
I tested the following scenarios:
1. Validation for not inline block :heavy_check_mark: 
2. Validation for errors thrown by the `validate` method on a block when manually saving a block  :heavy_check_mark: 
3. Validation for errors thrown by the `validate` method on a block when saving the page :warning:
4. Field level validation (putting a non-numeric value in a numeric field) :x:


Scenario 3 sort of works, but the validation message is attached to the page instead. I take it this is covered by https://github.com/silverstripe/silverstripe-elemental/issues/1155

Scenario 4 doesn't work either on block save or page save. It does work for non-inline-block. Are we splitting this up as well?

emteknetnz:
Yes 3 is will be covered by https://github.com/silverstripe/silverstripe-elemental/issues/1155

Re 4 - it's likely that the rendered `<input type="number">` is letting you enter non-numeric e.g. "abc" though it won't actually submit the value, either via inline save or page save e.g. network panel XHR request shows a blank value for the numeric field

Try using [UrlField](https://github.com/silverstripe/silverstripe-framework/blob/5/src/Forms/UrlField.php) instead - you should see that validation is working correctly for that.

Also for NumericField I do have this setup on my local, which does trigger for values for '1' on inline save:

```php
// app/src/MyNumbericField.php
use SilverStripe\Forms\NumericField;

class MyNumbericField extends NumericField
{
    private static $extensions = [
        MyNumericFieldExtension::class
    ];
}
```

```php
// app/src/MyNumericFieldExtension.php
use SilverStripe\Core\Extension;
use SilverStripe\Forms\NumericField;
use SilverStripe\Forms\Validator;

/**
 * @extends Extension<NumericField>
 */
class MyNumericFieldExtension extends Extension
{
    public function updateValidationResult($result, Validator $validator)
    {
        if ($this->owner->value == 1) {
            $validator->validationError($this->owner->getName(), 'This field cannot be 1');
        }
    }
}
```

```php
// app/src/MyBlock.php
// ...
{
    private static $db = [
        'MyInt' => 'Int',
    ];
    // ...
    {
        // ...        
        $fields->removeByName('MyInt');
        $fields->addFieldToTab('Root.Main', MyNumbericField::create('MyInt', 'My Int'));
    }
}
```

maxime-rainville:
Still doesn't work for me 

[validation-not-triggering-2024-04-09_17.59.22.webm](https://github.com/silverstripe/silverstripe-elemental/assets/1168676/b8fbbb85-e3d4-4a2b-ad4a-fa2723bf6f4c)

Here's the sample block I was using

```php
<?php

use SilverStripe\Forms\FieldList;
use SilverStripe\Forms\NumericField;
use SilverStripe\Forms\UrlField;

class ValidationBlock extends BaseElement
{
    private static $db = [
        'Integer' => 'Int',
        'Url' => 'Varchar(255)'
    ];

    private static $inline_editable = true;

    private static $singular_name = 'validation block';

    private static $plural_name = 'validation blocks';

    private static $icon = 'font-icon-block';

    private static $table_name = 'S_EB_ValidiationBlock';

    {
        $this->beforeUpdateCMSFields(function (FieldList $fields) {
            $fields->addFieldToTab(
                'Root.Main',
                NumericField::create('Integer'),
                UrlField::create('Url')
            );

        });

        return parent::getCMSFields();
    }

    public function getType()
    {
        return _t(__CLASS__ . '.BlockType', 'Validation');
    }

    public function getSummary()
    {
        return '';
    }

    public function validate()
    {
        $result = parent::validate();

        // This will add a field specific error to the ValidationResult
        if ($this->Integer > 10) {
            $result->addFieldError('Integer', 'Integer must be less than 11');
        }

        return $result;
    }
}

class FormValidationBlock extends ValidationBlock
{
    private static $inline_editable = false;

    private static $singular_name = 'not inline validation block';

    private static $plural_name = 'not inline validation blocks';

    public function getType()
    {
        return _t(__CLASS__ . '.BlockType', 'Not inline Validation block');
    }

}
```

The problem might be with how I'm using `UrlField`.

I tried hacking my numeric field to always fail. The page save went through, but the other scenarios got blocked.

[validation-not-triggering-2024-04-09_18.09.14.webm](https://github.com/silverstripe/silverstripe-elemental/assets/1168676/e203f019-6d6f-4463-9802-09838adc43e2)

emteknetnz:
Re numeric field - the inline validation was working correctly in your video. This PR is only for inline validation.  It does not cover non-inline dataobject validation / page validation.  Those concerns are being handled on https://github.com/silverstripe/silverstripe-elemental/issues/1155.  

Re the UrlField - that's really weird. I've copied your way of adding UrlField to my local using `$this->beforeUpdateCMSFields()` and validation works correctly for me when inline saving. Could you have a look in your network panel to see the XHR request that's being returned when you inline save a non-url in that field? For me I'm getting a formschema response with an `errors` node

Perhaps try dumping this into your project, it works for me

```php
<?php

use SilverStripe\Forms\UrlField;
use SilverStripe\Forms\FieldList;

class MrBlock extends BaseElement
{
    private static $db = [
        'MyUrlOne' => 'Varchar',
        'MyUrlTwo' => 'Varchar',
    ];

    private static $table_name = 'MrBlock';

    private static $singular_name = 'Mr Block';

    private static $plural_name = 'Mr Blocks';

    private static $description = 'This is mr block';

    private static $icon = 'font-icon-block-content';

    public function getType()
    {
        return 'Mr Block';
    }

    private static $inline_editable = true;
 
    {
        $this->beforeUpdateCMSFields(function (FieldList $fields) {
            $fields->addFieldToTab(
                'Root.Main',
                UrlField::create('MyUrlTwo', 'My Url Two')
            );
        });

        $fields->removeByName('MyUrlOne');
        $fields->addFieldToTab('Root.Main', UrlField::create('MyUrlOne', 'My Url One'));
        return $fields;
    }
}
```
image

If that's still not working, please confirm you have the latest hash of this PR as well as [this PR](https://github.com/silverstripe/silverstripe-admin/pull/1685)

maxime-rainville:
I'm an idiot.

`addFieldsToTab()` expects its second parameter to be an array of fields. If you give it a field without wrapping it in an array, it will still work however.

However, if you try passing another form field to `addFieldsToTab()`'s third parameter, nothing will happen and it will be ignored.

So my UrlField wasn't doing anything and the URL field was being scaffolded as a plain text field.

I created an issue to strongly type FieldList in CMS 6 https://github.com/silverstripe/silverstripe-framework/issues/11198

maxime-rainville (approved):
Begrudgingly approved with the understanding that #1155 and #1164 will follow up

GuySartorelli:
Merging on Max's behalf

PR #1151, opened 2024-02-26
FIX Remove workaround for nested redux form.

GuySartorelli:
Requires https://github.com/silverstripe/silverstripe-admin/pull/1694

## Issue
- https://github.com/silverstripe/silverstripe-admin/issues/1662


sabina-talipova (approved):
LGTM

Issue #1130, opened 2024-01-07
Adding elemental to an existing page publishes draft content 


satrun77:
### Module version(s) affected

4.11.7

### Description

Pages are recursively published when elemental is added to an existing page. This means draft content are published without the knowledge of the content authors. 

The issue is caused by the `isPublished` 
https://github.com/silverstripe/silverstripe-elemental/blob/5/src/Extensions/ElementalAreasExtension.php#L338

The issue exists in version `5.x`

### How to reproduce

- Install default CMS setup
- Publish a page.
- Modified any field within the page. Save only.
- The page on the live site does not show draft content.
- Add elemental
- Run `dev/build`
- The page on the live site shows draft content.

### Possible Solution

There is no need to check if a page is published or not. And we don't need to publish pages. This could be an internal process that saves the elemental area without creating a history record and does not publish the changes. If an author wants to publish new content, they can add blocks and then publish.

### Validations

- [X] Check that there isn't already an issue that reports the same bug
- [X] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

### Acceptance criteria
- Pre-existing page that are updated to contain an elemental area do not change versionned state following the subsequent `dev/build`.

### PR
- https://github.com/silverstripe/silverstripe-elemental/pull/1162

GuySartorelli:
@emteknetnz Can you please link to any relevant PRs?

emteknetnz:
Opps updated

GuySartorelli:
PR merged. Will be automatically tagged by GitHub actions

PR #1162, opened 2024-04-03
FIX Do not publish dataobject in requireDefaultRecords()

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-elemental/issues/1130

As far as I can tell there's absolutely no reason to publish the Page/DataObject at this point.  I think the intention was originally to ensure that any owned ElementalAreas would be published ... however they'll just get published when you publish the page/dataobject

I even tested the following succesfully:
- Remove the elemental extension, dev/build flush
- Create a blocks page, publish
- Edit page content, save without publish
- Enable elemental extension, dev/build flush
- Create an elemental and inline publish it (note this will mistakenly indicate the page publish button has published, however refreshing the page will show the correct state i.e. page is modified)

There are no network or console warnings when doing this, database records have the correct IDs etc


GuySartorelli (approved):
LGTM

PR #1167, opened 2024-04-11
MNT Add behat test - Restore archived block

sabina-talipova:
## Parent issue
- https://github.com/silverstripe/silverstripe-versioned-admin/issues/330


GuySartorelli (approved):
LGTM

PR #1169, opened 2024-04-17
API Remove deprecated code

emteknetnz:


GuySartorelli (approved):
I don't see any immediate problems with this, we'll see what happens after the follow-up cards, but for now this seems okay.

Issue #1164, opened 2024-04-08
Refactor inline saving and publishing


emteknetnz:
Follow up to https://github.com/silverstripe/silverstripe-elemental/pull/1150

The linked PR was merged with some suboptimal code due to it being open for a very long time. There are some comment in the peer review about possible ways to fix some of the suboptimal code. This card aims to fix that code.

The core reason for the suboptimal was there were existing GraphQL requests being made using a HOC pattern which meant the requests needed to be made from nested buttons, rather than centrally on Element.js.  However the version of Apollo that we're now on (v3) [uses hooks](https://www.apollographql.com/docs/react/data/mutations/), and Element.js is still a class component.

## Acceptance criteria
- [x] Refactor Element.js to be a functional component.
- [x] Refactor the save and publish GraphQL calls so they do not use the HOC pattern and instead are called within Element.js, probably via the useMutation() hook in apollo
- [ ] Refactor out the following, though only if what replaces it is less convoluted:
  - [ ] Triggering actions on state change
  - [ ] Use of setTimeout in Element.js
  - [ ] Moving context to its own file ... or remove the need for context.
  - [ ] Passing down callbacks in context (note that context is required to pass props to the buttons because they're added via js injector in registerTransforms.js)
  - [x] Use of window.ss.apolloClient.queryManager.reFetchObservableQueries to refetch elementalArea's

## New issues created
- https://github.com/silverstripe/silverstripe-elemental/issues/1170
- https://github.com/silverstripe/silverstripe-linkfield/issues/276
- https://github.com/silverstripe/silverstripe-elemental/issues/1177

## PRs
- https://github.com/silverstripe/silverstripe-elemental/pull/1171
- https://github.com/silverstripe/silverstripe-elemental/pull/1174
- https://github.com/silverstripe/silverstripe-elemental/pull/1175 << refactor without graphql update
- https://github.com/silverstripe/silverstripe-elemental/pull/1176

**Once the above PR is merged, assign back to Steve to create a follow up PR to do the rest of the refactoring**

GuySartorelli:
First PR merged, assigning back to Steve for next steps

emteknetnz:
Regarding this AC:
- Passing down callbacks in context (note that context is required to pass props to the buttons because they're added via js injector in registerTransforms.js)

[...]

PR #1171, opened 2024-04-18
ENH Refactor Element to function component

emteknetnz:

I'm splitting the linked issue into two PRs
- Firstly I'll do this one that only refactors the existing class component to a function component
- Once this is merged, I'll do a second PR to do the GraphQL refactor

Note that I've removed the `getDerivedStateFromError() / childRenderingError` functionality because:
- I didn't find a replacement for it using functional components
- We don't use getDerivedStateFromError() in any other components, so I think for consistency we don't actually want it


GuySartorelli:
I am reliably getting this error in the developer console whenever I expand an inline-editable block and then click "save" in the page (note: NOT when I click the `...` menu and save the block from there)

> Uncaught (in promise) Invariant Violation: Store reset while query was in flight (not completed in link chain)

Other than that it seems to work as expected locally. Haven't taken a proper look at the code yet but at a glance it looks okay.

emteknetnz:
That's an existing console error unrelated to this PR. It's coming from Apollo.

GuySartorelli:
Please replace `getDerivedStateFromError()` with `componentDidCatch()` as per https://react.dev/reference/react/Component#catching-rendering-errors-with-an-error-boundary

I'm not sure exactly what scenarios this is used in, but I have seen the "Something went wrong with this block. Please try saving and refreshing the CMS" error a bunch of times in my time developing with this module and it's always pointed out a problem that needed to be fixed.
Without this, it seems likely there will be silent failures, or failures which are only indicated in the developer console, which previously would have been clearly pointed out in the DOM.

This is just a refactor, not a change of functionality, so where there's a suitable functional component replacement (and in this case there is), we should just use it so that the functionality after the refactor is as close as possible to the functionality before the refactor.

If we decide later we want to remove this, we can remove it at that time - but that's not a decision I think we should be making here.

GuySartorelli:
I had originally requested replacing `getDerivedStateFromError()` but it looks like error boundaries are no longer directly supported in react. While I think losing that functionality isn't desirable, keeping it is probably more effort than it's worth right now.

GuySartorelli (approved):
LGTM

PR #1174, opened 2024-04-22
ENH Refactor Element to use mutation hook

emteknetnz:

Notes:
- I haven't implemented the AC "(refactor) Triggering actions on state change" - I've made a comment about it [here](https://github.com/silverstripe/silverstripe-elemental/issues/1164#issuecomment-2068481613) - I don't intend to implement it though if someone is able to demonstrate a non-convoluted way to implement an alternative I'm happy to refactor
- I haven't implemented the AC "Passing down callbacks in context (note that context is required to pass props to the buttons because they're added via js injector in registerTransforms.js)" - I've made a comment about it [here](https://github.com/silverstripe/silverstripe-elemental/issues/1164#issuecomment-2068381327). I do not intend to implement this AC
- I've removed some jest tests on SaveAction-test.js and PublishAction-test.js as there is simply now less logic and props passed via context in SaveAction and PublishAction. I choose not to add in equivalent tests in Element-test.js adding in "child Actions" would require doing stuff with injector to replicate what registerTransforms.js does now, and also because these functionality is already covered by behat tests in edit-block-element.feature (save button) and publish-block-element.feature (publish button)

**Update:**
Don't merge this, just keep this here for reference

Got extremely stuck where somehow a second Element was being created though not actually used. Looks like it was being created as part of handleFormSchemaSubmitResponse() where state hooks were being set, though possibly something wasn't wrapped in useEffect() things got out of sync.  This was detected because there was  failing behat test where the _page_ was submitted with invalid data, then immediately inline save the element, which should also fail and pop open the inline form. However the inline form wasn't popping open because the duplicated element was receiving the new props

### DIFF
--- a/client/src/components/ElementEditor/ElementEditor.js
+++ b/client/src/components/ElementEditor/ElementEditor.js
@@ -11,6 +11,7 @@ import sortBlockMutation from 'state/editor/sortBlockMutation';
 import ElementDragPreview from 'components/ElementEditor/ElementDragPreview';
 import withDragDropContext from 'lib/withDragDropContext';
 import { createSelector } from 'reselect';
+import md5 from 'md5';
 
 /**
  * The ElementEditor is used in the CMS to manage a list or nested lists of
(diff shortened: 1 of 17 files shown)


client/src/components/ElementEditor/ElementEditor.js
```diff
@@ -128,6 +133,7 @@ ElementEditor.propTypes = {
   }),
 };
 
+const filteredElementFormStates = {};
```
emteknetnz, line 136: I tried add this to [the "graphql2" PR](https://github.com/silverstripe/silverstripe-elemental/pull/1175) however it caused the issue where submitting a page with an invalid elemental field and then inline saving did not pop open the preview

Don't try and make this code work, it's not necessary

PR #1176, opened 2024-04-23
ENH Refactor Element and graphql hoc

emteknetnz:

Notes:

- I haven't implemented the AC "(refactor) Triggering actions on state change" - I've made a comment about it https://github.com/silverstripe/silverstripe-elemental/issues/1164#issuecomment-2068481613 - I don't intend to implement it though if someone is able to demonstrate a non-convoluted way to implement an alternative I'm happy to refactor
- I haven't implemented the AC "Passing down callbacks in context (note that context is required to pass props to the buttons because they're added via js injector in registerTransforms.js)" - I've made a comment about it https://github.com/silverstripe/silverstripe-elemental/issues/1164#issuecomment-2068381327. I do not intend to implement this AC
- I haven't implmented the AC "Use of setTimeout in Element.js" as things don't work when setTimeout is removed, presumably due to timing issues. I have made it to the original timeout happens inside a .then() block so it always happens after the HTTP request has returned.
- I haven't implemented the AC "Moving context to its own file ... or remove the need for context." as I was asked to revert it in [this comment](https://github.com/silverstripe/silverstripe-elemental/pull/1176#discussion_r1582434879) in peer review
- I've removed some jest tests on SaveAction-test.js and PublishAction-test.js as there is simply now less logic and props passed via context in SaveAction and PublishAction. I choose not to add in equivalent tests in Element-test.js adding in "child Actions" would require doing stuff with injector to replicate what registerTransforms.js does now, and also because these functionality is already covered by behat tests in edit-block-element.feature (save button) and publish-block-element.feature (publish button)
- I've commented out a failing behat test and raised a follow card to address https://github.com/silverstripe/silverstripe-elemental/issues/1177. It's a very weird and convoluted issue which I spent a great deal of time on though was ultimately unable to resolve


client/src/components/ElementActions/PublishAction.js
```diff
@@ -1,31 +1,21 @@
 /* global window */
-import React, { useContext, useEffect } from 'react';
+import React, { useContext } from 'react';
 import { compose } from 'redux';
 import AbstractAction from 'components/ElementActions/AbstractAction';
 import publishBlockMutation from 'state/editor/publishBlockMutation';
```
GuySartorelli: Do we still need `publishBlockMutation` here? I notice it's been added to `Element.js` so I'm hoping the intention is that it doesn't live here anymore?

client/src/components/ElementEditor/Element.js
```diff
@@ -104,6 +110,50 @@ const Element = (props) => {
...
+  }, [formHasRendered, doSaveElement]);
+
+  // Publish action
+  useEffect(() => {
+    if (formHasRendered && doPublishElement) {
+      publishBlock({ variables: { blockId: props.element.id } })
+        .then(() => handleAfterPublish(false))
+        .catch(() => handleAfterPublish(true));
+    }
+  }, [formHasRendered, doPublishElement]);
```
GuySartorelli, line 156: I notice the `handlePublishBlock` from `publishBlockMutation` isn't being used here - should that be removed?

client/src/components/ElementEditor/Element.js
```diff
@@ -104,6 +110,50 @@ const Element = (props) => {
...
+    // and set the form state to dirty under certain conditions
+    // setTimeout is a hackish way to do this, though I'm not sure how else we can do this
+    // The core issue is that redux-form will detect changes when a form is hydrated for the first
+    // time under certain conditions, specifically during a behat test when trying to publish a closed
+    // block when presumably the apollo cache is empty (or something like that). This happens late and
+    // there are no hooks/callbacks available after this happens the input onchange handlers are fired
+    Promise.all(refetchElementalArea())
+      .then(() => {
+        setTimeout(() => props.dispatchRemoveFormChanged(), 250);
+      });
```
GuySartorelli, line 138: This is the kind of thing the refactor is intended to avoid... it feels like we're just trading one bad code smell for another here. Do you have a sense for whether the refactor has overall improved code quality? If it has then I guess I can live with this....

client/src/components/ElementEditor/Element.js
GuySartorelli: Note that `handlePublishBlock` has a `refetchQueries` that doesn't seem to be represented here.

client/src/components/ElementEditor/ElementContext.js
```diff
@@ -0,0 +1,3 @@
+import { createContext } from 'react';
+
+export const ElementContext = createContext(null);
```
GuySartorelli: Having this in a separate file isn't a useful abstraction if we're not setting any values in the context here.
If there are no default values we can give the context, please just move it back into `Element`.

tests/Behat/features/individual-block-validation.feature
```diff
@@ -72,17 +72,19 @@ Feature: Blocks are validated when inline saving individual blocks
...
+#    Then I should see "MyField cannot be x" in the ".alert.error" element
+#    When I press the "View actions" button
+#    And I click on the ".element-editor__actions-save" element
+#    # Same validation error after page load with inline save
+#    And I should see "MyField cannot be x" in the ".form__validation-message" element
+#    Then I fill in "abc" for "My Field" for block 1
+#    # Ensure react field is filled in before submitting
+#    And I wait for 1 second
+#    # Need to save the whole page to stop the alert
+#    And I press the "Save" button
```
GuySartorelli: If this was passing before the refactor, it should be passing after the refactor instead of having that split as a separate issue.

GuySartorelli (changes requested):
Haven't actually tested it yet - will do that after these questions have been addressed in case they result in substantial changes.

client/src/components/ElementEditor/Element.js
emteknetnz: This is still an improvement we're always going to call props.dispatchRemoveFormChanged() after the graphql XHR in refetchElementalArea(), which previously wasn't guaranteed.

client/src/components/ElementActions/PublishAction.js
emteknetnz: Ah whoops, forgot to remove

client/src/components/ElementEditor/Element.js
emteknetnz: I've just removed the unused code from publishBlockMutation(), so that it's now only the `gql` mutation.

handleAfterPublish() calls refetchElementalArea() which is the equivalent of what was happening with the refetchQueries in publishBlockMutation() before

client/src/components/ElementEditor/ElementContext.js
emteknetnz: OK will move it back and leave a note in the description of this PR why we're not implementing that AC

tests/Behat/features/individual-block-validation.feature
emteknetnz: It was passing before, however only for the save button. The publish button is untested though if it was tested it would fail.  Therefore yes we are introducing a regression, though because it's related to an existing regression so I think it's correct commenting this out here

Note I spend about a day of time trying to fix this particular issue and I couldn't, so part of this is me deciding that I should quit trying to make this work and move on to more productive things. This regression is really a bit of a low importance edge case because it requires the user to first save the page, and then switch to inline saving which seems like a strange thing for a user to do. Also the inline form still submits and validation error still shows, it's just the user needs to manually open the element in order to see it

GuySartorelli: Either remove the test completely, with a note to add in the same or equivalent test in the linked issue, or uncomment the test and make it work here. We don't include commented out code/tests in PRs.

> This regression is really a bit of a low importance edge case because it requires the user to first save the page, and then switch to inline saving which seems like a strange thing for a user to do.

I don't think that's strange at all.
Consider a content author editing a page. They may make changes in several places. But then a specific block has validation errors. So they fix that block, and save it on its own to make sure it passes validation before trying to save the page (and the rest of their content) again.

Nevertheless, the impact *is* quite small since it's just whether the block is open or not. It can be handled as a follow up issue, but not left commented out like this.

client/src/components/ElementEditor/Element.js
```diff
@@ -37,6 +40,7 @@ const Element = (props) => {
   const [ensureFormRendered, setEnsureFormRendered] = useState(false);
   const [formHasRendered, setFormHasRendered] = useState(false);
   const [doDispatchAddFormChanged, setDoDispatchAddFormChanged] = useState(false);
+  const [publishBlock] = useMutation(publishBlockMutation);
```
GuySartorelli, line 43:
```suggestion
  const [publishBlock] = useMutation(publishBlockMutation, {
    // Refetch versions after mutation is completed
    refetchQueries: [{
      query: readBlocksQuery,
      variables: readBlocksConfig.options({ areaId }).variables
    }]
  });
```
Refetch queries as part of the mutation instead of as a separate action. I think this will be better for timing and will hopefully remove the need for that timeout, since this is the equivalent to how it was being done before the refactor.

GuySartorelli: You'll obviously need to make sure those variables are available - I pulled that code from `publishBlockMutation` based on [apollo docs for refetching queries post-mutation](https://www.apollographql.com/docs/react/data/mutations#refetching-queries)

GuySartorelli (changes requested):

GuySartorelli:
Seems to work fine in the browser as-is so we're just dealing with code quality at this point. If the refetch queries change I've suggested above doesn't remove the need for the timeout, it may not be worth doing.

tests/Behat/features/individual-block-validation.feature
emteknetnz: Removed and added note to the linked issue for add the test back in

client/src/components/ElementEditor/Element.js
emteknetnz: I don't think this is a great idea because while we do use it in handleAfterPublish() which do call refetchElementalArea() after the graphql query to publish, we also call refetchElementalArea() in handleFormSchemaSubmitResponse() which happens after an inline save which has nothing to do with graphql. 

Also note that the setTimeout() bit deals with props.dispatchRemoveFormChanged() which is to do with redux-form not graphql.  That happens AFTER the publish call to refetchElementalArea() which has been refactored in this PR to happen after a promise has been resolved.

GuySartorelli (approved):
Looks okay. I still don't like that timeout but it sounds like we have to accept some code smells in order to remove others.

Issue #1155, opened 2024-03-07
Non-inline validation only half works


emteknetnz:
While working on[ inline-validation for elemental](https://github.com/silverstripe/silverstripe-elemental/issues/329) it was discovered that the current state of validation for non-inline saving is pretty bad

## Validation on page save

Page save is somewhat busted. `MyBlock::validate()` style validation does work .. however the validation message will show at the top of the page rather than on the element as described in this bug

`MyBlock::getCMSCompositeValidator()` aka `RequiredFields` validation doesn't seem to work, nor does `FormField::validate()` - I tried putting a UrlField on my block which has its own validate() method though I could save invalid values.

## Validation on "not inline" block.

I assume that "not inline" block means setting private static $inline_editable = false;

It's just a regular DataObject EditForm at this point. It work much better than blocks on the Page EditForm, though RequiredFields is still a bit busted because instead of a nice red error message you get a red toast that says "validation error" and no indication of what's invalid.

Same issue with a toast showing when adding a message via `Model::validate()` using `addError()`, as opposed to `addFieldError()` which does work OK

## Acceptance criteria
- Validation rules for inline and non-inline block work the same. Any validation rule that would be triggered for an inline block will also be triggered if the block is not in-line.
- Errors originating from individual form fields (e.g. an invalid URL in a UrlField) are picked up by both block types.
- Saving/publishing a page will trigger validation on its child block if they are in a a dirty state.

## Related
- https://github.com/silverstripe/silverstripe-elemental/issues/329
- https://github.com/silverstripe/silverstripe-elemental/issues/1179 (Jianbinzhu)

## Multi PR CI
- https://github.com/emteknetnz/silverstripe-installer/actions/runs/8961925960

Note that the endtoend cms failures are [existing](https://github.com/silverstripe/silverstripe-installer/actions/runs/8869075415/job/24349372594)

## PRs
- https://github.com/silverstripe/silverstripe-framework/pull/11213
- https://github.com/silverstripe/silverstripe-frameworktest/pull/172
- https://github.com/silverstripe/silverstripe-elemental/pull/1178

maxime-rainville:
Chat GPT suggested using "standalone block" as the opposite of "inline block".

GuySartorelli:
PR merged. Neeeeeeext!

Issue #1170, opened 2024-04-18
Unable to inline save related data


emteknetnz:
After merging [Validation of fields when inline saving](https://github.com/silverstripe/silverstripe-elemental/pull/1150) looks like we're unable to consistently to inline save related data fields on custom blocks such as TreedropdownField and UploadField.

This will also affect the ability to inline clear a related field that currently has a value

When using the regular page save things seem to work correctly when using auto-scaffolded fields

### TreedropdownField

Auto-scaffolded field with an 'ID' suffix will xhr POST:
```
PageElements_4_MyPageID[value] "2"
PageElements_4_MyPageID[label] "About Us"
PageElements_4_MyPageID[selected] "false"
```
Which will result in a 500

Removing the suffixes will not xhr POST data for the Treedropdown field at all resulting in a 200, and field will just clear itself when the react component re-renders because the database value was never updated

### UploadField

Auto-scaffolded field without in an 'ID' suffix will xhr POST:
```
PageElements_4_MyFile[Files][0] "2"
```

Which results in a 200, but database isn't updated

Adding the suffix will xhr POST
```
PageElements_4_MyFileID[Files][0] "2"
```

Which results in a 500

```php
// app/src/MyBlock;

use DNADesign\Elemental\Models\BaseElement;

class MyBlock extends BaseElement
{
    private static $has_one = [
        'MyPage' => SiteTree::class,
        'MyFile' => File::class,
    ];

    public function getCMSFields()
    {
        $fields = parent::getCMSFields();

        // remove auto-scaffolded field with an ID suffix and add in field without ID suffix
        $fields->removeByName('MyPageID');
        $fields->addFieldToTab('Root.Main', TreeDropdownField::create('MyPage', 'My Page', SiteTree::class));

        // remove auto-scaffolded field without an ID suffix and add in a field with an ID suffix
        $fields->removeByName('MyFileID');
        $fields->addFieldToTab('Root.Main', UploadField::create('MyFileID', 'My File')->setIsMultiUpload(false));

        return $fields;
    }
}
```

## PRs
- https://github.com/silverstripe/silverstripe-frameworktest/pull/176
- https://github.com/silverstripe/silverstripe-elemental/pull/1180

GuySartorelli:
PRs merged

PR #1178, opened 2024-04-29
FIX Improve non-inline validation

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-elemental/issues/1155

Need to merge https://github.com/silverstripe/silverstripe-framework/pull/11213 first to get the has_one RequiredFields validation working correctly and https://github.com/silverstripe/silverstripe-frameworktest/pull/172 to get behat to pass.

There is a mutli-pr CI run linked on the issue

Note that this PR does not touch inline-saving

**Scenarios:**
1. DataObject validate() addFieldError()
2. DataObject validate() arrError()
3. UrlField validation
4. RequiredFields on default Title field
5. has_one relation to File and RequiredFields
6. has_one relation to SiteTree and RequiredFields

**Non-inline editable block:**
1. Already worked
2. Already worked
3. Already worked
4. Fixed in this PR - by default it's a composite field and validation works but doesn't show message on front-end
5. Already worked
6. Fixed in framework PR

**Inline-editable block page save**
1. Already worked - UX is bad not fixed in this PR - [issue](https://github.com/silverstripe/silverstripe-elemental/issues/1179)
2. Already worked - UX is bad not fixed in this PR - [issue](https://github.com/silverstripe/silverstripe-elemental/issues/1179)
3. Fixed in this PR
4. Fixed in this PR
5. Already worked
6. Fixed in framework PR

I used the following blocks to tests these scenarios

```php
<?php

use SilverStripe\Forms\CompositeValidator;
use SilverStripe\Forms\RequiredFields;
use SilverStripe\Forms\UrlField;
use SilverStripe\Assets\Image;

{
    private static $db = [
        'MyField' => 'Varchar',
        'MyHTML' => 'HTMLText',
        'MyUrl' => 'Varchar',
    ];

    private static $has_one = [
        'MyImage' => Image::class,
        'MyPage' => SiteTree::class,
    ];

    private static $table_name = 'MyBlock';

    private static $singular_name = 'My Block';

    private static $plural_name = 'My Blocks';

    private static $description = 'This is my block';

    private static $icon = 'font-icon-block-content';

    public function getType()
    {
        return 'My Block';
    }

    public function validate()
    {
        $result = parent::validate();
        if ($this->Title == 'x') {
            $result->addFieldError('Title', 'Title cannot be x', 'validation');
        }
        if ($this->MyField == 'x') {
            $result->addFieldError('MyField', 'MyField cannot be x', 'validation');
        }
        $c = trim(strip_tags($this->MyHTML ?? ''));
        if ($c == 'x') {
            $result->addFieldError('MyHTML', 'MyHTML cannot be x', 'validation');
        }
        if ($this->MyField == 'z' && $c == 'z') {
            $result->addError('This is a general error message');
        }
        return $result;
    }

    public function getCMSCompositeValidator(): CompositeValidator
    {
        return CompositeValidator::create([RequiredFields::create(['Title', 'MyImage', 'MyPageID'])]);
    }

    {
        $fields->dataFieldByName('MyHTML')->setRows(5);
        $fields->removeByName('MyUrl');
        // URLfield contains its own validate() methods
        $fields->addFieldToTab('Root.Main', UrlField::create('MyUrl', 'My URL'));
        return $fields;
    }
}
```

```php
<?php

class MyNonInlineBlock extends MyBlock
{
    private static $inline_editable = false;

    private static $table_name = 'MyNonInlineBlock';

    private static $singular_name = 'My non-inline Block';

    private static $plural_name = 'My non-inline Blocks';

    private static $description = 'This is my non-inline block';

    private static $icon = 'font-icon-block-content';

    public function getType()
    {
        return 'My non-inline Block';
    }
}
```

### DIFF
--- a/src/Controllers/ElementalAreaController.php
+++ b/src/Controllers/ElementalAreaController.php
@@ -139,7 +139,7 @@ class ElementalAreaController extends CMSMain
         $dataWithoutNamespaces = static::removeNamespacesFromFields($data, $element->ID);
 
         // Update and write the data object which will trigger model validation
-        $element->update($dataWithoutNamespaces);
+        $element->updateFromFormData($dataWithoutNamespaces);
         if ($element->isChanged()) {
             try {
                 $element->write();
--- a/src/Forms/ElementalAreaField.php
+++ b/src/Forms/ElementalAreaField.php
@@ -10,11 +10,14 @@ use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Forms\CompositeField;
 use SilverStripe\Forms\FieldGroup;
 use SilverStripe\Forms\FieldList;
+use SilverStripe\Forms\Form;
 use SilverStripe\Forms\FormField;
 use SilverStripe\Forms\GridField\GridField;
 use SilverStripe\Forms\TabSet;
 use SilverStripe\ORM\DataObjectInterface;
 use Symbiote\GridFieldExtensions\GridFieldAddNewMultiClass;
+use SilverStripe\Forms\GridField\GridFieldDetailForm;
+use SilverStripe\Forms\GridField\GridFieldDetailForm_ItemRequest;
 
 class ElementalAreaField extends GridField
 {
@@ -214,33 +217,84 @@ class ElementalAreaField extends GridField
         return $this->setValue(json_decode($value ?? '', true));
     }
 
+    /**
+     * This will perform FormField validation
+     * DataObject validation will happen in saveInto() as part of $element->write()
+     */
+    public function validate($validator)
+    {
+        $result = true;
+        foreach ($this->getElementData() as $arr) {
+            /** @var BaseElement $element */
+            /** @var array $data */
+            $element = $arr['element'];
+            $data = $arr['data'];
+            $elementForm = $this->getElementForm($element);
+            if (!$elementForm) {
+                continue;
+            }
+            $elementForm->loadDataFrom($data);
+            $formValidatorResult = $elementForm->getValidator()->validate();
+            if (!$formValidatorResult->isValid()) {
+                $validator->getResult()->combineAnd($formValidatorResult);
+                $result = false;
+            }
+        }
+        return $this->extendValidationResult($result, $validator);
+    }
+
     public function saveInto(DataObjectInterface $dataObject)
     {
         parent::saveInto($dataObject);
-
-        $elementData = $this->Value();
-        $idPrefixLength = strlen(sprintf(ElementalAreaController::FORM_NAME_TEMPLATE ?? '', ''));
-
-        if (!$elementData) {
-            return;
+        foreach ($this->getElementData() as $arr) {
+            /** @var BaseElement $element */
+            /** @var array $data */
+            $element = $arr['element'];
+            $data = $arr['data'];
+            $element->updateFromFormData($data);
+            $element->write();
         }
+    }
 
-        foreach ($elementData as $form => $data) {
+    private function getElementData(): array
+    {
+        $elementData = [];
+        $value = $this->Value();
+        if (!$value) {
+            return [];
+        }
+        $idPrefixLength = strlen(sprintf(ElementalAreaController::FORM_NAME_TEMPLATE ?? '', ''));
+        foreach ($value as $form => $data) {
             // Extract the ID
             $elementId = (int) substr($form ?? '', $idPrefixLength ?? 0);
-
-            /** @var BaseElement $element */
             $element = $this->getArea()->Elements()->byID($elementId);
-
             if (!$element) {
                 // Ignore invalid elements
                 continue;
             }
-
             $data = ElementalAreaController::removeNamespacesFromFields($data, $element->ID);
-
-            $element->updateFromFormData($data);
-            $element->write();
+            $elementData[] = ['element' => $element, 'data' => $data];
         }
+        return $elementData;
+    }
+
+    private function getElementForm(BaseElement $element): ?Form
+    {
+        // This is essentially the same method used to generate a form to edit an element
+        // that a non-inline edit form will use - see GridFieldDetailForm::handleItem()
+        $requestHandler = $requestHandler = $this->getForm()->getController();
+        $gridFieldDetailForm = new GridFieldDetailForm();
+        // The validator needs to be explicitly copied from the element to the form
+        $validator = $element->getCMSCompositeValidator();
+        $gridFieldDetailForm->setValidator($validator);
+        $gridFieldDetailFormItemRequest = new GridFieldDetailForm_ItemRequest(
+            $this,
+            $gridFieldDetailForm,
+            $element,
+            $requestHandler,
+            ''
+        );
+        $form = $gridFieldDetailFormItemRequest->ItemEditForm();
+        return $form;
     }
 }
--- a/src/Forms/TextCheckboxGroupField.php
+++ b/src/Forms/TextCheckboxGroupField.php
@@ -32,6 +32,49 @@ class TextCheckboxGroupField extends CompositeField
         $this->setTitle($title);
     }
 
+    public function getMessage()
+    {
+        $message = parent::getMessage();
+        if ($message) {
+            return $message;
+        }
+        foreach ($this->getChildren() as $field) {
+            $message = $field->getMessage();
+            if ($message) {
+                return $message;
+            }
+        }
+        return '';
+    }
+
+    public function getMessageType()
+    {
+        $message = parent::getMessage();
+        if ($message) {
+            return parent::getMessageType();
+        }
+        foreach ($this->getChildren() as $field) {
+            if ($field->getMessage()) {
+                return $field->getMessageType();
+            }
+        }
+        return '';
+    }
+
+    public function getMessageCast()
+    {
+        $message = parent::getMessage();
+        if ($message) {
+            return parent::getMessageCast();
+        }
+        foreach ($this->getChildren() as $field) {
+            if ($field->getMessage()) {
+                return $field->getMessageCast();
+            }
+        }
+        return '';
+    }
+
     /**
      * Don't use the custom template for readonly states
      *
--- a/tests/Behat/features/inline-block-validation.feature
+++ b/tests/Behat/features/inline-block-validation.feature
@@ -1,3 +1,4 @@
+@retry
 Feature: Blocks are validated when inline saving individual blocks
   As a CMS user
   I want to blocks have be validating when individual saving them
@@ -14,6 +15,10 @@ Feature: Blocks are validated when inline saving individual blocks
     And I go to "/admin/pages"
     And I follow "Blocks Page"
     And I click on the caret button for block 1
+    And I click on the "#Form_ElementForm_1_PageElements_1_MyPageID" element
+    And I click on the ".ss-searchable-dropdown-field__option:nth-of-type(2)" element
+    And I press the "View actions" button
+    And I click on the ".element-editor__actions-save" element
 
   # Note that each test is split into a seperate scenario instead a large single scenario which would
   # be faster due to a limitation with behat testing react where changing the value of a field can
@@ -26,6 +31,7 @@ Feature: Blocks are validated when inline saving individual blocks
     # Will not be an inline save button because formDirty not set yet, intercepted by JS validation
     Then I should not see a ".element-editor__actions-save" element
 
+@sboyd
   Scenario: Field validation error
     When I fill in "x" for "Title" for block 1
     When I press the "View actions" button
--- a/tests/Behat/features/non-inline-block-validation.feature
+++ b/tests/Behat/features/non-inline-block-validation.feature
@@ -1,0 +1,56 @@
+@retry
+Feature: Blocks are validated when non-inline saving blocks
+  As a CMS user
+  I want to blocks have be validating when non-inline saving them
+  So that I can see what content needs to be fixed
+
+  Background:
+    Given I add an extension "DNADesign\Elemental\Extensions\ElementalPageExtension" to the "Page" class
+    And I add an extension "SilverStripe\FrameworkTest\Elemental\Extension\ElementContentExtension" to the "DNADesign\Elemental\Models\ElementContent" class
+    And I add an extension "SilverStripe\FrameworkTest\Elemental\Extension\NumericFieldExtension" to the "SilverStripe\Forms\NumericField" class
+    And content blocks are not in-line editable
+    And I go to "/dev/build?flush"
+    And a "page" "Blocks Page" with a "My title" content element with "My content" content
+    And the "group" "EDITOR" has permissions "Access to 'Pages' section"
+    And I am logged in as a member of "EDITOR" group
+    And I go to "/admin/pages"
+    And I follow "Blocks Page"
+    And I click on the caret button for block 1
+
+  Scenario: Non-inline block validation
+    
+    # Related has_one RequiredFields
+    When I press the "Save" button
+    Then I should see "\"My page\" is required" in the "#message-Form_ItemEditForm_MyPageID" element
+    And I click on the "#Form_ItemEditForm_MyPageID" element
+    And I click on the ".ss-searchable-dropdown-field__option:nth-of-type(2)" element
+
+    # RequiredFields on TextCheckboxGroupField (composite) field
+    When I fill in "Title" with ""
+    And I press the "Save" button
+    Then I should see "\"Title\" is required" in the "#message-Form_ItemEditForm_Title" element
+    And I fill in "Title" with "My title"
+    
+    # FormField::validate()
+    When I fill in "My Int" with "1"
+    And I press the "Save" button
+    Then I should see "This field cannot be 1" in the "#message-Form_ItemEditForm_MyInt" element
+    And I fill in "My Int" with "2"
+
+    # DataObject::validate() addFieldError()
+    When I fill in "My Field" with "x"
+    And I press the "Save" button
+    Then I should see "MyField cannot be x" in the "#message-Form_ItemEditForm_MyField" element
+    And I fill in "My Field" with "lorem"
+
+    # DataObject::validate() addError()
+    When I fill in "Title" with "z"
+    And I fill in "My Field" with "z"
+    And I press the "Save" button
+    Then I should see "This is a general error message" in the "#Form_ItemEditForm_error" element
+    And I fill in "Title" with "My title"
+    And I fill in "My Field" with "lorem"
+
+    # Success message
+    When I press the "Save" button
+    Then I should see "Saved content block \"My title\"" in the "#Form_ItemEditForm_error" element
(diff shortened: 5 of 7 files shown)


src/Forms/TextCheckboxGroupField.php
```diff
@@ -32,6 +32,58 @@ public function __construct($title = null)
         $this->setTitle($title);
     }
 
+    /**
```
emteknetnz: TextCheckboxGroupField is the special elemental field that is a composite field of the default `Title` and `ShowTitle` fields

src/Forms/ElementalAreaField.php
```diff
@@ -214,33 +217,83 @@ public function setSubmittedValue($value, $data = null)
         return $this->setValue(json_decode($value ?? '', true));
     }
 
+    /**
+     * This will perform FormField validation
+     * DataObject validation will happen in saveInto() as part of $element->write()
+     */
+    public function validate($validator)
```
emteknetnz, line 224: Previously FormField validation wasn't happening on fields on elements e.g. `$db = [ 'MyUrl' => UrlField::class ]` when saving the page (as opposed to editing the element directly)

tests/Behat/features/inline-block-validation.feature
```diff
@@ -14,7 +14,11 @@ Feature: Blocks are validated when inline saving individual blocks
     And I go to "/admin/pages"
     And I follow "Blocks Page"
     And I click on the caret button for block 1
+    And I click on the "#Form_ElementForm_1_PageElements_1_MyPageID" element
+    And I click on the ".ss-searchable-dropdown-field__option:nth-of-type(2)" element
 
+  # This section deals with validation when inline saving individual blocks
```
GuySartorelli: If we need comments like this separating out scenarios, it's probably better to just create a new feature file instead.

src/Models/BaseElement.php
```diff
@@ -664,6 +665,7 @@ public function updateFromFormData($data)
             }
 
             $field->setSubmittedValue($datum);
+            // $field->validate();
```
GuySartorelli: Don't leave commented-out code

src/Forms/TextCheckboxGroupField.php
GuySartorelli: These methods feel like they belong on `CompositeField` directly. Please move them there, and include some basic unit tests.

tests/Behat/features/non-inline-block-validation.feature
```diff
@@ -0,0 +1,56 @@
+Feature: Blocks are validated when non-inline saving blocks
+  As a CMS user
+  I want to blocks have be validating when non-inline saving them
+  So that I can see what content needs to be fixed
+
+  Background:
+    Given I add an extension "DNADesign\Elemental\Extensions\ElementalPageExtension" to the "Page" class
+    And I add an extension "SilverStripe\FrameworkTest\Elemental\Extension\ElementContentExtension" to the "DNADesign\Elemental\Models\ElementContent" class
+    And I add an extension "SilverStripe\FrameworkTest\Elemental\Extension\NumericFieldExtension" to the "SilverStripe\Forms\NumericField" class
+    And I add an extension "SilverStripe\FrameworkTest\Elemental\Extension\NonInlineEditableExtension" to the "DNADesign\Elemental\Models\ElementContent" class
```
GuySartorelli:
```suggestion
    And content blocks are not in-line editable
```

tests/Behat/features/non-inline-block-validation.feature
```diff
@@ -0,0 +1,56 @@
...
+    Then I should see "MyField cannot be x" in the "#message-Form_ItemEditForm_MyField" element
+    And I fill in "My Field" with "lorem"
+
+    # DataObject::validate() addError()
+    When I fill in "Title" with "z"
+    And I fill in "My Field" with "z"
+    And I press the "Save" button
+    Then I should see "This is a general error message" in the "#Form_ItemEditForm_error" element
+    And I fill in "Title" with "My title"
+    And I fill in "My Field" with "lorem"
```
GuySartorelli: Can these be split up into separate scenarios? If so, please do so since they're clearly each testing a distinct piece of validation as indicated by the comments.

GuySartorelli: Please do the same thing in the other feature file as well, if possible.

src/Forms/TextCheckboxGroupField.php
```diff
@@ -32,6 +32,58 @@ public function __construct($title = null)
...
+            if ($field->getMessage()) {
+                return $field->getMessageType();
+            }
+        }
+        return '';
+    }
+
+    /**
+     * @return string
+     */
```
GuySartorelli: If these methods are inherited from a superclass or interface, we don't need to (and shouldn't, in case it gets stale) duplicate the phpdoc typehints here.

If these are **not** inherited from a superclass or interface, please make these strict typehints instead.

This applies to all such methods in this PR.

src/Models/BaseElement.php
```diff
@@ -654,6 +654,7 @@ public function getRenderTemplates($suffix = '')
      */
     public function updateFromFormData($data)
     {
+        /** @var FieldList $cmsFields */
```
```suggestion
```
That's not needed

src/Forms/ElementalAreaField.php
```diff
@@ -214,33 +217,85 @@ public function setSubmittedValue($value, $data = null)
...
+        $value = $this->Value();
+        if (!$value) {
+            return [];
+        }
+        $idPrefixLength = strlen(sprintf(ElementalAreaController::FORM_NAME_TEMPLATE ?? '', ''));
+        foreach ($value as $form => $data) {
             // Extract the ID
             $elementId = (int) substr($form ?? '', $idPrefixLength ?? 0);
-
             /** @var BaseElement $element */
```
GuySartorelli:
```suggestion
```
Not needed

src/Forms/ElementalAreaField.php
```diff
@@ -214,33 +217,85 @@ public function setSubmittedValue($value, $data = null)
...
+            /** @var BaseElement $element */
+            /** @var array $data */
+            $element = $arr['element'];
+            $data = $arr['data'];
+            $elementForm = $this->getElementForm($element);
+            if (!$elementForm) {
+                continue;
+            }
+            $elementForm->loadDataFrom($data);
+            $formValidatorResult = $elementForm->validationResult();
```
GuySartorelli:
```suggestion
            $formValidatorResult = $elementForm->getValidator()->validate();
```
We should skip the logic in `validationResult()` which might skip validation - at this point we know for sure we want to validate the form, so we should just do it directly.

src/Forms/ElementalAreaField.php
```diff
@@ -214,33 +217,85 @@ public function setSubmittedValue($value, $data = null)
...
+            if (!$formValidatorResult->isValid()) {
+                $validator->getResult()->combineAnd($formValidatorResult);
+                $result = false;
+            }
+        }
+        return $this->extendValidationResult($result, $validator);
+    }
+
     public function saveInto(DataObjectInterface $dataObject)
```
GuySartorelli, line 246: What do the changes to the `saveInto()` block resolve? I tried changing this method back to what it was before this PR, and it seems to work exactly the same way as before.

GuySartorelli (changes requested):
If I save the page without expanding the block, none of the validation seems to happen. e.g. with your test blocks from the description, I add the block and immediately save the page - it should trigger validation errors for the missing required fields but it doesn't.

src/Forms/ElementalAreaField.php
emteknetnz: Updated

src/Forms/ElementalAreaField.php
emteknetnz: There's logic I moved to `private function getElementData()` which is called by both the old `saveInto()` method and the new `validate()` method

src/Forms/ElementalAreaField.php
emteknetnz: Removed

src/Forms/TextCheckboxGroupField.php
emteknetnz: I've [raised a new card](https://github.com/silverstripe/silverstripe-framework/issues/11215) for the refactor to do this. I feel like there's already a lot going on in this card including an existing framework PR that does different things, and there's [another issue](https://github.com/silverstripe/silverstripe-elemental/issues/1170) that's dependent on the PRs in this issue

GuySartorelli: If we keep it in this PR, that means you need to add unit tests here and then to refactor we're removing both the src code and the tests, which shouldn't have been added to this repo in the first place.
Moving them now doesn't introduce any new complexity, it just puts the new code in the right place the first time around.

src/Forms/TextCheckboxGroupField.php
emteknetnz: Inherited, removed

src/Models/BaseElement.php

src/Models/BaseElement.php

tests/Behat/features/non-inline-block-validation.feature
emteknetnz: Updated

tests/Behat/features/inline-block-validation.feature
emteknetnz: Separated

tests/Behat/features/non-inline-block-validation.feature
emteknetnz: Putting them in separate scenarios greatly increases how long it takes for behat to run through the feature file because it needs to rebuild the test database and redo the background steps. Elemental behat speed is already painfully slow and I don't want it taking even longer if possible.

emteknetnz: Had an offline discussion about this, decided it's not worth doing irrespective of performance difference

src/Forms/TextCheckboxGroupField.php
emteknetnz: Like I said, there's already an [existing framework PR ](https://github.com/silverstripe/silverstripe-framework/pull/11213) that deals with RequiredFields that's required for this PR to work.  I don't want to add CompositeFields functionality to that PR because that has nothing to do with RequireFields, so I want to do a seperate PR.  I don't want to block this PR on another framework PR.

It's just easier to refactor this later I don't see the problem

GuySartorelli: Required fields PR doesn't seem necessary, see comment on that PR.

If that PR does end up being necessary (and please don't comment about that here, comment about that in ***that*** PR), we can keep this here.
If that PR ends up not being necessary, then I think this code should be moved.

GuySartorelli (changes requested):
The problem described in https://github.com/silverstripe/silverstripe-elemental/pull/1178#pullrequestreview-2032837575 has not been resolved

src/Forms/ElementalAreaField.php
```diff
@@ -214,33 +217,84 @@ public function setSubmittedValue($value, $data = null)
...
-            $element->write();
+            $elementData[] = ['element' => $element, 'data' => $data];
         }
+        return $elementData;
+    }
+
+    private function getElementForm(BaseElement $element): ?Form
+    {
+        // This is essentially the same method used to generate a form to edit an element
+        // that a non-inline edit for will use - see GridFieldDetailForm::handleItem()
```
michalkleiner:
```suggestion
        // that a non-inline edit form will use - see GridFieldDetailForm::handleItem()
```

src/Forms/ElementalAreaField.php
```diff
@@ -214,33 +217,84 @@ public function setSubmittedValue($value, $data = null)
...
+        $requestHandler = $requestHandler = $this->getForm()->getController();
+        $gridFieldDetailForm = new GridFieldDetailForm();
+        // The validator need to be explicitly copied from the element to the form
```
michalkleiner:
```suggestion
        // The validator needs to be explicitly copied from the element to the form
```

src/Forms/TextCheckboxGroupField.php
```diff
@@ -32,6 +32,49 @@ public function __construct($title = null)
...
+        $message = parent::getMessage();
+        if ($message) {
+            return $message;
+        }
+        foreach ($this->getChildren() as $field) {
+            $message = $field->getMessage();
+            if ($message) {
+                return $message;
+            }
+        }
```
michalkleiner, line 46: Should this concat the messages from multiple fields? At the moment it would skip the second message if there were two of them.

src/Forms/TextCheckboxGroupField.php
GuySartorelli: Actually nevermind - after refinment today I'm okay with this being done separately.

emteknetnz:
> If I save the page without expanding the block, none of the validation seems to happen. e.g. with your test blocks from the description, I add the block and immediately save the page - it should trigger validation errors for the missing required fields but it doesn't.

That would be hard to solve.

In the case of inline blocks, currently if they haven't been opened and the page is saved, then then because the child FormBuilder form hasn't been rendered then the element data will not be sent as part of the page form submission.  However if the inline-element is opened then page save will submit element data with the rest of the page, and closing the element then saving the page will also send the element data.  So to solve this we'd need to render the FormBuilder forms for ALL unopened elements on page save

In the case of non-inline blocks, I really don't know what can be done since their forms are never rendered on the page and and thus their data is never sent.

The fact this doesn't work doesn't actually matter given the AC is `Saving/publishing a page will trigger validation on its child block if they are in a a dirty state.` - the elements in question aren't dirty, they're just created in an invalid state.

We could split this off as a seperate card, though I honestly don't know how or even if we should attempt a fix here

src/Forms/TextCheckboxGroupField.php
emteknetnz: [Other card](https://github.com/silverstripe/silverstripe-framework/issues/11215) has now been refined so we'll do on that card

GuySartorelli:
> We could split this off as a seperate card, though I honestly don't know how or even if we should attempt a fix here

I think that's worth looking into. There's a tradeoff to be made between:
1. always having the form rendered (and just hidden until it's expanded) - which means we never have to do workarounds like explicitly render it when hitting the save button on the block
1. potential performance loss from eagerly getting the required data to render the form for all blocks on a given page

I'll open a separate card about this for discussion.

Edit: https://github.com/silverstripe/silverstripe-elemental/issues/1183

src/Forms/TextCheckboxGroupField.php
emteknetnz: I think it OK to skip. Templates are setup to only show a single validation message per field, so there would need to be further changes made to support multiple validation messages. In this case of this particular field, this field is a combination of the Title and ShowTitle fields, and since ShowTitle is a boolean checkbox field its impractical to have validation on it.

This code will make its way over to Composite field as part of [this card](https://github.com/silverstripe/silverstripe-framework/issues/11215) - it may make sense to look at multi validation messages there

src/Forms/ElementalAreaField.php
emteknetnz: Updated

src/Forms/ElementalAreaField.php
emteknetnz: Updated

src/Forms/TextCheckboxGroupField.php
GuySartorelli: I agree with Michal here, sorry I didn't pick this up in my review, we could have saved some time.

> Templates are setup to only show a single validation message per field

While that is broadly true, this is an edge case where this isn't really *a field* in itself - it's really a holder for multiple fields. If each of the child fields has a validation error message, we should see those.

It would be bad UX to display only one validation message at a time, so users have to fix errors one field at a time, save, and see what fails next.

> since ShowTitle is a boolean checkbox field its impractical to have validation on it.

We can't really decide for developers what they should or shouldn't put validation on, unless we actually lock them out of doing so with exception messages when they try.

## Options
1. Have a new `getMessages()` method, and use that in the template for this form field, loop through the child messages and display all of them
2. Just update this method to concatenate the error messages together - at least the messages will display, even if it's not perfect
3. Explicitly defer this by adding ACs to the follow-up card you made. That's probably best, because there are additional things to consider such as the message type and cast, which may be different for the different messages.

GuySartorelli (commented):
Just the one thing left and then we're good, I think.

src/Forms/TextCheckboxGroupField.php
emteknetnz: Go with 3, I've updated AC's on the follow up card

GuySartorelli:
Reran CI after merging frameworktest. Confirmed they are running with the correct commit:
`silverstripe/frameworktest                     1.x-dev d12463a`
https://github.com/silverstripe/silverstripe-frameworktest/commit/d12463ae60997ed7f9235b8ea222daba4c4e6413

GuySartorelli:
behat failing

emteknetnz:
I've added an extra code change that's from a [different PR](https://github.com/silverstripe/silverstripe-elemental/pull/1180/files#diff-942115e4b8f6030e4d72ebc2b3a0772ec65e5dfcd08fbd0e677c70d1231daf28R142 ) that's blocked on this one which is required to get the inline-block-validation.feature behat tests to pass

src/Controllers/ElementalAreaController.php
```diff
@@ -139,7 +139,7 @@ public function save(array $data, Form $form): HTTPResponse
         $dataWithoutNamespaces = static::removeNamespacesFromFields($data, $element->ID);
 
         // Update and write the data object which will trigger model validation
-        $element->update($dataWithoutNamespaces);
+        $element->updateFromFormData($dataWithoutNamespaces);
```
emteknetnz, line 142: I've added this code change in from a [different PR](https://github.com/silverstripe/silverstripe-elemental/pull/1180/files#diff-942115e4b8f6030e4d72ebc2b3a0772ec65e5dfcd08fbd0e677c70d1231daf28R142) that's blocked on this one which is required to get the inline-block-validation.feature behat tests to pass

GuySartorelli (changes requested):
I don't know how I didn't notice this before, but for inline-editable blocks when you save the page, only one required field validation message appears at a time.

So if I don't enter any data I only get told about the title, then I add a title and I'm told about the image, then I add an image and I'm told about the page.

The HTTP response does have all three validation errors at the same time - at this stage I don't know if getting them to display is part of this card or not so if there's another card to handle this please just link to it and I'll be okay with that.

GuySartorelli:
Rerunning failed behat tests now that the framework PR is merged, but regardless of whether that goes green I need at least a response to my above comment.

GuySartorelli:
Still need to respond to https://github.com/silverstripe/silverstripe-elemental/pull/1178#pullrequestreview-2041868677

emteknetnz:
Split of as [separate card](https://github.com/silverstripe/silverstripe-elemental/issues/1187)  that can be refined

GuySartorelli (approved):
LGTM with the understanding that there's still a few issues left to be done before this functionality is considered 100% working 😅

PR #1180, opened 2024-05-01
MNT MNT Add behat tests for related data

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-elemental/issues/1170

Requires [frameworktest PR](https://github.com/silverstripe/silverstripe-frameworktest/pull/176) to be merged first

~~Requires PRs on a [different issue](https://github.com/silverstripe/silverstripe-elemental/issues/1155) to be merged first~~

Main code change for this PR ended up [in this different PR](https://github.com/silverstripe/silverstripe-elemental/pull/1178/files#diff-942115e4b8f6030e4d72ebc2b3a0772ec65e5dfcd08fbd0e677c70d1231daf28R142), so this PR is now just adding a behat test


src/Models/BaseElement.php
```diff
@@ -656,8 +656,8 @@ public function updateFromFormData($data)
     {
         $cmsFields = $this->getCMSFields();
 
-        foreach ($data as $field => $datum) {
-            $field = $cmsFields->dataFieldByName($field);
+        foreach ($data as $fieldName => $datum) {
+            $field = $cmsFields->dataFieldByName($fieldName);
```
emteknetnz: This is unrelated refactoring

src/Controllers/ElementalAreaController.php
```diff
@@ -139,7 +139,7 @@ public function save(array $data, Form $form): HTTPResponse
         $dataWithoutNamespaces = static::removeNamespacesFromFields($data, $element->ID);
 
         // Update and write the data object which will trigger model validation
-        $element->update($dataWithoutNamespaces);
+        $element->updateFromFormData($dataWithoutNamespaces);
```
emteknetnz: Using updateFromFormData() goes on a slightly longer code path that includes FormField::setSubmittedValue() which means that field values submitted as arrays get handled correctly, which is the case for UploadField and SearchableDropdownField

GuySartorelli (commented):
The issue this PR is connected to is about saving values, not validating them.
Are there behat tests already for testing upload field and `TreeDropdownField` can have values saved in inline-editable blocks when inline-saving the block?

emteknetnz:
There are existing tests for:
- [non-inline saving related data](https://github.com/silverstripe/silverstripe-elemental/blob/5/tests/Behat/features/non-inline-block-validation.feature#L22)
- [page saving related data](https://github.com/silverstripe/silverstripe-elemental/blob/5/tests/Behat/features/page-save-validation.feature#L23)

Both of these tests validation, and also that you can save the data elsewhere in the feature file as they test for a success message. However they only has TreeDropdownField tests and not UploadField tests

I've updated everything to now also have UploadField tests

> The issue this PR is connected to is about saving values, not validating them.

Yeah, as mentioned the "saving values" bit was sorted in a different PR [here](https://github.com/silverstripe/silverstripe-elemental/pull/1178/files#diff-942115e4b8f6030e4d72ebc2b3a0772ec65e5dfcd08fbd0e677c70d1231daf28R142) 

This particular issue is now largely redundant because of that PR.  So this PR is now just behat tests, though we should definitely still merge this so that we're consistently testing related data on all x3 ways to save a content block

GuySartorelli:
> Both of these tests validation, and also that you can save the data elsewhere in the feature file as they test for a success message.

I don't think just checking a success message is enough to correctly regression test the bug that this issue talks about - we need to check that the value is actually there after saving. i.e. "I select the option and I save - then I check the field has the correct value"

emteknetnz:
I've added extra steps to check the that related data values were saved

GuySartorelli (approved):
LGTM

PR #1190, opened 2024-05-16
API Set extension hook implementation visibility to protected

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10514

### DIFF
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -319,16 +319,18 @@ class BaseElement extends DataObject implements CMSPreviewable
         return parent::write($showDebug, $forceInsert, $forceWrite, $writeComponents, $skipValidation);
     }
 
+    protected function onBeforeWrite()
+    {
+        parent::onBeforeWrite();
+        $this->ensureSortSet();
+    }
+
     /**
      * Increment the sort order if one hasn't been already defined. This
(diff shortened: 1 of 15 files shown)


src/Models/BaseElement.php
```diff
@@ -325,10 +325,20 @@ public function write(
...
+        $this->updateSort();
+    }
 
+    /**
+     * Increment the sort order if one hasn't been already defined. This
+     * ensures that new elements are created at the end of the list by default.
+     *
+     * {@inheritDoc}
+     */
+    public function updateSort()
```
emteknetnz: This method is added so that Resolver.php can call it, previously it was calling onBeforeWrite() direclty

src/Models/BaseElement.php
```diff
@@ -319,16 +319,18 @@ public function write(
...
+    }
+
     /**
      * Increment the sort order if one hasn't been already defined. This
      * ensures that new elements are created at the end of the list by default.
-     *
-     * {@inheritDoc}
      */
-    public function onBeforeWrite()
+    public function ensureSortSet()
```
emteknetnz, line 332: This method has been added so that Resolver.php can call it directly, previously it was calling onBeforeWrite() directly

GuySartorelli (approved):
LGTM

Issue #1192, opened 2024-05-20
Composer version not set correctly to framework version when tagged


mandrew:
### Module version(s) affected

5.2.1

### Description

Currently composer install gets the latest version of elemental for any version of v5 framework `"silverstripe/framework": "^5",` which creates issues if you are still running an older version of framework e.g. 5.0.0.

I only worked out after reading the release info that as I'm currently working on a v5.0.0 site, I should be using v5.0.4 of elemental. It's a straightforward fix as a Developer, however composer should really handle these dependancies so I don't have to think about it :)

### How to reproduce

1. Install SIlverstripe 5.0.0 with elemental
2. Search in the CMS for a word that you know exists on a page
3. Error appears saying:
```text
[Emergency] Uncaught BadMethodCallException: Object->__call(): the method 'eagerLoad' does not exist on 'SilverStripe\ORM\DataList'

GET /admin/pages?q%5BFilterClass%5D=DNADesign%5CElemental%5CControllers%5CElementSiteTreeFilterSearch&q%5BTerm%5D=test

Line 57 in /var/www/html/vendor/silverstripe/framework/src/Core/CustomMethods.php
```

<details><summary>click to see full error and stack trace</summary>

```text
Source

48      * @throws BadMethodCallException

49      */

50     public function __call($method, $arguments)

51     {

52         // If the method cache was cleared by an an Object::add_extension() / Object::remove_extension()

53         // call, then we should rebuild it.

54         $class = static::class;

55         $config = $this->getExtraMethodConfig($method);

56         if (empty($config)) {

57             throw new BadMethodCallException(

58                 "Object->__call(): the method '$method' does not exist on '$class'"

59             );

60         }

61

62         switch (true) {

63             case isset($config['callback']): {

Trace

SilverStripe\View\ViewableData->__call(eagerLoad, Array)
ElementalPageExtension.php:138
DNADesign\Elemental\Extensions\ElementalPageExtension->getEagerLoadedElements()
ElementalPageExtension.php:85
DNADesign\Elemental\Extensions\ElementalPageExtension->getContentFromElementsForCmsSearch()
call_user_func_array(Array, Array)
Extensible.php:135
SilverStripe\View\ViewableData->SilverStripe\Core\{closure}(ACC\SYA\Pages\TopicPage, Array)
CustomMethods.php:64
SilverStripe\View\ViewableData->__call(getContentFromElementsForCmsSearch, Array)
[...]

PR #1193, opened 2024-05-21
DEP Bump framework dependency

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-elemental/issues/1192

I had a look through the [diff between 5.0 and 5](https://github.com/silverstripe/silverstripe-elemental/compare/5.0...5) and I think the new eager loading API framework was the only thing that was missed

Note that it's basically too late to fix this on 5.2 and below.  We could fix this on elemental 5.2.2, however all that will do is cause composer to install 5.2.1 if you're on framwork 5.0.x which won't actually fix the missing eager loading API, so I think for the sake of less confusion it's better to just target this at elemental 5.3.0

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
     "prefer-stable": true,
     "require": {
         "php": "^8.1",
-        "silverstripe/framework": "^5",
+        "silverstripe/framework": "^5.1",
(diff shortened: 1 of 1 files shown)


GuySartorelli (approved):
LGTM

PR #1195, opened 2024-05-27
NEW Use react-hook-form

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1747

Thing to keep in mind is that elemental inline-editing is put together in a really annoying way because the inline save button `<SaveAction>` is NOT inside the `<InlineEditForm>` which is should be.  This means it's not viable to trigger a "remote submit" from `<SaveAction>` because it's not inside the `<ReactHookFormContext>`

```
<Element> << bit where we want to "remote submit" from
  <ElementContext>
    <Header>
      ...
        <SaveAction> << not in <ReactHookFormContext>
    <Content>
      <InlineEditForm>
        <FormBuilderLoader>
          <ReactHookForm>
            <ReactHookFormContext>
              <Form>
                <FormContext> << form handleSubmit here - thing that needs to be triggered for "remote submit"
                  ...
                    <TextCheckboxGropuField / ShareThing> << elemental component that's inside both <ElementContext> and <ReactHookFormContext>
```


emteknetnz:
This will probably not be completed

PR #1200, opened 2024-05-30
FIX Reload element form when element version changes

GuySartorelli:
Requires https://github.com/silverstripe/silverstripe-admin/pull/1771 to actually fix the bug - but this PR is safe on its own as a patch as seen by CI going green.
If the prop isn't used on the other end (i.e. if someone only updates elemental and not admin) then the bug will still be there, but it won't cause any new problems.

The form will be reloaded when the element is saved/published/unpublished, etc - but only if its version or versioned-state is changing.

## Issue
- https://github.com/silverstripe/silverstripe-linkfield/issues/256


client/src/types/elementType.js
```diff
@@ -6,8 +6,8 @@ const elementType = PropTypes.shape({
   title: PropTypes.string,
   blockSchema: PropTypes.object,
   inlineEditable: PropTypes.bool,
-  published: PropTypes.bool,
-  liveVersion: PropTypes.bool,
+  isPublished: PropTypes.bool,
+  isLiveVersion: PropTypes.bool,
```
GuySartorelli, line 10: sort-of-unrelated refactoring - I noticed these were wrong when I tried to use the old keys but they kept showing `undefined`

client/src/components/ElementEditor/InlineEditForm.js
```diff
@@ -42,17 +43,20 @@ class InlineEditForm extends PureComponent {
...
+    const schemaUrl = loadElementSchemaValue('schemaUrl', element.id);
 
     const formProps = {
       formTag: 'div',
       schemaUrl,
       identifier: 'element',
       refetchSchemaOnMount: !formHasState,
+      // Reload the form any time the element version changes, so other react components
+      // within the form have their state completely recreated
+      // refetchSchemaCriteria: JSON.stringify([element.version, element.isPublished, element.isLiveVersion]),
```
emteknetnz: Shouldn't this be uncommented?

emteknetnz (changes requested):

client/src/components/ElementEditor/InlineEditForm.js
GuySartorelli: How did I... Yes it should lol. I don't remember even commenting it- will fix that.

emteknetnz (changes requested):
I don't think this method is viable unfortunately, with this PR there's now a massive FOUT that looks really bad that wan't there before. Given that LinkField's on elemental is a pretty niche use-case, and inline-publishing itself isn't broken, it's just got the incorrect label until the page is refreshed, I'd consider this new FOUT to be a regression

Also subsequent inline editing and publishing doesn't actually reload the block, so the original issue hasn't been fully fixed

I've provided a video of before and after https://www.youtube.com/watch?v=yNryWP-m0CY

I'd be inclined to just close this one as too hard because I think there needs to be a more advanced solution to resolve this.

Note that upload field has the same issue of incorreclty saying it's draft when it's published, however I don't think we've had many complaints about this so perhaps this isn't that big of a real world issue

GuySartorelli:
Note to self: FOUT is "Flash Of Unstyled Text" (or in this case just flash of badness)

emteknetnz:
A couple of potential ideas for CMS 6 which approach the problem by just not allowing inline publishing:

1. Remove the inline publish/unpublish buttons. It's worth noting that there have been cases of people being frustrated that they cannot remove the publish button which has been a source of confusion to end users
2. Change the inline edit form to a modal, similar to linkfield. This means there would be no publish button and you need to publish the parent (page) to publish the block. Might be issues with scrolling the modal though for blocks with lots of fields.

1's easier and 2's a stretch though if we got scrolling working it would standardise how we 'inline edit' DataObject's and we could do away with a bunch of elemental code thus reducing the maintenance surface

maxime-rainville:
Given that we've received pretty clear indications that allowing publication of individual blocks is not ideal and that we have a card for CMS 6 to remove this action, I'm inclined to agree that fixing this specific bug is not a great use of time.

I think it's worth noting that it's not only LinkField that is affected by this. The UploadField has the same problem. Any relational field you might want to put in an Elemental block will have the same problem.

On the FOUT point, there's a pretty straight forward fix to this problem: don't nuke the current form state right away when the form builder gets invalidated ... instead wait until you get your form schema back. That should provide a nice smooth UI transition between the two state.

GuySartorelli:
I took one look at trying that and while it is conceptually a simple idea, it is not simple to implement. I'm going to say this just isn't worth our time right now.

PR #1201, opened 2024-06-05
ENH Use class name instead of self

emteknetnz:
Issue https://github.com/silverstripe/developer-docs/issues/486

### DIFF
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -477,7 +477,7 @@ JS
             return $this->controller;
         }
 
-        $controllerClass = self::config()->controller_class;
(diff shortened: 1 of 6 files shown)


src/Models/BaseElement.php
```diff
@@ -477,7 +477,7 @@ public function getController()
             return $this->controller;
         }
 
-        $controllerClass = self::config()->controller_class;
+        $controllerClass = static::config()->controller_class;
```
emteknetnz, line 480: This should have been static previously, though self also works seemingly untintuitively. BaseElement::config() behaves differently from BaseElement::config()

GuySartorelli (approved):
LGTM

Issue #1183, opened 2024-05-06
Elemental block validation not triggered on page save if the block hasn't been expanded


GuySartorelli:
If I save the page without expanding the block, none of the validation seems to happen. e.g. with your test blocks from the description, I add the block and immediately save the page - it should trigger validation errors for the missing required fields but it doesn't.

Note that if you expand and collapse the block, the data is included in the page submission and therefore validated against correctly. This is *only* a problem when the block hasn't been expanded at all before clicking the save button.

## Possible solution

We could *always* load the data for the form and render it, but just keep it hidden until the block gets expanded.

This would mean all page submissions would always include all data for all inline-editable blocks.
It would also mean we can do away with the logic we currently have for inline-saving blocks - we currently have logic to render a hidden form when inline-saving blocks (using the `...` menu on the block) if the block wasn't expanded.

This would likely have a performance impact, especially for pages with a lot of blocks. We'd need to measure that.
That performance impact comes in two flavours:
1. Page load time when loading the edit form for the page that owns the blocks
2. Form submission time

maxime-rainville:
We have too many elemental cards in flight right now. Let's have an architecture catch up about them.

Issue #1177, opened 2024-04-24
Block does not open when saving or publishing a blocks page with block validation errors 


emteknetnz:
## Replication steps

Create an inline editable content block that has validation on a field e.g.

```php
    public function validate(ValidationResult $result)
    {
        if ($this->owner->MyField == 'x') {
            $result->addFieldError('MyField', 'MyField cannot be x');
        }
    }
```

Set an invalid value for the field and click page save

Then attempt to inline save the block without first opening the inline edit form

### Expected 

The element form should submit, and when validation errors come back the form should be open and show validation errors

### Actual

The inline form will submit, but the inline form will not display to the user.  If you manually open the inline form you'll see a validation message because it did submit.

### Notes

This seems to happen due to a weird timing issue with the redux-form remote submit happening at the slightly wrong time, causing a duplicate Element component to be rendered which isn't attached to the DOM.

Prior to refactoring done in [this PR](https://github.com/silverstripe/silverstripe-elemental/pull/1176) - the remote submit happened within the SubmitAction.js and things worked as expected. However moving the remote submit to Element.js, for some bizarre reason causes this issue.

The duplicate Element component seems to get created as part of `Element.js::handleFormSchemaSubmitResponse()`. The duplicate Element.js can be observed by adding a ref via useRef() to a new `<div>` element as a child or Element and then console.log() it out.  What seems to happen is that `setPreviewExpanded(true);` is called on the dupilcate Element that's not attached to the DOM, rather than the one that is.

This issue also happens with the publish button - though it wasn't picked up because it's not covered by a behat test.  It also does a remote submit in Element.js (props.submitForm()).

As part fixing this issue, uncomment this behat test back in to individual-block-validation.feature, and also add one for inline publishing
```
Scenario: I can save a closed block after saving a page with a validation error
   When I fill in "x" for "My Field" for block 1
   And I press the "Save" button
   # Page validation error
   Then I should see "MyField cannot be x" in the ".alert.error" element
   When I press the "View actions" button
   And I click on the ".element-editor__actions-save" element
   # Same validation error after page load with inline save
   And I should see "MyField cannot be x" in the ".form__validation-message" element
   Then I fill in "abc" for "My Field" for block 1
   # Ensure react field is filled in before submitting
   And I wait for 1 second
   # Need to save the whole page to stop the alert
   And I press the "Save" button
```

## Acceptance criteria
- [ ] When inline saving an invalid block after page saving the invalid block pops open
[...]

PR #1214, opened 2024-06-24
NEW Inline save all rendered element forms on parent form submit

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-elemental/issues/1215

Requires https://github.com/silverstripe/silverstripe-admin/pull/1792 and https://github.com/silverstripe/silverstripe-behat-extension/pull/277 and https://github.com/silverstripe/silverstripe-frameworktest/pull/193 and https://github.com/silverstripe/silverstripe-framework/pull/11301

**What this PR does**
- When clicking the page save/publish button, a promise is created which waits for all dirty elemental forms to be submitted.
- If any of the inline forms responses fails validation then the promise will resolve(false) and the page form will not submit.
- If the element forms are valid then the promise will resolve (true) then the page form will submit as usual
- The promise is checked in LeftAndMain in [this PR](https://github.com/silverstripe/silverstripe-admin/pull/1792)
- A lot of logic has has been added to let the parent ElementList component know the current state of its child elements if there are unsaved changes, and the last validation result. This determines when and the value the promise will resolve with.


client/src/components/ElementEditor/ElementEditor.js
```diff
@@ -105,21 +101,15 @@ class ElementEditor extends PureComponent {
           dragSpot={dragSpot}
           isDraggingOver={isDraggingOver}
           dragTargetElementId={dragTargetElementId}
+          sharedObject={sharedObject}
         />
         <ElementDragPreview elementTypes={elementTypes} />
-        <input
```
emteknetnz, line 110: This `<input>` was how the elemental data was POSTed along with the rest of the parent page data

src/Forms/ElementalAreaField.php
```diff
@@ -216,85 +216,4 @@ public function setSubmittedValue($value, $data = null)
         // Content comes through as a JSON encoded list through a hidden field.
         return $this->setValue(json_decode($value ?? '', true));
     }
-
```
emteknetnz, line 219: All the code delete here was only use when doing a page/parent form submit.

The public methods aren't API deletions because those methods are defined on ancestor classes

Inline submits go via ElementController.php instead

tests/Behat/features/validation-failure.feature
```diff
@@ -1,52 +0,0 @@
-@javascript
-Feature: Don't lose content when page or block is invalid
```
emteknetnz, line 2: This whole file was testing validation messages after page save. Since we've removed that functionality entirely this feature file is no longer relevant

client/src/components/ElementEditor/Element.js
```diff
@@ -52,17 +74,12 @@ const Element = (props) => {
         captureDraggingState: true,
       });
     }
-    // Check if formSchema state has already been loaded before opening a block
-    // This can happen if there was a validation error on a block after performing a Page save
-    if (props.formStateExists) {
-      setFormHasRendered(true);
-    }
```
GuySartorelli, line 59: Why is this being removed? The comment makes it sound like it was important?

client/src/components/ElementEditor/Element.js
```diff
@@ -503,6 +519,10 @@ Element.propTypes = {
   onDragOver: PropTypes.func, // eslint-disable-line react/no-unused-prop-types
   onDragEnd: PropTypes.func, // eslint-disable-line react/no-unused-prop-types
   onDragStart: PropTypes.func, // eslint-disable-line react/no-unused-prop-types
+  saveElement: PropTypes.bool.isRequired,
+  onBeforeSubmitForm: PropTypes.func.isRequired, // eslint-disable-line react/no-unused-prop-types
+  onAfterSubmitResponse: PropTypes.func.isRequired,
+  increment: PropTypes.number.isRequired,
```
GuySartorelli, line 526:
```suggestion
  // Used to ensure form gets re-rendered on submission so it can be submitted again if there's validation errors
  increment: PropTypes.number.isRequired,
```
The name isn't self-explanatory, so we should explain what this is here for.

client/src/components/ElementEditor/ElementEditor.js
```diff
@@ -128,50 +118,13 @@ ElementEditor.propTypes = {
...
-
-  return filteredElementFormState;
-});
-
-function mapStateToProps(state) {
-  // Memoize form state and value changes
-  const formState = elementFormSelector(state);
-
-  return { formState };
-}
```
GuySartorelli, line 165: I don't want to have to parse all of this code to see what it's doing and try to infer why it was deleted - so I'll just ask. Why is this being deleted? What was it doing?

client/src/components/ElementEditor/ElementList.js
```diff
@@ -11,6 +11,101 @@ import { getDragIndicatorIndex } from 'lib/dragHelpers';
...
+    // Update the sharedObject so that setState() can be called from entwine.js
+    this.props.sharedObject.setState = this.setState.bind(this);
+  }
+
+  componentDidUpdate(prevProps, prevState) {
+    // Number of blocks just changed after graphql response
+    if (this.props.blocks !== prevProps.blocks) {
+      this.resetState(prevState, false);
+      return;
+    }
```
GuySartorelli: Can you please reword this comment? I'm not sure what you mean by it. Is it something to do with adding new or removing existing blocks?
Does the sort order matter?
Why are we resetting the state anyway?
Why do we not do all the rest of the logic below?

client/src/components/ElementEditor/ElementList.js
```diff
@@ -11,6 +11,101 @@ import { getDragIndicatorIndex } from 'lib/dragHelpers';
...
+        this.resetState(prevState, allValid);
+        this.setState({ saveAllElements: false });
+      }
+    }
+  }
+
+  resetState(prevState, resetHasUnsavedChangesBlockIDs) {
+    if (this.props.blocks === null) {
+      return;
+    }
```
GuySartorelli: Why do we not reset state if there's no blocks?

client/src/components/ElementEditor/ElementList.js
```diff
@@ -11,6 +11,101 @@ import { getDragIndicatorIndex } from 'lib/dragHelpers';
...
+    this.props.blocks.forEach(block => {
+      const blockID = parseInt(block.id, 10);
+      if (resetHasUnsavedChangesBlockIDs) {
+        hasUnsavedChangesBlockIDs[blockID] = false;
+      } else if (prevState.hasUnsavedChangesBlockIDs.hasOwnProperty(blockID)) {
+        hasUnsavedChangesBlockIDs[blockID] = prevState.hasUnsavedChangesBlockIDs[blockID];
+      } else {
+        hasUnsavedChangesBlockIDs[blockID] = false;
+      }
+      validBlockIDs[blockID] = null;
```
GuySartorelli, line 77: So we're saying all blocks are "not saved" in valid block IDs, but there's a separate `hasUnsavedChangesBlockIDs` as well? I'm really confused here. Do we have two states both tracking the saved/unsaved state of blocks?
And why does "validBlockIDs" track saved status? Shouldn't it just track include valid IDs and exclude everything else like the name implies?

This might just be a naming thing but this is really confusing me.

client/src/components/ElementEditor/ElementList.js
```diff
@@ -11,6 +11,101 @@ import { getDragIndicatorIndex } from 'lib/dragHelpers';
...
+    });
+    this.setState({ hasUnsavedChangesBlockIDs, validBlockIDs });
+  }
+
+  handleChangeHasUnsavedChanges(elementID, hasUnsavedChanges) {
+    this.setState(prevState => ({
+      hasUnsavedChangesBlockIDs: {
+        ...prevState.hasUnsavedChangesBlockIDs,
+        [elementID]: hasUnsavedChanges,
```
GuySartorelli, line 86: Why is `[elementID]` an array? Is that even valid as a key for an object?

client/src/legacy/ElementEditor/entwine.js
```diff
@@ -64,12 +82,20 @@ jQuery.entwine('ss', ($) => {
       root.render(<ElementEditorComponent {...props} />);
     },
 
+    onmatch() {
+      this.render();
+    },
```
GuySartorelli: It doesn't look like `render()` is being called at any other time? Is that happening in one of the other PRs?

GuySartorelli (changes requested):

GuySartorelli:
Problems encountered when testing locally:

## There's no feedback

1. Open a block
2. Make an edit which makes the block invalid
3. Close the block
4. Click "Save" on the page

There's nothing at all indicating that something went wrong. No toast, and the block doesn't open. It looks like the page save is just broken but in reality there's a validation error I'm not being told about.

There should definitely be a toast message, and ideally the block should open so that I can immediately see the validation error message without opening all the blocks to try find it.

This is partially https://github.com/silverstripe/silverstripe-elemental/issues/1177 which is meant to be fixed by this PR.

## Blocks get double saved - which turns them back to "modified"

1. Have some valid blocks.
2. Click "Publish" on the page. The blocks should all be in a published state.
3. Open one of the blocks. Made _no modifications_.
4. Click "Save" on the page

The block is now "Modified" even though I didn't make any changes! Note that the page itself isn't "modified", so this isn't just how the CMS treats a save that happens after a publish.

## `validate()` method validation still gets confused with fields on the page

1. Implement a `validate()` method on a block which adds this: `$result->addFieldError('Title', 'Nah bro');`
2. Do not open any blocks
3. Click "Publish" on the page (doesn't happen with save)

The "Nah bro" message appears underneath the page title field. This is https://github.com/silverstripe/silverstripe-elemental/issues/1183 which is meant to be fixed by this PR.

client/src/components/ElementEditor/Element.js
emteknetnz: The scenario for this being imporant is when clicking the page/parent save button and there was a validation issue with on the the child elements

This is no longer required as we're now inline saving all elements and not sending any elemental data with the page/parent POST request

client/src/components/ElementEditor/ElementEditor.js
emteknetnz: The whole point of it was to extract the redux state and to it into a prop called `formState`

The only place that prop was used was to populate the now deleted `<input>` element whose purpose was to send elemental data in the POST request when clicking the page/parent save button

client/src/components/ElementEditor/ElementList.js
emteknetnz: Need to wrap in array so that elementID is used as a variable rather than as the actual key:

```js
const key = "3";
const prevState = { "hasUnsavedChangesBlockIDs": {"1":"a","2":"b","3":"c"} };
const someValue = {
  ...prevState.hasUnsavedChangesBlockIDs,
  [key]: "x"
};
console.log(someValue);
// => Object { 1: "a", 2: "b", 3: "x" }

const key = "3";
const prevState = { "hasUnsavedChangesBlockIDs": {"1":"a","2":"b","3":"c"} };
const someValue = {
  ...prevState.hasUnsavedChangesBlockIDs,
  key: "x"
};
console.log(someValue);
// => Object { 1: "a", 2: "b", 3: "c", key: "x" }

```

client/src/legacy/ElementEditor/entwine.js
emteknetnz: Removed

client/src/components/ElementEditor/Element.js
emteknetnz: Updated

client/src/components/ElementEditor/ElementList.js
emteknetnz: Updated

client/src/components/ElementEditor/ElementList.js
emteknetnz: Because we're looping this.props.blocks below, can't loop blocks if they're not there

client/src/components/ElementEditor/ElementList.js
emteknetnz: hasUnsavedChangesBlockIDs is the blocks dirty state - true/false

validBlockIDs is the blocks validation state - true/false/null(i.e. unknown - haven't sent and/or received response from server)

GuySartorelli: Ahh okay. That makes sense after you explained it.... I don't know what we can do about the variable names to make it clearer... can you please add a comment somewhere to that effect so we have that context immediately clear when looking at the code?

client/src/components/ElementEditor/ElementList.js
GuySartorelli: Oh, weird, I hate it lol. Javascript is weird.

client/src/components/ElementEditor/ElementList.js
GuySartorelli: We should ensure `hasUnsavedChangesBlockIDs` and `validBlockIDs` are both empty though, surely?

client/src/components/ElementEditor/ElementList.js
GuySartorelli: What was changed? I still don't know the answers to any of the questions I asked.

emteknetnz (commented):
Need to set a composer requirement of silverstripe/admin ^2.3

GuySartorelli:
Note that I still have the "no feedback" and "Blocks get double saved - which turns them back to "modified"" issues as described above.

Note that the "no feedback"  problem only happens with form validation (i.e. `getCMSCompositeValidator()` - I specifically add a `RequiredFields` validator for the `Title` field). Using the `validate()` method I do get a toast and the block does open as expected.

emteknetnz:
I've solved the "Blocks get double saved - which turns them back to "modified" issue - change [here](https://github.com/silverstripe/silverstripe-admin/pull/1792/files#r1682311498)

For the "no feedback" issue with RequiredFields ... I don't know how to resolve this one without globally removing client-side validation (which admittedly would simplify things, but we probably shouldn't. Though I also won't really mind :-).

The issue is that RequiredFields has client-side (pure js) validation which prevents an XHR request from being sent to the server in the first place. The 'feedback' of popping opening the element happens as part of the XHR response which is formSchema formatted and contains the validation errors.

However with the client-side validation there's nothing "returned" to trigger opening the element. If we did globally remove client-side validation then it'll make an XHR request and fail validation on the server and behave the same UX wise as DataObject->validate() failing

I think the client-side validation is passed in as a callback into redux-form ... somehow ([this maybe?](https://github.com/silverstripe/silverstripe-admin/blob/2/client/src/boot/applyTransforms.js#L68)).  I don't think there's anyway to somehow extract a 'response' from it, particularly as we're remote-submitting the form which does not return a promise

So I guess, options are:
a) Globally disable client-side validation
b) Hack is something kind of nasty that emits an event from Validator.js and we listen for it in Element.js .. somehow ..
c) Just live with the bad UX

GuySartorelli:
> Just live with the bad UX

I don't think that's really an option, since `RequiredFields` is the only validator we offer out-of-the-box so it's presumably used a lot. Not having any feedback at all about why submitting a form did nothing is really bad IMO. If I was a content author and clicking "save" seemed to do literally nothing, I'd be contacting my vendor and reporting a critical bug. If there was either a toast or the block popped open, it wouldn't be so bad.

It's interesting to note that a `RequiredFields` validator for any field _on the page directly_ doesn't have this same behaviour - in that case a toast _is_ triggered.

Is that front-end validation JS able to trigger a toast? That would resolve the problem well enough for now, I think. Better if the block opens, but at least if there's a toast it prompts the content author to start looking for a validation error message as opposed to just thinking something's not working the way it should be.

emteknetnz:
I've the validation fixed it by hardcoding something into admin to disable client-side validation for elemental specifically - https://github.com/silverstripe/silverstripe-admin/pull/1792/files#diff-cc40afd32326a7c9ddea47afd5623f5dd54abfe99db517c1f6e72ca43337a4cbR73

It's obviously not the prettiest, though the UX experience is now much better.

Note that I'm pretty sure this change will require updating a behat test that expects client side validation, so rerun tests for this is the admin PR is merged

tests/Behat/features/edit-block-element.feature
GuySartorelli, line 1: Since `SiteTree` and `CMSMain` save/publish is known to be different from regular `DataObject` and `ModelAdmin` save/publish`, we should add some tests for the validation logic in here, similar to the page save validation feature.

tests/Behat/features/page-save-validation.feature
GuySartorelli, line 1: Since `SiteTree` and `CMSMain` save/publish is known to be different from regular `DataObject` and `ModelAdmin` save/publish, we should add some tests for the validation logic when a non-page object has a/multiple `ElementalArea`.

client/src/components/ElementEditor/ElementList.js
GuySartorelli: @emteknetnz Please respond to this

client/src/components/ElementEditor/ElementList.js
GuySartorelli: @emteknetnz Please respond to this

client/src/components/ElementEditor/ElementList.js
GuySartorelli: @emteknetnz Please respond to this

GuySartorelli:
There are some unresolved comments:
- https://github.com/silverstripe/silverstripe-elemental/pull/1214#discussion_r1688795721
- https://github.com/silverstripe/silverstripe-elemental/pull/1214#discussion_r1688795757
- https://github.com/silverstripe/silverstripe-elemental/pull/1214#discussion_r1688795961

client/src/components/ElementEditor/ElementList.js
emteknetnz: Comment was changed to `// Scenario: Number of blocks just changed after graphql response`

For the other questions .. I don't know / can't remember - It's been weeks since I wrote this.  I guess test it and if it doesn't behave as you'd expect it to then we could change something.

client/src/components/ElementEditor/ElementList.js
emteknetnz: As above - It's been too long since I've wrote this and I've lost context. If it's not working as expected when testing then we should change something, otherwise leave as is

client/src/components/ElementEditor/ElementList.js
emteknetnz: Have added inline comments

client/src/components/ElementEditor/ElementList.js
GuySartorelli: The original comment: "Number of blocks just changed after graphql response"
The new comment: "Scenario: Number of blocks just changed after graphql response"

All you did was put the word "Scenario" in front of it. It still doesn't explain what's happening sufficiently, which is why I have all of those questions - and why you now aren't able to answer them.

I don't know what this is doing because the comment doesn't explain it, so I don't know how to craft a test that will specifically test for the questions I listed above.

client/src/components/ElementEditor/ElementList.js
GuySartorelli: I don't know what edge cases might be caused and therefore don't know how to test them. I think it's worth the time to regain context on this and understand whether we need to fully reset the state here or not so we don't merge something that's broken in ways we don't understand.

emteknetnz: > We should ensure hasUnsavedChangesBlockIDs and validBlockIDs are both empty though, surely?

No, because we're resetting the component state.  Both of those variables are used in the call to setState below.

client/src/components/ElementEditor/ElementList.js
emteknetnz: > Is it something to do with adding new or removing existing blocks?

It's in the componentDidUpdate() method, and we're checking that `(this.props.blocks !== prevProps.blocks) ` means that this.props.blocks changed.

this.props.blocks comes from graphql ... somehow (the graphql implementation is so confusing I actually don't know how it populates the prop)

Hence the comment `"Scenario: Number of blocks just changed after graphql response".  I'll update the comment to say "Scenario: blocks props just changed after a graphql query response updated it" because that's more accurate.

> Does the sort order matter?

I'm not sure quite what this means, though if you drag and drap the elements this it will trigger this code block. There is a graphql request made whenever you drag and drop sort elements.

> Why are we resetting the state anyway?

If we've retched the blocks then we've essentially reset the component, so the state needs to be reset

> Why do we not do all the rest of the logic below?

Because it's for a totally different scenario `// Scenario Saving all elements and state has just updated because of a formSchema response from an inline save - see Element.js handleFormSchemaSubmitResponse()`

This is in the context of the `componentDidUpdate()` method which will trigger in any change in state or props. This is a class component, if it were a functional component I'd be wrapping everything in their own `useEffect()` which would make it more obvious that they're totally separate things

client/src/components/ElementEditor/ElementList.js
GuySartorelli: The call to `setState()` below is skipped by this `return;` so that's not relevant. I'm talking about this condition, where there are no blocks and we are just returning.

We've called `resetState()`, but because there's no blocks, we return without doing anything? Seems like we should still reset these two parts of the state even when there's no blocks.

GuySartorelli (changes requested):

client/src/components/ElementEditor/ElementList.js
emteknetnz: It functionally makes no difference since they're already set to empty objects in the class constructor, and the only time this.props.blocks will be null in when the component is initialised, though I'll make the update anyway just to make it a little clearer

GuySartorelli (approved):
LGTM - will merge the other PRs then rerun CI.

GuySartorelli:
Behat failing

emteknetnz:
Needed to tag 5.4.0 of behat-extension, have done that an re-run, working now

Issue #1215, opened 2024-06-27
Implement inline submitting elements on page/parent save


emteknetnz:
This card is for the implementation of "Approach one" from [SPIKE Look at solutions between field validation conflicting between block level and page level](https://github.com/silverstripe/silverstripe-elemental/issues/1204)

## Issues this resolves
- https://github.com/silverstripe/silverstripe-elemental/issues/1179
- https://github.com/silverstripe/silverstripe-elemental/issues/1177

## Questions:
- Do we want green toast notifications for individual content blocks published when page is published? (e.g. you'll get three green toasts on page save if two elements are saved)
- Do we want red toast notifications for individual content blocks published when they fail validation? (currently there are no red toasts)
- Do we want to block saving individual elements until all elements are valid? (Would be significantly more complex, need validation endpoints)

## Acceptance criteria
- [ ] Elements with a dirty form state are inline saved when the page/parent save/publish button is clicked
- [ ] The page/parent save/publish is blocked when there are validation errors with one or more of the elements that were inline saved
- [ ] Solution works with 
  - [ ] ElementalAreas on Pages 
  - [ ] ElementalAreas on non-page DataObjects
  - [ ] Multiple ElementalAreas on one form
- [ ] Elements are still cascade published on parent publish
- [ ] PHP logic for handling saving elemental data on page/parent save is no longer called and formally deprecated.
- [ ] If viable, stop submitting elemental data on page/parent save with POST request.
 
## Note
- We've thought about adding an opt-out for the new behaviour but decide against it under the assumption people would have had to customise the individual block save anyway.
- Our testing team has offered to help regression test the recent elemental work prior to the beta. Follow up with them once this card is done.

## Shared CI run (elemental endtoend) are the relevant jobs)
- https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/10071368437

Notes:
- The elemental endtoend job is the relevant one
- I cannot replicate any of the other endtoend failures locally
- The phpunit failures are unrelated

## PRs
- https://github.com/silverstripe/silverstripe-frameworktest/pull/193
- https://github.com/silverstripe/silverstripe-behat-extension/pull/277
- https://github.com/silverstripe/silverstripe-framework/pull/11301
- https://github.com/silverstripe/silverstripe-admin/pull/1792
- https://github.com/silverstripe/silverstripe-elemental/pull/1214
- https://github.com/silverstripe/developer-docs/pull/542

## Test PRs
[...]

PR #1219, opened 2024-07-15
MNT Split behat jobs

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/278

--prefer-lowest unit test failure is existing

### DIFF
--- a/tests/Behat/features/add-block-element-to-data-object.feature
+++ b/tests/Behat/features/add-block-element-to-data-object.feature
@@ -1,4 +1,4 @@
-@javascript
+@javascript @retry @job1
 Feature: Add elements in the CMS DataObject
   As a CMS user
   I want to add elements in the CMS DataObject
--- a/tests/Behat/features/add-block-element.feature
+++ b/tests/Behat/features/add-block-element.feature
@@ -1,4 +1,4 @@
-@javascript
+@javascript @retry @job1
 Feature: Add elements in the CMS
   As a CMS user
   I want to add elements in the CMS
--- a/tests/Behat/features/add-link-to-anchor.feature
+++ b/tests/Behat/features/add-link-to-anchor.feature
@@ -1,3 +1,4 @@
+@retry @job1
 Feature: Link to anchors in elements
 As a cms author
 I want to link to anchors in my content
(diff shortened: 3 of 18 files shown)


GuySartorelli (commented):
6 jobs seems excessive - how did you determine which features would be for which job?

Will hold off looking at the others pending a response.

emteknetnz:
I looked at the total run time of the endtoend job in github actions for each module. If it was over 10 minutes then I'd split look to split it so the total run time should be roughly 8 minutes or so e.g. a 24 minute run time would get split to 3x jobs

Elemental endtoend was something like 55 minutes so I split it 6 ways

I split things by simply dividing the total number of features files by the number of jobs I wanted, e.g. if I was going to split into 3 jobs, and there were 14 feature file, then the first 5 would be job1, the next 5 job2, and the last 4 job3

I don't expect things to be perfectly even, some jobs are likely to still be over 10 minutes because they have a couple of very long running feature files. We can make fine adjustments later on if needed

GuySartorelli:
Most of the jobs are around 5 mins, but job 3 is around 13 mins - up to you if you want to rebalance that. If you don't, just say so and I'll merge as-is (assuming that last one that's running right now goes green)

emteknetnz:
Have tried to rebalance, we'll see how it goes

emteknetnz:
Have rebalanced - note that the PHP 8.3 / MySQL 8.0 jobs are probably actual legit failures, not sure why they've only cropped up now. They should be resolved in a PR that is not this one.

Another note - this targets `5.2`, while theres about another 200 steps in `5`, so we'll need to rebalance this again after merge up, probably wait until after https://github.com/silverstripe/silverstripe-elemental/issues/1215 is all merged before doing that

GuySartorelli:
Running CI on `5.2` to see if those failures happen there: https://github.com/silverstripe/silverstripe-elemental/actions/runs/9998847588

GuySartorelli:
That CI run passes behat... so not sure what's causing the failures, but it's new to this PR.

emteknetnz:
Looks like an auto-increment issue with mysql8 - should be fixed with this PR https://github.com/silverstripe/silverstripe-framework/pull/11309

GuySartorelli (approved):
LGTM

PR #1220, opened 2024-07-17
API Remove GraphQL

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1799


src/Controllers/ElementalAreaController.php
```diff
@@ -31,21 +37,317 @@ class ElementalAreaController extends CMSMain
         'elementForm/$ItemID' => 'elementForm',
         'POST api/saveForm/$ID' => 'apiSaveForm',
         '$FormName/field/$FieldName' => 'formAction',
+        // new api
+        'GET readElements/$elementalAreaID!' => 'apiReadElements',
+        'POST add' => 'apiAdd',
```
emteknetnz: these new endpoints should have api/* prefixed, like api/saveForm/$ID does

client/src/state/history/revertToBlockVersionRequest.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import backend from 'lib/Backend';
+import { getConfig } from 'state/editor/elementConfig';
+import Config from 'lib/Config';
+
+// TODO: See if it's viable to move this out of 'state' and 'registerTransforms'
```
GuySartorelli: Don't have TODO comments - either do it now or create a card to do it later.

src/Controllers/ElementalAreaController.php
```diff
@@ -29,23 +33,217 @@ class ElementalAreaController extends CMSMain
...
+        $element = BaseElement::get()->byID($id);
+        if (!$element) {
+            $this->jsonError(400);
+        }
+        // There is no check if the Versioned extension is not applied and then checking
+        // canDelete() instead, because elemental does not officially support unversioned elements
+        if (!$element->canDelete()) {
+            $this->jsonError(403);
+        }
+        $element->doArchive();
```
GuySartorelli:
```suggestion
        if (!$element->canDelete()) {
            $this->jsonError(403);
        }
        // There is no check if the Versioned extension is not applied and then calling
        // delete() instead, because elemental does not officially support unversioned elements
        $element->doArchive();
```
Comment was in the wrong place about the wrong method call.

src/Controllers/ElementalAreaController.php
```diff
@@ -29,23 +33,217 @@ class ElementalAreaController extends CMSMain
...
+        'apiSaveForm',
+        'apiSort',
+        'apiUnpublish',
     ];
 
+    public function apiArchive(HTTPRequest $request): HTTPResponse
+    {
+        $this->checkJsonCsrfToken($request);
+        $id = (int) $request->param('ID');
+        $element = BaseElement::get()->byID($id);
```
GuySartorelli: In `AssetAdmin` there's a lot of validation being done around IDs - shouldn't we be doing the same validation for these?
I'm okay with not doing it, you'll get a 400 ('cause it's not found) for garbage data anyway. Just seems inconsistent to do it there but not here is all.

src/Controllers/ElementalAreaController.php
```diff
@@ -29,23 +33,217 @@ class ElementalAreaController extends CMSMain
...
+        if (!$element) {
+            $this->jsonError(400);
```
GuySartorelli: As I've mentioned in other PRs please add a code comment explaining why it's not 404, if indeed it shouldn't be.

src/Controllers/ElementalAreaController.php
```diff
@@ -29,23 +33,217 @@ class ElementalAreaController extends CMSMain
...
+        }
+        if (!$element->canPublish()) {
+            $this->jsonError(403);
+        }
+        $element->publishRecursive();
+        return $this->jsonSuccess(204);
+    }
+
+    public function apiReadElementalArea(HTTPRequest $request): HTTPResponse
```
GuySartorelli: Should maybe be `apiReadElements` and rename the route to match? We're not actually reading information about the area record, just about the elements inside it.

src/Controllers/ElementalAreaController.php
```diff
@@ -29,23 +33,217 @@ class ElementalAreaController extends CMSMain
...
+        if (!$element->canEdit()) {
+            $this->jsonError(403);
+        }
+        try {
+            $reorderingService = Injector::inst()->create(ReorderElements::class, $element);
+            $reorderingService->reorder($afterBlockID);
+            return $this->jsonSuccess(204);
+        } catch (Exception) {
+            $this->jsonError(500);
+        }
```
GuySartorelli: Shouldn't that just happen anyway? Why do we need to explicitly catch and throw like this?
If this was needed I'd expect to see it in literally all of our actions on all of our controllers.

src/Controllers/ElementalAreaController.php
```diff
@@ -158,6 +355,36 @@ public static function removeNamespacesFromFields(array $data, $elementID)
...
+    }
+
+    // TODO: upgrade to AssesAdmin:: get-post-data
+    private function getPostData()
+    {
+        $request = $this->getRequest();
+        $postData = json_decode($request->getBody() ?? '', true);
+        return $postData;
+    }
```
GuySartorelli: I assume the TODO means you intended to centralise that method - I agree it seems like it should be centralised. This is the third time I've seen this method (or one very similar) being added I think.

src/Models/BaseElement.php
```diff
@@ -1066,28 +1061,13 @@ public static function getBlockConfig()
...
-     *
-     * By default the only action is "edit" which is simply the URL where the block may be edited.
-     *
-     * To modify the actions, either use the extension point or overload the `provideBlockSchema` method.
-     *
-     * @internal This API may change in future. Treat this as a `final` method.
-     * @return array
-     */
-    public function getBlockSchema()
+    public function getBlockSchema(): array
```
GuySartorelli, line 1070: Please make sure there's a PHPDoc that explains what it's for. All public and protected methods should have a PHPDoc description.

src/Models/BaseElement.php
```diff
@@ -1066,28 +1061,13 @@ public static function getBlockConfig()
...
     {
-        $blockSchema = $this->provideBlockSchema();
-
-        $this->extend('updateBlockSchema', $blockSchema);
-
-        return $blockSchema;
+        return $this->provideBlockSchema();
```
GuySartorelli: There's no sense having a method that just returns the value of another method - get rid of one or the other (or put the extension hook back)

src/Models/BaseElement.php
```diff
@@ -1272,15 +1252,8 @@ public function EvenOdd()
      * @return string
      * @deprecated 5.3.0 Will be replaced with getTypeName()
```
```suggestion
```

src/Models/BaseElement.php
```diff
@@ -1272,15 +1252,8 @@ public function EvenOdd()
...
      */
-    public static function getGraphQLTypeName(): string
+    public static function getTypeName(): string
```
GuySartorelli: We now have `getType()`, `getTypeNice()`, and `getTypeName()` - can you please look and see if these can be joined together, and if not, see if they can be renamed to be more clearly distinct?

GuySartorelli (changes requested):

client/src/state/history/revertToBlockVersionRequest.js
emteknetnz: Removed, I won't bother to raise a new card it's an old idea I didn't bother pursuing

src/Controllers/ElementalAreaController.php
emteknetnz: Updated to use new shared method in LeftAndMain

src/Controllers/ElementalAreaController.php
emteknetnz: As discussed offline, should be 400 because it's a POST request

src/Controllers/ElementalAreaController.php
emteknetnz: Updated with a slightly different comment

src/Controllers/ElementalAreaController.php
emteknetnz: Updated

src/Controllers/ElementalAreaController.php
emteknetnz: ReorderElements will throw an exception if $afterBlockID is invalid. ReorderElements doesn't need to be specific to JSON controllers so it would be wrong to go $this->jsonError(500) inside of it.

It makes it easier to consistently unit test this if we throw just jsonError a 500 here. There's a few instances of this so I'll move it to a new private method

src/Models/BaseElement.php
emteknetnz: Reverted it to how it was, not sure why I removed the extension hook, looks like an oversight by me

src/Models/BaseElement.php

src/Models/BaseElement.php
emteknetnz: There's this (kind of awful) concept of an "Element Type" which has many references to it, e.g. [ElementTypeRegistry](https://github.com/silverstripe/silverstripe-elemental/blob/5/src/Services/ElementTypeRegistry.php) and [elementTypeType](https://github.com/silverstripe/silverstripe-elemental/blob/5/client/src/types/elementTypeType.js) :-/ 

"TypeName" is basically elementType.name, or something like that

While I don't like this concept at all, it's also outside the scope of this PR to change this

src/Controllers/ElementalAreaController.php
emteknetnz: Am now using the new method on LeftAndMain

src/Models/BaseElement.php
emteknetnz: Done

src/Controllers/ElementalAreaController.php
GuySartorelli: My point is more that a 500 is a 500. And an exception will always result in a 500.

If we are catching exceptions _at all_ and using `$this->jsonError(500);` in the catch block, then we should do this _everywhere_ we want a json response, because anything could throw an exception we're not expecting.

And we _shouldn't_ do that everywhere because it should be handled globally instead. Which it is already, isn't is?

src/Models/BaseElement.php
GuySartorelli: There's a lot of information in the old doc block that's been removed. Was it actually incorrect?
It seems like we're losing a lot of context here.

GuySartorelli (changes requested):
Sorry, forgot to hit start review. Or rather GitHub didn't give me the option because I was responding in the regular view and not the files view. Hate that.

I only added this comment before noticing my blunder:
- https://github.com/silverstripe/silverstripe-elemental/pull/1220#discussion_r1726340891

src/Controllers/ElementalAreaController.php
emteknetnz: That's true, have removed

src/Models/BaseElement.php
emteknetnz: It was incorrect information

GuySartorelli (approved):
LGTM, will merge when all the other PRs are happy

GuySartorelli (approved):
Other than the rebase, looks like the only change since my last approval was updating yarn lock. Approved.

GuySartorelli (approved):
re-approved most recent changes

PR #1226, opened 2024-07-29
MNT Reuse existing classes for behat test

GuySartorelli:
Reuses existing classes, to replace the ones we're removing in https://github.com/silverstripe/silverstripe-frameworktest/pull/196

## Issue
- https://github.com/silverstripe/.github/issues/287


GuySartorelli:
Found a simpler solution

Issue #1218, opened 2024-07-11
Sorting blocks in draft mode can cause changes on the frontend as it updates the Element_Live table sort value


Jianbinzhu:
### Module version(s) affected

4.11.0

### Description

When re-order the blocks without publishing the page, I would expect no changes to the frontend. So I looked into the sort values in the database and finds out that sorting blocks in draft mode also updates the `Element_Live` table.

### How to reproduce

I have created some dummie data to re-create the bug, before re-ordering

Both `Element` & `Element_Live` table look like this

<img width="232" alt="Screenshot 2024-07-11 at 11 29 48 PM" src="https://github.com/silverstripe/silverstripe-elemental/assets/11606683/1f1ba7ac-97f1-4803-82b7-a0813dd1ebcd">

the following data is after moving content block 8,9,10 up below content block 1.

The `Element` table

```
Select `ID`, `Title`, `Sort` FROM `Element` ORDER BY `Sort` ASC;
```

<img width="236" alt="Screenshot 2024-07-11 at 10 12 24 PM" src="https://github.com/silverstripe/silverstripe-elemental/assets/11606683/515aaa19-3c94-4147-87fa-27126fb6e8ef">

The `Element_Live` table

```
Select `ID`, `Title`, `Sort` FROM `Element_Live` ORDER BY `Sort` ASC;
```

<img width="237" alt="Screenshot 2024-07-11 at 10 22 57 PM" src="https://github.com/silverstripe/silverstripe-elemental/assets/11606683/866ba69e-7f41-462c-8f42-aa8ba430028d">

On the front-end, the content block 7 is under content block 10, because they both have the same sort value on the `Element_Live` table, the next thing it sorts by if they both have the same sort value is the `ID`, in this case content block 7 has a higher `ID` than content block 10.

### Validations

- [X] Check that there isn't already an issue that reports the same bug
- [X] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

### Acceptance criteria
- [x] First validate if the sort order goes straight to live in symbiote gridfieldextensions
- [x] (assuming it doesn't) remove the straight to live logic in CMS 5.3

## PRs
- https://github.com/silverstripe/silverstripe-elemental/pull/1234
- https://github.com/silverstripe/developer-docs/pull/554

emteknetnz:
[...]

PR #1234, opened 2024-08-15
FIX Don't update sort order in live until a block has been published

GuySartorelli:
Uses the same mechanism used in [`SiteTree::onAfterPublish()`](https://github.com/silverstripe/silverstripe-cms/blob/81bfa84f0a43048668df67afa2018f162cc9115f/code/Model/SiteTree.php#L2624-L2636) to update the published sort order of all blocks in the same elemental area.

There are tradeoffs to this which are mentioned in https://github.com/silverstripe/silverstripe-elemental/issues/1218#issuecomment-2290356986 - but IMO this is the best of the options available to us right now.

Targetting `5` because this change might be a bit unexpected for some content authors, so I'd like to give them a heads up in the changelog.

## Issue
- https://github.com/silverstripe/silverstripe-elemental/issues/1218


emteknetnz (approved):
Tested locally, works well

PR #1239, opened 2024-08-19
MNT Temporarily add Deprecation class

emteknetnz:
The Deprecation class was somehow missed from being included merge-up

Have used an MNT prefix because we'll be removing the deprecated cos pretty shortly, though we just need this here for now so that elemental is installable


GuySartorelli (approved):
LGTM

PR #1244, opened 2024-08-22
API Strong typing for the view layer

GuySartorelli:
I want a good base of known types for things that will be going into the template layer before I start abstracting it, and this gets me part of the way there.

## Issue
- https://github.com/silverstripe/silverstripe-framework/issues/11322

### DIFF
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -600,10 +600,8 @@ JS
      * Default way to render element in templates. Note that all blocks should
      * be rendered through their {@link ElementController} class as this
      * contains the holder styles.
-     *
-     * @return string|null HTML
      */
-    public function forTemplate($holder = true)
+    public function forTemplate($holder = true): string
     {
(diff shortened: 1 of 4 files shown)


src/Models/BaseElement.php
```diff
@@ -600,18 +600,16 @@ private function getFieldNamesExcludedFromCmsSearch(): array
...
+    public function forTemplate($holder = true): string
     {
         $templates = $this->getRenderTemplates();
 
         if ($templates) {
             return $this->renderWith($templates);
         }
 
-        return null;
+        return '';
```
GuySartorelli, line 612: There's no parent class enforcing non-nullable values yet - but part of the next piece of work will include adding `forTemplate()` directly on `ViewableData` where its return type will be `string`.
There's just no reason for this method to ever return `null`, in comparison with an empty string.

emteknetnz (approved):

PR #1260, opened 2024-10-14
MNT Update test to reflect changes to template/model layers

GuySartorelli:
Relies and reflects on changes in https://github.com/silverstripe/silverstripe-framework/pull/11405

## Issue
- https://github.com/silverstripe/silverstripe-framework/issues/11322

### DIFF
--- a/tests/ElementalAreaTest.php
+++ b/tests/ElementalAreaTest.php
@@ -78,8 +78,8 @@ class ElementalAreaTest extends SapphireTest
         // OwnerClassName set
         $ownerpage2 = $area2->getOwnerPage();
 
-        $this->assertEquals("DNADesign\Elemental\Tests\Src\TestPage", $ownerpage1);
-        $this->assertEquals("DNADesign\Elemental\Tests\Src\TestPage", $ownerpage2);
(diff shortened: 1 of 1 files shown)


tests/ElementalAreaTest.php
```diff
@@ -78,8 +78,8 @@ public function testGetOwnerPage()
         // OwnerClassName set
         $ownerpage2 = $area2->getOwnerPage();
 
-        $this->assertEquals("DNADesign\Elemental\Tests\Src\TestPage", $ownerpage1);
-        $this->assertEquals("DNADesign\Elemental\Tests\Src\TestPage", $ownerpage2);
+        $this->assertInstanceOf(TestPage::class, $ownerpage1);
+        $this->assertInstanceOf(TestPage::class, $ownerpage2);
```
GuySartorelli, line 82: This was always just a bad test that relied on `ModelData::__toString()` returning the class name.
This change isn't _necessary_ anymore since I had to revert the change to what that method returns, but I think _this_ change is still objectively good and worth doing. Check the class to check the class, ya know?

emteknetnz (approved):

Issue #1265, opened 2024-11-04
Unable to remove last UploadField item in inline editing mode


lozcalver:
### Module version(s) affected

5.3.0

### Description

When inline-editing an item that contains an UploadField, it’s not possible to remove the last item from the UploadField. It appears to be removed, but after saving & refreshing the page it re-appears. Removing other items works as expected, but removing the final item triggers the issue. Affects both multi-upload fields and single item ones.

https://github.com/user-attachments/assets/4c76be52-b73b-4dea-86da-d94abf73c9c2

### How to reproduce

1. Upload any image(s) to an upkoad field (not clear if it has to be images specifically, haven't tested)
1. Publish the block
1. Refresh the page (may not be necessary)
1. Remove the file(s) from the block
1. Publish the block
1. Refresh the page

<details><summary>Expand to see example elemental block with upload fields</summary>

```php
<?php

declare(strict_types=1);

namespace App\Model\Elements;

use DNADesign\Elemental\Models\BaseElement;
use SilverStripe\AssetAdmin\Forms\UploadField;
use SilverStripe\Assets\Image;
use SilverStripe\Forms\FieldList;

class ElementLogos extends BaseElement
{
    private static string $table_name = 'ElementLogos';

    private static array $has_one = [
        'Test' => Image::class,
    ];

    private static array $many_many = [
        'Logos' => Image::class,
    ];

    private static array $many_many_extraFields = [
        'Logos' => [
            'SortOrder' => 'Int',
        ],
    ];

    private static array $owns = [
        'Test',
        'Logos',
    ];

    private static array $cascade_deletes = [
        'Test',
        'Logos',
    ];

    private static array $cascade_duplicates = [
        'Test',
        'Logos',
    ];

    private static string $singular_name = 'logos block';

    private static string $plural_name = 'logo blocks';

    private static string $description = 'Logos block';

    private static string $icon = 'font-icon-block-carousel';

    public function getCMSFields(): FieldList
    {
        $this->beforeUpdateCMSFields(
            function (FieldList $fields) {
                $fields->removeByName(
                    [
                        'Test',
                        'Logos',
                    ]
                );

                $fields->addFieldsToTab(
                    'Root.Main',
                    [
                        UploadField::create('Test', 'Test')
                            ->setAllowedFileCategories(IMAGE_SUPPORTED)
                            ->setFolderName('logos'),
                        UploadField::create('Logos', 'Logos')
[...]

PR #1267, opened 2024-11-06
FIX Ensure files can be removed from elemental blocks

GuySartorelli:
In CMS 5.2 and earlier, _all_ fields in the form are included in the POST data, including empty upload fields.

In CMS 5.3 the formschema form submission logic is used instead, which omits data for empty upload fields (and possibly for other empty fields as well).

Normally this would be handled just fine with `$form->saveInto($record)`, but elemental doesn't use that so missing data just gets ignored instead of being set to null.

## CI note

The new behat test relies on a test extension added in https://github.com/silverstripe/silverstripe-frameworktest/pull/210
That PR needs to be merged before this can go green.

## Issue
- https://github.com/silverstripe/silverstripe-elemental/issues/1265

### DIFF
--- a/src/Controllers/ElementalAreaController.php
+++ b/src/Controllers/ElementalAreaController.php
@@ -138,7 +138,9 @@ class ElementalAreaController extends CMSMain
         // Remove the namespace prefixes that were added by EditFormFactory
         $dataWithoutNamespaces = static::removeNamespacesFromFields($data, $element->ID);
 
-        // Update and write the data object which will trigger model validation
+        // Update and write the data object which will trigger model validation.
+        // Would usually be handled by $form->saveInto($element) but since the field names
+        // in the form have been namespaced, we need to handle it ourselves.
         $element->updateFromFormData($dataWithoutNamespaces);
         if ($element->isChanged()) {
             try {
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -656,15 +656,9 @@ JS
      */
     public function updateFromFormData($data)
     {
-        $cmsFields = $this->getCMSFields();
-
-        foreach ($data as $field => $datum) {
-            $field = $cmsFields->dataFieldByName($field);
-
-            if (!$field) {
-                continue;
-            }
-
+        $cmsFields = $this->getCMSFields()->saveableFields();
+        foreach ($cmsFields as $fieldName => $field) {
+            $datum = $data[$fieldName] ?? null;
             $field->setSubmittedValue($datum);
             $field->saveInto($this);
         }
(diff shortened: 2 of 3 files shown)


src/Controllers/ElementalAreaController.php
```diff
@@ -138,7 +138,9 @@ public function save(array $data, Form $form): HTTPResponse
         // Remove the namespace prefixes that were added by EditFormFactory
         $dataWithoutNamespaces = static::removeNamespacesFromFields($data, $element->ID);
 
-        // Update and write the data object which will trigger model validation
+        // Update and write the data object which will trigger model validation.
+        // Would usually be handled by $form->saveInto($element) but since the field names
+        // in the form have been namespaced, we need to handle it ourselves.
         $element->updateFromFormData($dataWithoutNamespaces);
```
GuySartorelli, line 144: Really we should just rename the fields _in the form_, and then call `$form->saveInto($element)` - but doing that now would be an API break because someone may have customised the `updateFromFormData()` method in their custom elements.

src/Models/BaseElement.php
```diff
@@ -656,15 +656,9 @@ public function getRenderTemplates($suffix = '')
...
-        $cmsFields = $this->getCMSFields();
-
-        foreach ($data as $field => $datum) {
-            $field = $cmsFields->dataFieldByName($field);
-
-            if (!$field) {
-                continue;
-            }
-
+        $cmsFields = $this->getCMSFields()->saveableFields();
```
GuySartorelli, line 659: Getting the `saveableFields()` explicitly - things like `LiteralField` don't need to be touched here.

src/Models/BaseElement.php
```diff
@@ -656,15 +656,9 @@ public function getRenderTemplates($suffix = '')
...
+        foreach ($cmsFields as $fieldName => $field) {
+            $datum = $data[$fieldName] ?? null;
```
GuySartorelli, line 661: If there's data, set to the value in the data - otherwise set to null.

This is still skipping a bunch of stuff that `$form->saveInto()` would do, but fully replicating that logic isn't in scope here. This change is the least disruptive change required to fix the bug.

tests/Behat/features/file-upload.feature
```diff
@@ -0,0 +1,45 @@
+@retry @job2 @gsat
```
emteknetnz:
```suggestion
@retry @job2
```

emteknetnz (changes requested):
Code change looks good

I've merged the frameworktest PR and re-rerun behat and it all works

One tiny change

tests/Behat/features/file-upload.feature
GuySartorelli: Nuts, I remember thinking "I must remember to remove that" lol
Done

emteknetnz (approved):

Issue #1268, opened 2024-11-06
Form submission doesn't go through `Form::saveInto()`


GuySartorelli:
https://github.com/silverstripe/silverstripe-elemental/issues/1265 highlighted that the data submitted through the elemental area form doesn't get saved into elements using the standard `Form::saveInto()`. Instead, similar (but less complex) logic is re-implemented inside `BaseElement::updateFromFormData()`.

We should rely on the standard form submission flow as much as possible, which means removing `BaseElement::updateFromFormData()` in favour of `Form::saveInto()`

## Acceptance Criteria
- [x] Form submissions for elemental blocks get saved into elements using `Form::saveInto()` instead of custom logic

## CMS 5 PRs

- https://github.com/silverstripe/silverstripe-elemental/pull/1270
- https://github.com/silverstripe/developer-docs/pull/624

After merging, reassign to Guy to rebase the CMS 6 PR.

## CMS 6 PRs

- https://github.com/silverstripe/silverstripe-elemental/pull/1269

PR #1269, opened 2024-11-06
API Remove custom logic in favour of Form::saveInto()

GuySartorelli:
I don't think this warrants a note in the changelog beyond what will be automatically generated for the removed API - though feel free to disagree and if so I'll add a docs PR.

## Issue
- https://github.com/silverstripe/silverstripe-elemental/issues/1268

### DIFF
--- a/src/Controllers/ElementalAreaController.php
+++ b/src/Controllers/ElementalAreaController.php
@@ -298,8 +298,7 @@ class ElementalAreaController extends CMSMain
     }
 
     /**
-     * Arrive here from FormRequestHandler::httpSubmission() during a POST request to
-     * /admin/linkfield/linkForm/<LinkID>
+     * Arrive here from FormRequestHandler::httpSubmission() during a POST request.
      * The 'save' method is called because it is the FormAction set on the Form
      */
     public function save(array $data, Form $form): HTTPResponse
@@ -331,13 +330,15 @@ class ElementalAreaController extends CMSMain
             $this->jsonError(403);
         }
 
-        // Remove the namespace prefixes that were added by EditFormFactory
-        $dataWithoutNamespaces = static::removeNamespacesFromFields($data, $element->ID);
+        // Remove the namespace prefixes
+        $factory = Injector::inst()->get(EditFormFactory::class);
+        $factory->removeNamespaceFromFields($form->Fields(), ['Record' => $element]);
+        // Update the record
+        $form->saveInto($element);
+        // Add the namespace prefixes back - this is necessary for the response to be handled correctly
+        $factory->namespaceFields($form->Fields(), ['Record' => $element]);
 
-        // Update and write the data object which will trigger model validation.
-        // Would usually be handled by $form->saveInto($element) but since the field names
-        // in the form have been namespaced, we need to handle it ourselves.
-        $element->updateFromFormData($dataWithoutNamespaces);
+        // Write the data object which will trigger model validation
         if ($element->isChanged()) {
             try {
                 $element->write();
@@ -352,31 +353,6 @@ class ElementalAreaController extends CMSMain
         $response = $this->getSchemaResponse($schemaID, $form);
 
         return $response;
-    }
-
-    /**
-     * Remove the pseudo namespaces that were added to form fields by the form factory
-     *
-     * @param array $data
-     * @param int $elementID
-     * @return array
-     * @deprecated 5.4.0 Will be removed without equivalent functionality to replace it.
-     */
-    public static function removeNamespacesFromFields(array $data, $elementID)
-    {
-        Deprecation::noticeWithNoReplacment('5.4.0');
-        $output = [];
-        $template = sprintf(EditFormFactory::FIELD_NAMESPACE_TEMPLATE, $elementID, '');
-        foreach ($data as $key => $value) {
-            // Only look at fields that match the namespace template
-            if (substr($key ?? '', 0, strlen($template ?? '')) !== $template) {
-                continue;
-            }
-
-            $fieldName = substr($key ?? '', strlen($template ?? ''));
-            $output[$fieldName] = $value;
-        }
-        return $output;
     }
 
     private function reorderElements(BaseElement $element, int $afterElementID): void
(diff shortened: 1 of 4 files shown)


src/Controllers/ElementalAreaController.php
```diff
@@ -354,25 +353,19 @@ public function save(array $data, Form $form): HTTPResponse
 
     /**
      * Remove the pseudo namespaces that were added to form fields by the form factory
-     *
-     * @param array $data
-     * @param int $elementID
-     * @return array
      */
-    public static function removeNamespacesFromFields(array $data, $elementID)
+    private function removeNamespacesFromFields(Form $form, int $elementID): void
```
GuySartorelli, line 357: This method wasn't used anywhere else, so no need for it to be public nor static.

src/Controllers/ElementalAreaController.php
```diff
@@ -298,8 +298,7 @@ public function getElementForm(): Form
     }
 
     /**
-     * Arrive here from FormRequestHandler::httpSubmission() during a POST request to
-     * /admin/linkfield/linkForm/<LinkID>
+     * Arrive here from FormRequestHandler::httpSubmission() during a POST request.
```
GuySartorelli, line 301: Unrelated change here to remove copy/paste error that references linkfield

src/Forms/EditFormFactory.php
```diff
@@ -103,4 +100,22 @@ protected function namespaceFields(FieldList $fields, array $context)
             $field->setName($namespacedName);
         }
     }
+
+    /**
+     * Remove the pseudo namespaces that were added in namespaceFields()
+     */
+    public function removeNamespaceFromFields(FieldList $fields, array $context): void
```
GuySartorelli, line 107: Using the same method signature as the corresponding `namespaceFields()` since they're essentially the reverse of one another.

emteknetnz, line 107: Add a unit test

src/Controllers/ElementalAreaController.php
```diff
@@ -331,11 +330,15 @@ public function save(array $data, Form $form): HTTPResponse
             $this->jsonError(403);
         }
 
-        // Remove the namespace prefixes that were added by EditFormFactory
-        $dataWithoutNamespaces = static::removeNamespacesFromFields($data, $element->ID);
+        // Remove the namespace prefixes
+        $factory = Injector::inst()->get(EditFormFactory::class);
+        $factory->removeNamespaceFromFields($form->Fields(), ['Record' => $element]);
```
emteknetnz, line 335: The way these methods are being called on a factory class (whose purpose to a to create objects) looks weird as they're essentially utility methods.  Change them to be statics and don't instantiate the factory, or just move them both to a standalone utility class as statics

emteknetnz (changes requested):

src/Controllers/ElementalAreaController.php
GuySartorelli: I don't want to do that, because the factory is injected and the existing `namespaceFields()` method could be altered in a project-specific customisation. Breaking that customisation option for the sake of "it looks weird" doesn't sit well with me.

The namespacing is intrinsic to the building of this specific form, so it makes sense for the factory to own that process.

GuySartorelli:
None of the things you've pointed out in this review block the CMS 5 PRs. Please merge the CMS 5 PRs.

src/Forms/EditFormFactory.php
GuySartorelli: There's no test for `namespaceFields()` nor for the old `removeNamespacesFromFields()` method on `ElementalAreaController`... these methods are both indirectly tested already.
But to avoid ping pong, I'll add tests for both of these methods after rebasing on top of the CMS 5 PR.

src/Controllers/ElementalAreaController.php
emteknetnz: Looking at this again, renaming the fields on an existing form twice in order to save it seems kind of wrong

Wouldn't it make more sense to just get the factory to create an entirely new form without the namespaces, load the submitted $data into that and then call $newFormWithoutNamespaces->saveInto($element)

That would allow us to keep keep namespaceFields() as a protected method and we wouldn't even need removeNamespacesFromFields()

**Update: ignore this, just occured to me the the submitted data will include namespaces**

emteknetnz (changes requested):

src/Forms/EditFormFactory.php
GuySartorelli: Done

tests/Forms/EditFormFactoryTest.php
```diff
@@ -4,15 +4,18 @@
...
+use SilverStripe\Forms\LiteralField;
+use SilverStripe\Forms\TextField;
 
 class EditFormFactoryTest extends SapphireTest
 {
     protected static $fixture_file = 'EditFormFactoryTest.yml';
 
     public function testFormFieldsHaveNamespaces()
     {
-        /** @var ElementContent $record */
```
GuySartorelli, line 15: Unrelated, but that's not needed anymore due to the addition of generics

tests/Forms/EditFormFactoryTest.php
```diff
@@ -21,4 +24,42 @@ public function testFormFieldsHaveNamespaces()
...
         $this->assertNotNull($fields->dataFieldByName('PageElements_' . $record->ID . '_Title'));
     }
+
+    public function testNamespaceFields(): void
+    {
+        $record = $this->objFromFixture(ElementContent::class, 'content_block');
+        $factory = new EditFormFactory();
+        $fields = new FieldList([
+            new TextField('FieldOne'),
+            new LiteralField('IgnoredField', ''),
```
GuySartorelli, line 34: `LiteralField` is ignored because it's not a "data field" so it doesn't affect form submission

tests/Forms/EditFormFactoryTest.php
```diff
@@ -21,4 +24,42 @@ public function testFormFieldsHaveNamespaces()
...
+        $this->assertSame($expectedNames, $fields->column('Name'));
+    }
+
+    public function testRemoveNamespaceFromFields(): void
+    {
+        $record = $this->objFromFixture(ElementContent::class, 'content_block');
+        $factory = new EditFormFactory();
+        $fields = new FieldList([
+            new TextField('PageElements_' . $record->ID . '_FieldOne'),
+            new LiteralField('IgnoredField', ''),
```
GuySartorelli, line 53: This is how the literal field will look when passed in, so we're just including it here to validate it doesn't get mangled

emteknetnz (approved):

PR #1271, opened 2024-11-07
API Update API to reflect changes in silverstripe/framework

GuySartorelli:
Replaces use of the overly-broadly named `description` configuration in favour of the new `class_description` config added in https://github.com/silverstripe/silverstripe-framework/pull/11460 and uses `i18n_classDescription` instead of custom i18n code.

Also updates status flag rendering to render any flags in `getStatusFlags()`.

Ends up looking something like this (custom flag added by overriding `getStatusFlags()`)
Screenshot from 2024-11-20 14-56-56

## Issue
- https://github.com/silverstripe/silverstripe-cms/issues/2947


client/src/components/ElementEditor/Header.scss
```diff
@@ -29,8 +29,6 @@
     max-width: calc(100% - 60px);
 
     .badge {
-      color: $state-draft;
-      background-color: $state-modified-bg;
```
GuySartorelli, line 33: Rely on silverstripe/admin styling for colours

client/src/components/ElementEditor/Header.js
```diff
@@ -209,7 +195,7 @@ class Header extends Component {
             </Tooltip>}
           </div>
           <h3 className={titleClasses}>{title}</h3>
-          {this.renderStatusBadge()}
+          {this.renderStatusFlagBadges()}
```
GuySartorelli, line 198: We're rendering potentially multiple badges now, so I've updated the logic to handle that and to pull the status flags from `BaseElement`.

src/Controllers/ElementalAreaController.php
```diff
@@ -205,6 +205,7 @@ public function apiReadElements(HTTPRequest $request): HTTPResponse
                 'canPublish' => $element->canPublish(),
                 'canUnpublish' => $element->canUnpublish(),
                 'canCreate' => $element->canCreate(),
+                'statusFlags' => $element->getStatusFlags(),
```
GuySartorelli, line 208: Pulls through status flags for use in javascript

emteknetnz (approved):

PR #1273, opened 2024-11-12
API Make ElementalAreaController a subclass of FormSchemaController

GuySartorelli:
Needs https://github.com/silverstripe/silverstripe-admin/pull/1850 for CI to go green. Please merge that first.

## Issue
- https://github.com/silverstripe/silverstripe-admin/issues/1762

### DIFF
--- a/src/Controllers/ElementalAreaController.php
+++ b/src/Controllers/ElementalAreaController.php
@@ -5,7 +5,6 @@ namespace DNADesign\Elemental\Controllers;
 use DNADesign\Elemental\Forms\EditFormFactory;
 use DNADesign\Elemental\Models\BaseElement;
 use DNADesign\Elemental\Services\ElementTypeRegistry;
-use SilverStripe\CMS\Controllers\CMSMain;
 use SilverStripe\Control\HTTPResponse;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Forms\Form;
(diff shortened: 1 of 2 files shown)


src/Controllers/ElementalAreaController.php
```diff
@@ -20,18 +19,17 @@
 use Exception;
 use SilverStripe\Control\HTTPRequest;
 use InvalidArgumentException;
+use SilverStripe\Admin\FormSchemaController;
 
 /**
  * Controller for "ElementalArea" - handles loading and saving of in-line edit forms in an elemental area in admin
  */
-class ElementalAreaController extends CMSMain
+class ElementalAreaController extends FormSchemaController
```
GuySartorelli, line 27: This was originally changed from subclassing `LeftAndMain` to `CMSMain` in https://github.com/silverstripe/silverstripe-elemental/pull/449

The reasoning was
> It makes more sense for ElementalAreaController to inherit from CMSMain rather than LeftAndMain, as this means that permissions are checked in relation to the Page Level rather than for the entire CMS.

Maybe that made sense at the time, but you can have an elemental area on any `DataObject` including inside a `ModelAdmin`.... so requiring `CMS_ACCESS_CMSMain` permissions is just wrong. Instead, require CMS access more generally, and let the model permissions dictate what can and can't be done within the controller.

tests/Controllers/ElementalAreaControllerTest.php
```diff
@@ -180,9 +180,9 @@ public function testElementFormPost(string $idType, string $dataType, string $fa
         $response = $this->post($url, $data, $headers);
         $this->assertSame($expectedCode, $response->getStatusCode());
         if ($fail === 'csrf-token') {
-            // Will end up at an HTML page with "Silverstripe - Bad Request"
-            $this->assertSame('text/html; charset=utf-8', $response->getHeader('Content-type'));
-            $this->assertStringContainsString('Silverstripe - Bad Request', $response->getBody());
+            // Gives suitable error message for XHR request
+            $this->assertStringStartsWith('text/plain', $response->getHeader('Content-type'));
+            $this->assertStringContainsString('There seems to have been a technical problem', $response->getBody());
```
GuySartorelli, line 185: The modal links used to be routed through `LeftAndMain` which meant the error handling in that class was used. For some reason they weren't picked up as XHR requests so the full HTML error page was being rendered which wasn't really the correct behaviour, but it's what happened.

Now, a simple `400` response (the default one produced when a form's CSRF token is missing) is produced and isn't altered. This is the correct behaviour.

The end result for the user is no change.

emteknetnz (approved):

Issue #1290, opened 2024-12-06
JS-Error blocks Fluent actions for unlocalised objects


MLindenhofer:
### Module version(s) affected

5.3.2

### Description

Since upgrading to 5.3.x any page or object which is not localized can not be published, drafted. When checking the console there was this error message: 

Uncaught TypeError: Cannot read properties of undefined (reading 'elemental-area-id')
    at e.fn.init.onbeforesubmitform (bundle.js?m=1725229120:1:52032)
    at e.fn.init.<anonymous> (vendor.js?m=1731968536:1:111017)
    at o (vendor.js?m=1731968536:1:102610)
    at e.fn.init.from .cms-edit-form_beforesubmitform (vendor.js?m=1731968536:1:102806)
    at HTMLDocument.<anonymous> (vendor.js?m=1731968536:1:110895)
    at HTMLDocument.dispatch (vendor.js?m=1731968536:1:741792)
    at m.handle (vendor.js?m=1731968536:1:739783)
    at Object.trigger (vendor.js?m=1731968536:1:771793)
    at HTMLFormElement.<anonymous> (vendor.js?m=1731968536:1:772391)
    at Function.each (vendor.js?m=1731968536:1:705023)

### How to reproduce

- Install Fluent module
- Apply ElementalPageExtension to Page.
- Create a new locale
- Try to hit "Copy & publish" - cms action button

Behaviour:
Nothing happens, only that error-msg in console shows up

### Validations

- [X] Check that there isn't already an issue that reports the same bug
- [X] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## PRs
- https://github.com/silverstripe/silverstripe-elemental/pull/1292

emteknetnz:
Linked PR has been merged, a patch released will automatically tagged shortly

lerni:
Pretty cool, it got fixes so quickly. Thank you!

PR #1292, opened 2024-12-11
FIX Don't try to save blocks in a readonly elemental area

GuySartorelli:
Turns out when the field is read-only we don't have the schema data - but even if we did, we shouldn't try to save data for a read-only field.

Note this fixes two separate issues - please test both of them to ensure the fix works as expected in both scenarios.

## Issues
- https://github.com/silverstripe/silverstripe-elemental/issues/1290
- https://github.com/silverstripe/silverstripe-versioned-admin/issues/378


emteknetnz (approved):

emteknetnz:
Tested both scenarios, both are fixed by this PR

andreaspiening:
Hi Guy,

this mr fixed an issue I had with the combination of elemental and advanced-workflow, where clicking the approve button failed with a JS error in the `onbeforesubmitform` hook, due to the form being readonly.

However I wonder if the `onmatch` hook needs the same treatment: it is triggering the same problem (accessing the the `data-schema` attribute, which is absent in the readonly form), hence showing an error in the console, though not having any negative impact on submitting of the form.

Andy

GuySartorelli:
> I wonder if the `onmatch` hook needs the same treatment: it is triggering the same problem (accessing the the `data-schema` attribute, which is absent in the readonly form), hence showing an error in the console, though not having any negative impact on submitting of the form.

If there's no actual problem other than an error in the console log, then yes technically it is a bug that should be fixed but the priority is too low for me to personally create a PR for it.

PR #1296, opened 2024-12-16
MNT Fix behat test

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/349

Fixes https://github.com/silverstripe/silverstripe-elemental/actions/runs/12342587285/job/34442697226#step:12:258

    Then I click "MyPage" in the "#Form_EditForm" element                                # SilverStripe\BehatExtension\Context\BasicContext::iClickInTheElement()
      "MyPage" not found

Broke because of the new 'Archived' badge showing in the gridfield

PR #1297, opened 2024-12-16
ENH Convert ElementList into a functional component

GuySartorelli:
This is needed to allow the use of hooks for dnd-kit.

This PR should change nothing about the user experience or functionality of the module, it should be a perfect like-for-like comparison to `6.0` without this PR.

Test everything you can think to test - sorting, saving, publishing, archiving, etc both from a page and an element level to check everything works the way it used to.

## Issue
- https://github.com/silverstripe/.github/issues/343


GuySartorelli:
behat failure will be fixed by https://github.com/silverstripe/silverstripe-elemental/pull/1296 separately

client/src/legacy/ElementEditor/entwine.js
```diff
@@ -116,12 +117,10 @@ jQuery.entwine('ss', ($) => {
...
         this.setIncrement(increment);
         sharedObject.entwineResolve = entwineResolve;
-        // setState() is bound in the constructor of the ElementList react component
-        // setting saveAllElementst to true will trigger a re-render in the react component
-        sharedObject.setState({
-          saveAllElements: true,
-          increment
-        });
+        // These state setting methods are set by the ElementList react component
+        // setting saveAllElementst state to true will trigger a re-render in the react component
```
emteknetnz:
```suggestion
        // setting saveAllElements state to true will trigger a re-render in the react component
```

emteknetnz (changes requested):
It appears to do the things correctly

One tiny change

client/src/legacy/ElementEditor/entwine.js
GuySartorelli: done

emteknetnz (approved):

PR #1298, opened 2024-12-16
DEP Replace react-dnd with dnd-kit

GuySartorelli:
This PR replaces the `react-dnd` with `dnd-kit` with like-for-like functionality, though there is a visual change.

When dragging, the element collapses and the entire element is moved along with the mouse, giving a preview of where the element will be after sorting. This is the same behaviour with other sorting in the CMS (e.g. linkfield, and the `GridFieldOrderableRows` component in gridfieldextensions).

I haven't implemented sorting with the keyboard like linkfield has because that was non-trivial and isn't functionality that was already available for elemental blocks. Improving accessibility to include sorting by keyboard would be a good next step in a separate card and could be accomplished in a minor release.

## Issue
- https://github.com/silverstripe/.github/issues/343


emteknetnz (approved):
Works well

PR #1306, opened 2025-01-14
API Rename FormField Value to getFormattedValue

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/11468

### DIFF
--- a/src/Forms/TextCheckboxGroupField.php
+++ b/src/Forms/TextCheckboxGroupField.php
@@ -93,7 +93,7 @@ class TextCheckboxGroupField extends CompositeField
             if ($titleField) {
                 $field->replaceField('Title', LiteralField::create(
                     'Title',
-                    $titleField->Value()
+                    $titleField->getFormattedValue()
                 ));
             }
 
@@ -103,7 +103,8 @@ class TextCheckboxGroupField extends CompositeField
             if ($showTitle) {
                 $field->replaceField('ShowTitle', LiteralField::create(
                     'ShowTitle',
-                    $showTitle->Value() === 'Yes' ? $displayedText : $notDisplayedText
+                    // Intentionally using getValue() instead of getFormattedValue() to get the raw value
+                    $showTitle->getValue() === 1 ? $displayedText : $notDisplayedText
                 )->addExtraClass('show-title'));
             }
         }


src/Controllers/ElementalAreaController.php
```diff
@@ -456,10 +456,10 @@ private function createValidationExceptionWithNamespaces(
             $messageCast = $message['messageCast'] ?? ValidationResult::CAST_TEXT;
             if ($messageFieldName) {
                 $fieldName = sprintf(EditFormFactory::FIELD_NAMESPACE_TEMPLATE, $element->ID, $messageFieldName);
-                $params = [$fieldName, $messageText, $messageType, null, $messageCast];
+                $params = [$fieldName, $messageText, $messageType, '', $messageCast];
                 $validationResultWithNameSpaces->addFieldError(...$params);
             } else {
-                $validationResultWithNameSpaces->addError($messageText, $messageType, null, $messageCast);
+                $validationResultWithNameSpaces->addError($messageText, $messageType, '', $messageCast);
```
GuySartorelli: This doesn't seem relevant? Looks like https://github.com/silverstripe/silverstripe-elemental/pull/1311 accidentally snuck into this PR?

src/Forms/TextCheckboxGroupField.php
```diff
@@ -103,7 +103,8 @@ public function performReadonlyTransformation()
             if ($showTitle) {
                 $field->replaceField('ShowTitle', LiteralField::create(
                     'ShowTitle',
-                    $showTitle->Value() === 'Yes' ? $displayedText : $notDisplayedText
+                    // Intentionally using getValue() instead of getFormattedValue() to get the raw value
+                    $showTitle->getValue() === 'Yes' ? $displayedText : $notDisplayedText
```
GuySartorelli: Why do we want to check against the raw value here? `ShowTitle` is a `CheckboxField` which will by default use ints `1` and `0` (but probably allows other value types) so this condition will often fail.

GuySartorelli (changes requested):

src/Controllers/ElementalAreaController.php
emteknetnz: Probably makes sense to block this PR until the other one is merged, then rebase this PR and these changes should vanish

src/Forms/TextCheckboxGroupField.php
emteknetnz: Good spot, updated

GuySartorelli: The "but probably allows other value types" part of my comment also worries me a little. Presumably it deals in truthy and falsy, not explicitly 1 and 0? If I'm right about that, we should make this a loose truthy check, not an exact equals check.

For example, see the check in `CheckboxField::getAttributes()`

emteknetnz: While CheckboxField::getValue() has a mixed return type, it does simply return 1 or 0.  I think it's safe to rely on that not being changed by a subclass as there's no real reason to change it. If you want we could remove the === 1 bit of the comparison but I don't think it's necessary

GuySartorelli: Right you are - sorry, I was conflating `$value` with `getValue()`.

GuySartorelli (approved):
LGTM

PR #1307, opened 2025-01-15
FIX Call property directly to prevent infinite recursion

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/357

Fixes issue where framework FormMessage::getMessage() and elemental TextCheckboxGroupField::getMessageCast() end up in an infinite loop e.g. https://github.com/silverstripe/silverstripe-elemental/actions/runs/12761071213/job/35617198112 though also happens in quite a few other places

Can be triggered by add-block-element-to-data-object.feature:11


GuySartorelli (approved):
LGTM

PR #1311, opened 2025-01-28
FIX Pass empty string instead of null for error

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/11423


GuySartorelli (approved):
LGTM will merge when main PR is merged

Issue #1310, opened 2025-01-27
Non-inline editable blocks cannot be published via "three dots" menu


TheSceneman:
### Module version(s) affected

5.3.5

### Description

When attempting to publish a block that is non-inline editable via "three dots" menu on the list of elements, an error toast is returned: "Error displaying the edit form for this block"

### How to reproduce

Add a non-inline editable block that contains a Gridfield to a page. Click on it, make some edits and save (but don't publish). Go back the page and attempt to publish via the three dots. You'll see an error toast.

### Additional Context

This could be caused by this change in 5.3: https://docs.silverstripe.org/en/5/changelogs/5.3.0/#elemental-validation

### Validations

- [x] Check that there isn't already an issue that reports the same bug
- [ ] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## PRs

- https://github.com/silverstripe/silverstripe-elemental/pull/1312
- https://github.com/silverstripe/silverstripe-frameworktest/pull/225

GuySartorelli:
I can't reproduce this with the steps provided on a fresh installation of `silverstripe/installer` with `dnadesign/silverstripe-elemental` as the only additional dependency, and no custom code.

Can you please update the reproduction steps to include everything needed to reproduce this?

TheSceneman:
Thanks for the response Guy, I will double check the exact scenario. Might have something to do with child objects

TheSceneman:
@GuySartorelli I think this is tied to Gridfields, I commented out the GridField in the getCMSFields() and the block was then publishable. Updated steps to mention gridfield

GuySartorelli:
Haven't looked at that yet but that would make sense - I believe in 5.3 we made a change so that the elemental area loads the form in react in order to submit a publish event which obviously will fail with gridfields, since there's no react component for those. 

## A workaround

As a workaround if you need something urgent you can replace `GridField` with injector, and in the subclass set the protected `schemaType` property to `FormField::SCHEMA_DATA_TYPE_STRUCTURAL`.

This should work, and _probably_ won't have unintended side effects, though I can't promise that.

## A fix

A proper fix for this will need to _stop_ loading the form when publishing a non-inline-editable block, which makes sense since you can't edit them in this context anyway, so you don't need to send the field values along with the publish request.

[...]

PR #1312, opened 2025-01-30
FIX Allow publishing non-inline-editable blocks

GuySartorelli:
Needs https://github.com/silverstripe/silverstripe-frameworktest/pull/225 for CI to go green, so merge that first

## Issue
- https://github.com/silverstripe/silverstripe-elemental/issues/1310


client/src/components/ElementEditor/Element.js
```diff
@@ -165,12 +167,12 @@ const Element = (props) => {
 
   // Publish action
   useEffect(() => {
-    if (formHasRendered && doPublishElement) {
+    if (doPublishElement) {
```
emteknetnz: Seems like this bit should not be getting removed? Checking the formHasRendered is presumably still required for regular inline editable blocks?

client/src/components/ElementEditor/Element.js
```diff
@@ -43,6 +43,8 @@ const Element = (props) => {
   const [hasUnsavedChanges, setHasUnsavedChanges] = useState(false);
   const [publishBlock] = useMutation(publishBlockMutation);
 
+  const formRenderedIfNeeded = () => formHasRendered || !props.type.inlineEditable;
```
emteknetnz: Honestly think it's clearer to not have this new method and just update the one place it's used to check `&& props.type.inlineEditable`

emteknetnz (changes requested):

client/src/components/ElementEditor/Element.js
GuySartorelli: Oops yup this should be using `formRenderedIfNeeded`. FWIW it doesn't _seem_ to be needed based on my testing and behat - but presumably it was there for a reason and should remain there for the type of element this bug doesn't affect.

client/src/components/ElementEditor/Element.js
GuySartorelli: It's used in two places (after I fix https://github.com/silverstripe/silverstripe-elemental/pull/1312#discussion_r1938574598) - I'd prefer to keep it since it makes those two places clearer to me, but if you still think it's better inlined after seeing it in situ with both conditions, I'll change it.

GuySartorelli: I *have* converted it from a function to a variable though. Dunno why I thought it should be a function lol

emteknetnz (changes requested):
I performed and updated kitchen sink CI run where there was a failing behat test with the latest PRs - https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/13148450879

That is not an existing failure - https://github.com/silverstripe/recipe-kitchen-sink/actions/runs/13148969515

GuySartorelli:
I would be shocked if this small change to elemental's react components broke the CMS behat for real. Alerts are usually intermittent failures in behat - I've rerun the failure and will take a look when it's done. 

Edit: Went green after rerunning it.

emteknetnz (approved):

PR #1314, opened 2025-02-03
Reports print view - escaped html bug

thisisannie:
## Description

Print Report view was double-escaping the mdash, showing the escaped html "&mdash;" instead of "—"
Screenshot 2025-02-04 at 11 10 20 AM

This fix replaces the escaped character with the character itself.

I've also removed the redundant brackets and interpolated the string, because I can't help myself.

## Manual testing steps

1. In the CMS, navigate to Reports
2. Select a report
3. Click Print button

## Issues

- https://github.com/silverstripe/silverstripe-framework/issues/11596

## Pull request checklist

- [x] The target branch is correct
- [x] All commits are relevant to the purpose of the PR (e.g. no debug statements, unrelated refactoring, or arbitrary linting)
- [ ] The commit messages follow our [commit message guidelines](https://docs.silverstripe.org/en/contributing/code/#commit-messages)
- [x] The PR follows our [contribution guidelines](https://docs.silverstripe.org/en/contributing/code/)
- [x] Code changes follow our [coding conventions](https://docs.silverstripe.org/en/contributing/coding_conventions/)
- [x] This change is covered with tests (or tests aren't necessary for this change)
- [x] Any relevant User Help/Developer documentation is updated; for impactful changes, information is added to the changelog for the intended release

### DIFF
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -1162,7 +1162,7 @@ JS
     public function getTypeNice()
     {
         $description = Deprecation::withSuppressedNotice(fn () => $this->getDescription());
-        $desc = ($description) ? ' <span class="element__note"> &mdash; ' . $description . '</span>' : '';
+        $desc = $description ? " <span class=\"element__note\"> — {$description}</span>" : '';
 
         return DBField::create_field(
             'HTMLVarchar',


GuySartorelli (commented):
This is handling the very narrow case for the `&mdash;` here - but it looks like you've actually uncovered a much wider bug that will need to be patched in framework directly - it seems the value is being double escaped by the `GridFieldPrintButton` component.

I'll submit a PR for fixing this in framework once I've finished investigating where exactly the fix should be - I have a proof-of-concept but need to be careful not to introduce a potential XSS vector with the fix.

In any case because this PR doesn't fix the wider bug, I'll be closing it.

GuySartorelli:
Reopening - turns out this will still be needed to resolve the problem for this specific report, due to the fact that it formats this column to include HTML - and because it includes HTML, escaping the HTML to print it escapes the mdash character as well.

emteknetnz (approved):

Issue #1288, opened 2024-12-03
Validation error popup incorrectly appears when inline editing


lozcalver:
### Module version(s) affected

5.3.0

### Description

After the first AJAX save has happened, subsequent ones trigger a validation error popup when there are no validation issues. Clicking to save a second time (basically ignoring the error popup and trying again) works as expected.

In the video below, I add two content blocks but the same thing happens when only one is added.

**Note:** I have _not_ yet validated that this issue occurs on a blank install, though I’m fairly sure it will as I removed any extensions from `BaseElement`/`ElementContent` before recording the video, I just wanted to write this up while I have the affected project open to demonstrate.

https://github.com/user-attachments/assets/d91c2dcd-a66f-4c03-8e87-22f8667feabb

### How to reproduce

- Add a content block, save and publish
- Click to edit, make a change
- Click save and publish, a validation error popup appears
- Click save and publish again, the content is saved as expected

### Validations

- [X] Check that there isn't already an issue that reports the same bug
- [ ] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

GuySartorelli:
This will be fixed by https://github.com/silverstripe/silverstripe-elemental/pull/1319 when that gets merged - there's a lot of tangled mess with the changetracking failure that seems likely to have caused other problems such as this one.

GuySartorelli:
PR merged - will be automatically tagged by GitHub Actions

Issue #1287, opened 2024-12-03
Changetracking doesn’t appear to be working correctly for inline editing


lozcalver:
### Module version(s) affected

5.3.0

### Description

When inline editing elements, changetracking doesn’t appear to work. If you edit an element’s title/content, the “Save” & “Save and Publish” buttons don’t get highlighted, and navigating away from the page doesn’t open the warning alert about unsaved changes.

This appears to be a regression, as it functions as expected in 5.2

### How to reproduce

Create a standard “Content” block, add some content, save & publish.

Edit the title or content, note that the save buttons aren’t highlighted.

Close the tab, refresh the page, or click a link to navigate away - note that the alert to warn about losing unsaved changes doesn’t appear.

### Validations

- [X] Check that there isn't already an issue that reports the same bug
- [X] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## PRs
- https://github.com/silverstripe/silverstripe-elemental/pull/1319

guci0:
Same story here. EDIT: Latest working 5.2.4

GuySartorelli:
PR merged - will be automatically tagged by GitHub Actions

PR #1319, opened 2025-02-19
FIX Make change tracking work again for inline editable blocks

GuySartorelli:
This PR effectively reverts dirty form detection for elemental back to how it was in 5.2 which was the last time it worked.
Change tracking is weirdly intertwined with a few other functionalities so other changes were also needed.

Behat tests have been made more robust to help prevent this sort of regression happening again.

## Issues
- https://github.com/silverstripe/silverstripe-elemental/issues/1287
- https://github.com/silverstripe/silverstripe-elemental/issues/1288


client/src/components/ElementEditor/Element.js
```diff
@@ -435,6 +436,7 @@ const Element = (props) => {
         broken={type.broken}
         onFormSchemaSubmitResponse={handleFormSchemaSubmitResponse}
         onFormInit={() => handleFormInit(activeTab)}
+        formDirty={formDirty}
```
GuySartorelli, line 402: I assume this was originally removed with the intention of using the new context in the `Content` component - but that's a class component so it can't use hooks. Using the prop that already exists is easier and more suitable as a patch compared with completely refactoring that component to a functional one.

tests/Behat/features/file-upload.feature
```diff
@@ -8,12 +8,52 @@ Feature: Files can be saved in and removed from elemental blocks
     And I add an extension "SilverStripe\FrameworkTest\Elemental\Extension\FileElementalExtension" to the "DNADesign\Elemental\Models\ElementContent" class
     And I go to "/dev/build?flush"
     And a "image" "file1.jpg"
+    And a "image" "file2.jpg"
```
GuySartorelli, line 11: If we add a file, save, remove the file and try to add the same file back, since it's the same file the form isn't dirty so we can't save.

The dirty state didn't change before, but now that it does we need to test this.
Anything with a `file2` is for testing that scenario.

client/src/components/ElementEditor/Element.js
```diff
@@ -39,32 +39,19 @@ const Element = (props) => {
...
 
   useEffect(() => {
-    // Note that formDirty from redux can be set to undefined after failed validation
-    // which is confusing as the block still has unsaved changes, hence why we create
-    // this state variable to track this instead
-    // props.formDirty is either undefined (when pristine) or an object (when dirty)
-    const formDirty = typeof props.formDirty !== 'undefined';
-    if (formDirty && !hasUnsavedChanges) {
-      setHasUnsavedChanges(true);
-    }
```
GuySartorelli, line 56: That's not how formDirty works now - it's just a boolean same as it used to be in 5.2.

client/src/components/ElementEditor/Element.js
```diff
@@ -490,16 +457,6 @@ function mapDispatchToProps(dispatch, ownProps) {
...
-    dispatchAddFormChanged() {
-      // Ensures the form identifier is in unsavedForms in the global redux state
-      // This is used to derive the formDirty prop in mapStateToProps
-      dispatch(addFormChanged(`element.${elementName}`));
-    },
-    dispatchRemoveFormChanged() {
-      // Removes the form identifier from unsavedForms in the global redux store
-      // Opposite of beheaviour of dispatchAddFormChanged()
-      dispatch(removeFormChanged(`element.${elementName}`));
-    },
```
GuySartorelli, line 502: Just rely on the existing formdirty data from redux.

tests/Behat/features/inline-block-validation.feature
```diff
@@ -24,6 +25,7 @@ Feature: Blocks are validated when inline saving individual blocks
     And I press the "Insert" button
     And I press the "View actions" button
     And I click on the ".element-editor__actions-save" element
+    And I dismiss all toasts
```
GuySartorelli, line 28: Sometimes I was getting the "saved block" toast from this save when checking in the test scenarios.

tests/Behat/features/inline-block-validation.feature
GuySartorelli, line 1: There was a lot of weirdness in toasts (e.g. validation toast firing after saving the page) so I've added some robustness around these test scenarios.

tests/Behat/features/inline-block-validation.feature
```diff
@@ -33,68 +35,112 @@ Feature: Blocks are validated when inline saving individual blocks
...
     And I click on the ".element-editor__actions-save" element
-    Then I should see "Title cannot be x" in the ".form__validation-message" element
+    Then I should not see a "Saved 'x' successfully" success toast
+    And I should see "Title cannot be x" in the ".form__validation-message" element
+    # Reset value to something valid and resave - validates saving after validation message works
+    # and avoids an alert after the test ends
+    # Specifically swapping back to the original value - we don't actually save the block so not block save toast
+    And I fill in "My title" for "Title" for block 1
+    And I wait for 1 second
+    And I press the "Save" button
```
GuySartorelli, line 45: Because change tracking works now, we can't just leave the unsaved changes hanging there.

client/src/components/ElementEditor/Element.js
```diff
@@ -39,32 +39,19 @@ const Element = (props) => {
   const [doPublishElementAfterSave, setDoPublishElementAfterSave] = useState(false);
   const [ensureFormRendered, setEnsureFormRendered] = useState(false);
   const [formHasRendered, setFormHasRendered] = useState(false);
-  const [doDispatchAddFormChanged, setDoDispatchAddFormChanged] = useState(false);
-  const [hasUnsavedChanges, setHasUnsavedChanges] = useState(false);
```
GuySartorelli, line 43: No need to separately keep track of `hasUnsavedChanges` because we can just rely on `formDirty` instead.
No need to use `doDispatchAddFormChanged` because redux form keeps track of that for us automatically.

client/src/components/ElementEditor/Element.js
```diff
@@ -462,7 +427,9 @@ function mapStateToProps(state, ownProps) {
 
   const tabSetName = tabSet && tabSet.id;
   const uniqueFieldId = `element.${elementName}__${tabSetName}`;
-  const formDirty = state.unsavedForms.find((unsaved) => unsaved.name === `element.${elementName}`);
+
+  const formName = loadElementFormStateName(ownProps.element.id);
+  const formDirty = isDirty(`element.${formName}`, getFormState)(state);
```
GuySartorelli, line 432: This is just swapping us back to how this was in CMS 5.2. This was changed in 5.3 but it's unclear why.
Dirty state tracking worked in 5.2 so I've just reverted this back to how it was then.

client/src/components/ElementEditor/ElementList.js
```diff
@@ -39,6 +39,13 @@ class ElementList extends Component {
       this.resetState(prevState, false);
       return;
     }
+    // Scenario we've just clicked "save" or "submit" on the form
+    if (this.state.saveAllElements && !prevState.saveAllElements) {
+      // Reset the validation state of all blocks.
+      // This mirrors handleBeforeSubmitForm() for individual blocks.
+      this.resetState(prevState, false);
+      return;
+    }
```
GuySartorelli, line 48: Needed because otherwise we retain the old "invalid" state. As noted in the code comment, resetting the validation state is already done when saving a block with the `...` menu, so makes sense to reset it when submitting via the page save button as well.

client/src/components/ElementEditor/ElementList.js
```diff
@@ -154,9 +161,7 @@ class ElementList extends Component {
     }
 
     let output = blocks.map(element => {
-      const saveElement = this.state.saveAllElements
-        && this.state.hasUnsavedChangesBlockIDs[element.id]
-        && this.state.validBlockIDs[element.id] === null;
+      const saveElement = this.state.saveAllElements && this.state.hasUnsavedChangesBlockIDs[element.id];
```
GuySartorelli, line 164: We should attempt to save the block regardless of the validation state it was in prior to trying to save the block.

emteknetnz (approved):
Seems to work well

PR #1323, opened 2025-02-25
API Update search functionality to work in a SearchContext

GuySartorelli:
This relied on the weird non-standard way CMSMain page filtering worked. This same functionality could be reintroduced by replacing the SiteTreeSearchContext with a custom one that filters by callback, but it won't be available in this module.

## Issue
- https://github.com/silverstripe/silverstripe-cms/issues/2949


emteknetnz (changes requested):
We'll need to resolve https://github.com/silverstripe/developer-docs/pull/705#discussion_r1976866479 first so we know what the scope of changes here needs to be

emteknetnz:
Following on discussion from https://github.com/silverstripe/developer-docs/pull/705#discussion_r1976866479

There's a sliding scale of options for including element content as something that's included in general page search:

A) Status quo - keep existing functionality, allow developers to opt-out when site grows and performance degrades
B) Keep functionality available though make it an opt-in, will inevitably be some end-users unhappy that functionality vanished as it was missed during CMS 6 upgrade. On the flipside search performance will be much better.
C) Remove functionality under justification that not using a dedicated search index system like elastic will always have awful scaling performance. However there's no quick drop in solution for this leading to an inevitable loss in user experience. Tell users in changelog they can implement themselves if they want.

I don't think C) is really viable given the lack of a drop in alterative, and because it's presumably very easy with our current knowledge to develop this, we may as well just do B) at that point rather than telling people to do it themselves.

Between A) and B), I'd say we should probably still do A), because this being is done in the context of a refactoring card where the outcome should be that functionality remains the same while the under the hood code is now in a better state. I'm not liking the fact where just randomly dropping functionality which seems quite out of scope

GuySartorelli:
Sorry, looks like I forgot to push some of those requested changes til now.

GuySartorelli:
I don't have time to argue unfortunately so I've just done what you want, though I still don't think it's the best course of action.

Please next time advise you want these sorts of changes before merging the deprecation PRs - I have to go back and redo those now.

src/ORM/Search/ElementalSiteTreeSearchContext.php
```diff
@@ -0,0 +1,84 @@
...
+        $pageIDs = [];
+        // The same extension can't be applied to the multiple classes in the same hierarchy
+        // without causing a host of problems, so we can be confident that we're not getting any double ups here.
+        $pageClassesWithExtension = ClassInfo::classesWithExtension(ElementalPageExtension::class, SiteTree::class, true);
+        // Get a list of classes that can't have elemental blocks despite having the extension to reduce the amount of
+        // records we're filtering through needlessly
+        $ignoredClasses = Config::forClass(ElementalPageExtension::class)->get('ignored_classes');
+        foreach ($ignoredClasses as $class) {
+            $ignoredClasses = array_merge($ignoredClasses, ClassInfo::subclassesFor($class, false));
+        }
```
GuySartorelli, line 51: This part and the `exclude` call are new, so we don't filter through records that just aren't relevant at all.
Everything else is the same which some minor refactoring to work in this method as opposed to the old one.

emteknetnz (changes requested):
I've merged the updated 5.4 PR - https://github.com/silverstripe/silverstripe-elemental/pull/1328

That will need to be merged-up, though I tried doing this myself though because of the recently merged [change-tracker](https://github.com/silverstripe/silverstripe-elemental/pull/1319) work it's a non-trival merge-up, you're best placed to resolve this, so could you please sort that, and then rebase this PR and run `yarn build`

GuySartorelli:
Rebased. There's no JS or CSS changes in this PR so I didn't `yarn build`

emteknetnz (approved):

PR #1334, opened 2025-03-06
MNT Add new context to behat config

GuySartorelli:
Needed because of https://github.com/silverstripe/silverstripe-cms/pull/3059#discussion_r1984185238

## Issue
- https://github.com/silverstripe/silverstripe-htmleditor-tinymce/issues/2

### DIFF
--- a/behat.yml
+++ b/behat.yml
@@ -15,6 +15,7 @@ default:
         - SilverStripe\Framework\Tests\Behaviour\CmsUiContext
         - SilverStripe\BehatExtension\Context\BasicContext
         - SilverStripe\BehatExtension\Context\LoginContext
+        - SilverStripe\CMS\Tests\Behaviour\AnchorContext
         -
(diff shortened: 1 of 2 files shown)


emteknetnz (changes requested):
Will need to add a composer dev dependency for `silverstripe/silverstripe-htmleditor-tinymce` since there are behat tests that rely on it?

Will need a new PR for subsites to do the same?

GuySartorelli:
We typically accept that silverstripe/installer will be included, don't we?

emteknetnz:
Yeah, in practice it'll work fine in CI and on our locals, though from memory we still include dev-deps that behat needs, similar to how we do for phpunit which in practice also always has installer too

I might have remember that wrong though.  It's certainly less obvious for behat when deps are actually needed

GuySartorelli:
Dev dep added

emteknetnz (approved):

Issue #1333, opened 2025-03-06
CMS 6 REGRESSION: Can't click-and-drag select title field in inline-editable blocks


GuySartorelli:
In CMS 6 specifically, if I try to click-and-drag to select text inside the "Title" text field in an inline-editable block, it starts dragging the block instead.

Dragging to select text in the WYSIWYG works. I didn't test any other fields but presumably any `input` tag has the problem.

## PR
- https://github.com/silverstripe/silverstripe-elemental/pull/1336

GuySartorelli:
PR merged

PR #1336, opened 2025-03-19
FIX Prevent drag and drop on input fields

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-elemental/issues/1333


GuySartorelli (approved):
LGTM, works as expected.

PR #1342, opened 2025-04-01
FIX Multi locale localised copy correction.

mfendeksilverstripe:
# FIX Multi locale localised copy correction.

Part of https://github.com/tractorcow-farm/silverstripe-fluent/pull/941 change-set.
Fixes https://github.com/tractorcow-farm/silverstripe-fluent/issues/940

Issue spans two modules: Fluent & Elemental.

### DIFF
--- a/src/TopPage/FluentExtension.php
+++ b/src/TopPage/FluentExtension.php
@@ -44,6 +44,11 @@ class FluentExtension extends DataExtension
         DataObject $localisedOwner,
         ?DataObject &$duplicate
     ): void {
+        // Some other extension already populated the duplicate so it doesn't have to be actioned here
+        if ($duplicate) {
+            return;
+        }
+
         $originalRelation = $this->getOwner();
         $localisedRelation = $localisedOwner->getComponent($relation);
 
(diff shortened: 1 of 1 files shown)


src/TopPage/FluentExtension.php
```diff
@@ -35,6 +35,37 @@ public function updateCMSFields(FieldList $fields)
         $fields->removeByName('TopPageLocale');
     }
 
+    /**
+     * Covers the case where top page metadata is populated during duplication process of localised data
```
n8-dev, line 39: Prob be good to add a bit more to this function about why it exists yeah?

Its likely the **only** mention of `Locale` in this entire codebase right?
Perhaps a note of this is to help compatibility with Fluent, a link to the fluent repo, etc...

🤔

mfendeksilverstripe: There are already docs related to this extension and what's the purpose of it.

GuySartorelli: Since this is explicitly the `FluentExtension` I think it's clear enough that this pertains to logic in `tractorcow/silverstripe-fluent` as mentioned in the PHPDoc for the class itself.

GuySartorelli (approved):
LGTM and works as expected.

Issue #1348, opened 2025-04-28
Performance issue (setting TopPageID) when creating a new page on a big site


AljosaB:
## Module version(s) affected
5.4.0 / all

## Environment
- latest SS (also tested on SS 6)
- latest Elemental
- https://github.com/silverstripe/silverstripe-elemental/blob/5/docs/en/02_advanced_setup.md#top-page-reference-performance-enhancement applied
- 4 elemental areas added to Page
- 50 page types
- 150k page records

## The flow
When someone creates a new page inside CMS this is what happens:

1. Elemental areas get created (in my case 4) and each of them triggers `onAfterWrite()` (added by TopPage extension) which has one purpose only - to set `TopPageID` in ElementalArea (calls `setTopPage($page)`) for future performance improvements
2. At that time the page doesn't exist in the DB yet and it doesn't have ID, so we enter `setTopPage()` without passing the $page parameter which means `getTopPage()` gets triggered
3. `getTopPage()` usually returns page from getting it by ID from the TopPageID field, but in this case it doesn't have it yet so it goes into `getOwnerPage()`
4. `getOwnerPage()` queries the DB looking for page having `ElmentalAreaID` of the area it just created which can't exist as page is not yet written to DB; it does that for each "supportedPageType" which in my case are all 50 page types

## The problem
The search for TopPage shouldn't be performed at that time as the page doesn't exist yet but it is, which isn't a problem on small DBs, but in my case it sums up to: 4 (elemental areas) * 50 (page types) * 0.2s (query time - depends on amount of page records and it varies) = **~30s**

It seems the authors were aware that TopPageID wasn't going to be written at that point (in this case) and added extra `onAfterWrite()` to the page (SiteTreeExtension) that calls `setTopPageForElementalArea()` where it actually sets it (page is passed to `setTopPage()` this time). The function comment explains it

```
     * Elemental area is created before related page is written so we have to set top page explicitly
     * after page is written and the relations are available
```

## Solution
Unfortunately I don't know all (edge) cases to come up with a solution, but maybe it's possible to detect the page being created in elemental area's `onAfterWrite` and skip the `setTopPage()` ?

## PRs
- https://github.com/silverstripe/silverstripe-elemental/pull/1356

GuySartorelli:
PR merged. This will be included in the CMS 6.1 release in October.

PR #1356, opened 2025-06-04
ENH Do not attempt to set top page when owner is not persisted

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-elemental/issues/1348

### DIFF
--- a/src/Extensions/ElementalAreasExtension.php
+++ b/src/Extensions/ElementalAreasExtension.php
@@ -14,10 +14,8 @@ use SilverStripe\Core\Extensible;
 use SilverStripe\Forms\FieldList;
 use SilverStripe\Forms\LiteralField;
 use SilverStripe\ORM\DataObject;
-use SilverStripe\ORM\RelatedData\StandardRelatedDataService;
-use SilverStripe\Versioned\Versioned;
-use SilverStripe\Model\ModelData;
 use SilverStripe\Core\Extension;
+use DNADesign\Elemental\Extensions\TopPageElementExtension;
 
 /**
  * This extension handles most of the relationships between pages and element
@@ -278,17 +276,25 @@ class ElementalAreasExtension extends Extension
      */
     public function ensureElementalAreasExist($elementalAreaRelations)
     {
+        $owner = $this->getOwner();
         foreach ($elementalAreaRelations as $eaRelationship) {
             $areaID = $eaRelationship . 'ID';
-
-            if (!$this->owner->$areaID) {
+            if (!$owner->$areaID) {
                 $area = ElementalArea::create();
                 $area->OwnerClassName = get_class($this->owner);
-                $area->write();
-                $this->owner->$areaID = $area->ID;
+                // Do not attempt to set the ElementalArea.TopPageID if the owner (e.g. Page)
+                // has not yet been persisted to the database, as this will cause a potentially
+                // large number of unneccessary database queries
+                if (!$owner->isInDB() && $area::has_extension(TopPageElementExtension::class)) {
+                    /** @var ElementalArea&TopPageElementExtension $area */
+                    $area->withoutCallingSetTopPage(fn() => $area->write());
+                } else {
+                    $area->write();
+                }
+                $owner->$areaID = $area->ID;
             }
         }
-        return $this->owner;
+        return $owner;
     }
 
     /**
--- a/src/Extensions/TopPageElementExtension.php
+++ b/src/Extensions/TopPageElementExtension.php
@@ -41,6 +41,11 @@ class TopPageElementExtension extends Extension
     ];
 
     /**
+     * Whether to call setTopPage() when onAfterWrite() is called
+     */
+    private bool $doCallSetTopPage = true;
+
+    /**
      * Global flag which indicates that automatic page determination is enabled or not
      * If this is set to a page ID it will be used instead of trying to determine the top page
      *
@@ -56,7 +61,9 @@ class TopPageElementExtension extends Extension
      */
     protected function onAfterWrite(): void
     {
-        $this->setTopPage();
+        if ($this->doCallSetTopPage) {
+            $this->setTopPage();
+        }
     }
 
     /**
@@ -194,6 +201,20 @@ class TopPageElementExtension extends Extension
         } finally {
             $this->fixedTopPageID = $original;
         }
+    }
+
+    /**
+     * Prevents onAfterWrite() from calling setTopPage() for the duration of the callback
+     *
+     * @return mixed The value returned by $callback or null if $callback has no return value
+     */
+    public function withoutCallingSetTopPage(callable $callback): mixed
+    {
+        $orig = $this->doCallSetTopPage;
+        $this->doCallSetTopPage = false;
+        $ret = call_user_func($callback);
+        $this->doCallSetTopPage = $orig;
+        return $ret;
     }
 
     protected function updateCMSFields(FieldList $fields)
--- a/tests/Extensions/TopPageElementExtensionTest.php
+++ b/tests/Extensions/TopPageElementExtensionTest.php
@@ -1,0 +1,54 @@
+<?php
+
+namespace DNADesign\Elemental\Tests\Extensions;
+
+use DNADesign\Elemental\Extensions\ElementalAreasExtension;
+use DNADesign\Elemental\Extensions\TopPageElementExtension;
+use PHPUnit\Framework\Attributes\DataProvider;
+use SilverStripe\CMS\Model\SiteTree;
+use SilverStripe\Dev\SapphireTest;
+use DNADesign\Elemental\Models\ElementalArea;
+
+class TopPageElementExtensionTest extends SapphireTest
+{
+    protected static $required_extensions = [
+        SiteTree::class => [
+            ElementalAreasExtension::class,
+        ],
+    ];
+
+    public static function provideWithoutCallingSetTopPage(): array
+    {
+        return [
+            'with-call' => [
+                'doCall' => true,
+                'expectTopPageIDSet' => true,
+            ],
+            'without-call' => [
+                'doCall' => false,
+                'expectTopPageIDSet' => false,
+            ],
+        ];
+    }
+
+    #[DataProvider('provideWithoutCallingSetTopPage')]
+    public function testWithoutCallingSetTopPage(bool $doCall, bool $expectTopPageIDSet): void
+    {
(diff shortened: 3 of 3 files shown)


src/Extensions/ElementalAreasExtension.php
```diff
@@ -14,10 +14,8 @@
 use SilverStripe\Forms\FieldList;
 use SilverStripe\Forms\LiteralField;
 use SilverStripe\ORM\DataObject;
-use SilverStripe\ORM\RelatedData\StandardRelatedDataService;
-use SilverStripe\Versioned\Versioned;
-use SilverStripe\Model\ModelData;
```
emteknetnz, line 19: Unused imports

src/Extensions/ElementalAreasExtension.php
```diff
@@ -278,17 +276,27 @@ public function supportsElemental()
...
+                // Do not attempt to set the ElementalArea.TopPageID if the owner (e.g. Page)
+                // has not yet been persisted to the database, as this will cause a potentially
+                // large number of unneccessary database queries
+                if ($owner->ID === 0 && $area::has_extension(TopPageElementExtension::class)) {
+                    /** @var ElementalArea&TopPageElementExtension $area */
+                    $area->withoutCallingSetTopPage($writeAreaFunc);
+                } else {
+                    call_user_func($writeAreaFunc);
+                }
                 $area->write();
```
GuySartorelli: Why are we calling `write()` twice (once above in a weird callback and once here)?

src/Extensions/ElementalAreasExtension.php
```diff
@@ -278,17 +276,27 @@ public function supportsElemental()
...
                 $area->OwnerClassName = get_class($this->owner);
+                $writeAreaFunc = fn() => $area->write();
+                // Do not attempt to set the ElementalArea.TopPageID if the owner (e.g. Page)
+                // has not yet been persisted to the database, as this will cause a potentially
+                // large number of unneccessary database queries
+                if ($owner->ID === 0 && $area::has_extension(TopPageElementExtension::class)) {
+                    /** @var ElementalArea&TopPageElementExtension $area */
+                    $area->withoutCallingSetTopPage($writeAreaFunc);
+                } else {
+                    call_user_func($writeAreaFunc);
```
GuySartorelli: Why are we calling a callback here? Isn't the callback explicitly only for use in `withoutCallingSetTopPage()`?

src/Extensions/ElementalAreasExtension.php
```diff
@@ -278,17 +276,27 @@ public function supportsElemental()
...
-
-            if (!$this->owner->$areaID) {
+            if (!$owner->$areaID) {
                 $area = ElementalArea::create();
                 $area->OwnerClassName = get_class($this->owner);
+                $writeAreaFunc = fn() => $area->write();
+                // Do not attempt to set the ElementalArea.TopPageID if the owner (e.g. Page)
+                // has not yet been persisted to the database, as this will cause a potentially
+                // large number of unneccessary database queries
+                if ($owner->ID === 0 && $area::has_extension(TopPageElementExtension::class)) {
```
GuySartorelli:
```suggestion
                if (!$owner->isInDB() && $area::has_extension(TopPageElementExtension::class)) {
```
This method exists specifically for these scenarios.

src/Extensions/ElementalAreasExtension.php
```diff
@@ -278,17 +276,27 @@ public function supportsElemental()
...
+                    /** @var ElementalArea&TopPageElementExtension $area */
+                    $area->withoutCallingSetTopPage($writeAreaFunc);
```
GuySartorelli: I don't like this coupling - ideally this extension shouldn't have to know about `TopPageElementExtension` at all. Is there another way we can do this?

My immediate thought is to add a new API to `DataObject` which is something like `isFirstWrite()` which will be set to true in `preWrite()` if `!$this->isInDB()` (or perhaps better -always returns true until some flag is set to false) and set to false at the end of `write()`. Then in `TopPageElementExtension::onAfterWrite()` just don't set top page for the first write?

There are a bunch of times people have wanted to detect a first write in `onAfterWrite()` hooks so this would be useful for other developers' use cases as well.

GuySartorelli (changes requested):

src/Extensions/ElementalAreasExtension.php
emteknetnz: Removed

src/Extensions/ElementalAreasExtension.php
emteknetnz: Updated

src/Extensions/ElementalAreasExtension.php
emteknetnz: Updated

src/Extensions/ElementalAreasExtension.php
emteknetnz: I think the coupling here is fine because both extensions are in the elemental module, both in the `DNADesign\Elemental\Extensions` namespace

GuySartorelli: I really don't like it. It feels like the abstraction isn't robust enough if two separate extensions have to directly reference one another like this. Especially if one has a new method that's only used in the other - and especially when there's a way to resolve this without adding this coupling.

If anything, `TopPageElementExtension` should reference `ElementalAreasExtension` (since top page feature is built on top of elemental feature) - not the other way around.

emteknetnz:
> My immediate thought is to add a new API to DataObject which is something like isFirstWrite() which will be set to true in preWrite() if !$this->isInDB() (or perhaps better -always returns true until some flag is set to false) and set to false at the end of write(). Then in TopPageElementExtension::onAfterWrite() just don't set top page for the first write?

With TopPageElementExtension, which is applied to both ElementalAreas and Elements, it won't always be the case (in fact won't usually be the case) that Page doesn't exist in the database, so it's inappropriate use something like isFirstWrite(), because normally the page exists when an Element is created.

This particular scenario is only happens when creating the ElementalArea on a new page, as the ElementalArea is intentionally created  first (I've looked into changing this so that the Page is created first, then the ElementalArea, I forget the exact reason but did double check that it does makes sense to create the ElementalArea first.

> Especially if one has a new method that's only used in the other

I'm not sure how that's relevant. The new method is pubic and could be used by anything.

> If anything, TopPageElementExtension should reference ElementalAreasExtension (since top page feature is built on top of elemental feature) - not the other way around.

I get that reasoning, though I don't see how the logic can be reversed.
- ElementalAreasExtension::ensureElementalAreasExist() / Page::onBeforeWrite() is what creates the ElementalArea.
- TopPageElementExtension::onAfterWrite() / ElementalArea::onAfterWrite() is what calls TopPageElementExtension::setTopPage()

How is TopPageElementExtension/ElementalArea::onAfterWrite() supposed know to not call setTopPage() unless we tell it ahead of time not to? (which is what the call to `$area->withoutCallingSetTopPage(fn() => $area->write());` does)

src/Extensions/ElementalAreasExtension.php
emteknetnz: Replied here https://github.com/silverstripe/silverstripe-elemental/pull/1356#issuecomment-2942627070

GuySartorelli: Please don't make a reply outside the thread, it makes it hard to reply to. I'll reply here.

> (I've looked into changing this so that the Page is created first, then the ElementalArea, I forget the exact reason but did double check that it does makes sense to create the ElementalArea first.

Can't do `onAfterWrite()` because the `has_one` is on the $this->getOwner()` - so we'd have to either:
1. call `write()` on it a second time if we did that which would be horrendous
1. do a manual SQL `UPDATE` (which `TopPageElementExtension` already does so wouldn't be the end of the world) but that could mess up project-level extension hooks that expect the elemental area to already exist.

So yup, `onBeforeWrite()` is the correct hook for the page.

>> Especially if one has a new method that's only used in the other
>
> I'm not sure how that's relevant. The new method is pubic and could be used by anything.

It _could_ be used for anything, but realistically it won't be. It really exists specifically for this one workaround of a weird abstraction.

>> If anything, TopPageElementExtension should reference ElementalAreasExtension (since top page feature is built on top of elemental feature) - not the other way around.
>
> I get that reasoning, though I don't see how the logic can be reversed.

Unfortunately after looking at it some more I can't see a cleaner way forward either - not without breaking changes, anyway.

I had thought we could just call `setComponent('Owner', $owner)` on the new elemental area when creating it, which could then be pulled out during `getOwnerPage()` (i.e. we have all of the records _in memory_ - there's no reason to do any database calls) - but of course there is no actual `Owner` relation on the elemental area to store the page in despite the `OwnerClassName` column.

Ideally this would be turned into a multi-relational has_one similar to silverstripe/linkfield's `Owner` relation (which frankly would make `TopPage` redundant)... but I'm on the fence about if we could do that in a minor release.

Do you have thoughts on that? If that's too risky then I'll just merge this PR (presuming it works when I test it locally)

emteknetnz: That seems too risky, that seems more like the sort of thing we should do in a major release than a minor.

We're just trying to do a performance fix/enhancement here, not re-architect things.

src/Extensions/ElementalAreasExtension.php
```diff
@@ -278,17 +276,25 @@ public function supportsElemental()
...
-            if (!$this->owner->$areaID) {
+            if (!$owner->$areaID) {
                 $area = ElementalArea::create();
                 $area->OwnerClassName = get_class($this->owner);
-                $area->write();
-                $this->owner->$areaID = $area->ID;
+                // Do not attempt to set the ElementalArea.TopPageID if the owner (e.g. Page)
+                // has not yet been persisted to the database, as this will cause a potentially
+                // large number of unneccessary database queries
+                if ($owner->isInDB() && $area::has_extension(TopPageElementExtension::class)) {
```
GuySartorelli:
```suggestion
                if (!$owner->isInDB() && $area::has_extension(TopPageElementExtension::class)) {
```

GuySartorelli (changes requested):
One small fix - works as expected when this change is made.

src/Extensions/ElementalAreasExtension.php
emteknetnz: Opps, had this locally, forgot to push

GuySartorelli (approved):
LGTM

Issue #1362, opened 2025-07-02
Saving a new page creates two ElementalAreas


wernerkrauss:
### Module version(s) affected

6.0.0

### Description

I stumbled over the issue, that for each new page multiple (two) ElementalAreas are created. One in `ElementalAreasExtension::ensureElementalAreasExist()` which is ok, and one in `onBeforeWrite()` L233ff, when it checks if the OwnerClassName has chagned.

There it seems to get a fresh ElementalArea (ID = 0) from `$ea = $this->owner->$eaRelation();`, cause the Page is currently not saved to DB. This causes to write another ElementalArea and save to DB...

### How to reproduce

fresh SS6 project (with Fluent, two locales) and a very simple BlockPage

```php
<?php

namespace Netwerkstatt\Site\Page;

use Page;
use DNADesign\Elemental\Models\ElementalArea;
use DNADesign\Elemental\Extensions\ElementalPageExtension;

/**
 * Class \Netwerkstatt\Site\Page\BlockPage
 *
 * @property int $ElementalAreaID
 * @method ElementalArea ElementalArea()
 * @mixin ElementalPageExtension
 */
class BlockPage extends Page
{
    /**
     * @var string
     * @config
     */
    private static $table_name = 'BlockPage';


    /**
     * @var string[]
     * @config
     */
    private static $extensions = [
        ElementalPageExtension::class
    ];

}
```

create a new page; 

In ElementalArea table are two Areas per TopPageID, but only one is linked from BlockPage table.

### Possible Solution

I added `ea->exists()` to the if clause; this fixed the problem for me.

```php
//ElementalAreasExtension onBeforeWrite()

        // Update the OwnerClassName on EA if the class has changed
        foreach ($elementalAreaRelations as $eaRelation) {
            $ea = $this->owner->$eaRelation();
            if ($ea->exists() && $ea->OwnerClassName !== $ownerClassName) {
                $ea->OwnerClassName = $ownerClassName;
                $ea->write();
            }
        }
```

### Validations

- [x] Check that there isn't already an issue that reports the same bug
- [x] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## PRs
- https://github.com/silverstripe/silverstripe-elemental/pull/1363
- https://github.com/silverstripe/silverstripe-elemental/pull/1364

emteknetnz:
[...]

PR #1363, opened 2025-07-14
FIX Ensure only a single ElementalArea is created

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-elemental/issues/1362

Duplicate elemental areas were being created when a page was being created for the first time because onBeforeWrite() was being called for both the draft and live record creation.  The duplicate entry was being created because when the reading mode was set to live

```php
            $ea = $this->owner->$eaRelation();
            if ($ea->OwnerClassName !== $ownerClassName) {
                $ea->OwnerClassName = $ownerClassName;
                $ea->write();
```
`$ea` would get an ElementalArea (non-persisted object i.e. ID = 0) and then `write()` it

### DIFF
--- a/src/Extensions/ElementalAreasExtension.php
+++ b/src/Extensions/ElementalAreasExtension.php
@@ -223,24 +223,27 @@ class ElementalAreasExtension extends Extension
         if (!$this->supportsElemental()) {
             return;
         }
-
-        $elementalAreaRelations = $this->owner->getElementalRelations();
-
+        $owner = $this->getOwner();
+        $elementalAreaRelations = $owner->getElementalRelations();
         $this->ensureElementalAreasExist($elementalAreaRelations);
-
-        $ownerClassName = get_class($this->owner);
+        $ownerClassName = $owner->ClassName;
 
         // Update the OwnerClassName on EA if the class has changed
-        foreach ($elementalAreaRelations as $eaRelation) {
-            $ea = $this->owner->$eaRelation();
-            if ($ea->OwnerClassName !== $ownerClassName) {
-                $ea->OwnerClassName = $ownerClassName;
-                $ea->write();
+        // Do not update if using live versioning mode i.e. publishing,
+        // as it can result in a duplicate ElementalArea being created
+        // Note: checking against LIVE instead of DRAFT as get_stage() can return NULL in unit tests
+        if (Versioned::get_stage() !== Versioned::LIVE) {
+            foreach ($elementalAreaRelations as $relation) {
+                $area = $owner->$relation();
+                if ($area->OwnerClassName !== $ownerClassName) {
+                    $area->OwnerClassName = $ownerClassName;
+                    $area->write();
+                }
             }
         }
 
         if (Config::inst()->get(ElementalAreasExtension::class, 'clear_contentfield')) {
-            $this->owner->Content = '';
+            $owner->Content = '';
         }
     }
 
@@ -278,17 +281,23 @@ class ElementalAreasExtension extends Extension
      */
     public function ensureElementalAreasExist($elementalAreaRelations)
     {
-        foreach ($elementalAreaRelations as $eaRelationship) {
-            $areaID = $eaRelationship . 'ID';
-
-            if (!$this->owner->$areaID) {
-                $area = ElementalArea::create();
-                $area->OwnerClassName = get_class($this->owner);
-                $area->write();
-                $this->owner->$areaID = $area->ID;
-            }
+        $owner = $this->getOwner();
+        // Only create the elemental area on draft stage
+        // Note: checking against LIVE instead of DRAFT as get_stage() can return NULL in unit tests
+        if (Versioned::get_stage() === Versioned::LIVE) {
+            return $owner;
         }
-        return $this->owner;
+        foreach ($elementalAreaRelations as $relation) {
+            $field = "{$relation}ID";
+            if ($owner->$field) {
+                continue;
+            }
+            $area = ElementalArea::create();
+            $area->OwnerClassName = $owner->ClassName;
+            $area->write();
+            $owner->$relation = $area;
+        }
+        return $owner;
     }
 
     /**
--- a/tests/Extensions/ElementalAreasExtensionTest.php
+++ b/tests/Extensions/ElementalAreasExtensionTest.php
@@ -15,6 +15,7 @@ use DNADesign\Elemental\Tests\Src\TestVersionedDataObject;
 use DNADesign\Elemental\Extensions\ElementalAreasExtension;
 use SilverStripe\ORM\DB;
 use PHPUnit\Framework\Attributes\DataProvider;
+use SilverStripe\ORM\DataObject;
 
 class ElementalAreasExtensionTest extends SapphireTest
 {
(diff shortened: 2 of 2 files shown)


src/Extensions/ElementalAreasExtension.php
```diff
@@ -223,24 +223,32 @@ protected function onBeforeWrite()
...
-        foreach ($elementalAreaRelations as $eaRelation) {
-            $ea = $this->owner->$eaRelation();
-            if ($ea->OwnerClassName !== $ownerClassName) {
-                $ea->OwnerClassName = $ownerClassName;
-                $ea->write();
+        $doUpdate = true;
+        // Do not update if using live versioning mode i.e. publishing,
+        // as it can result in a duplicate ElementalArea being created
+        if ($owner::has_extension(Versioned::class)
+            && Versioned::get_reading_mode() === 'Stage.' . Versioned::LIVE
```
emteknetnz: @GuySartorelli You mentioned you were going to have have a look during peer review to see if there was a better way to do this check

GuySartorelli:
@emteknetnz Can you please add a brief explanation of the root cause of the bug? That will help both me reviewing this now but also help us if we need to look back at this in the future.

emteknetnz:
Have updated description with explanation

src/Extensions/ElementalAreasExtension.php
```diff
@@ -278,17 +280,21 @@ public function supportsElemental()
...
-            $areaID = $eaRelationship . 'ID';
-
-            if (!$this->owner->$areaID) {
-                $area = ElementalArea::create();
-                $area->OwnerClassName = get_class($this->owner);
-                $area->write();
-                $this->owner->$areaID = $area->ID;
+        $owner = $this->getOwner();
+        // Only create the elemental area on draft stage
+        if (Versioned::get_reading_mode() === 'Stage.' . Versioned::LIVE) {
```
GuySartorelli:
```suggestion
        if (Versioned::get_stage() !== Versioned::DRAFT) {
```
Check the stage specifically rather than the whole reading mode, and explicitly check for "not draft" rather than live (e.g. `null` might be returned if we're in archived reading mode)

src/Extensions/ElementalAreasExtension.php
```diff
@@ -223,24 +223,26 @@ protected function onBeforeWrite()
...
 
         // Update the OwnerClassName on EA if the class has changed
-        foreach ($elementalAreaRelations as $eaRelation) {
-            $ea = $this->owner->$eaRelation();
-            if ($ea->OwnerClassName !== $ownerClassName) {
-                $ea->OwnerClassName = $ownerClassName;
-                $ea->write();
+        // Do not update if using live versioning mode i.e. publishing,
+        // as it can result in a duplicate ElementalArea being created
+        if (Versioned::get_reading_mode() !== 'Stage.' . Versioned::LIVE) {
```
GuySartorelli:
```suggestion
        // Only update if on draft stage, otherwise it can result in a duplicate ElementalArea being created
        if (Versioned::get_stage() === Versioned::DRAFT) {
```

src/Extensions/ElementalAreasExtension.php
```diff
@@ -223,24 +223,26 @@ protected function onBeforeWrite()
...
         }
-
-        $elementalAreaRelations = $this->owner->getElementalRelations();
-
+        $owner = $this->getOwner();
+        $elementalAreaRelations = $owner->getElementalRelations();
         $this->ensureElementalAreasExist($elementalAreaRelations);
-
-        $ownerClassName = get_class($this->owner);
+        $ownerClassName = $owner->ClassName;
```
GuySartorelli, line 229: The `ClassName` might _not actually match_ `get_class` if:
1. Injector has replaced the class, or
2. `ClassName` is explicitly intentionally changed

BUT in both of those cases we want `ClassName` anyway so this is actually an unintended bug fix.

src/Extensions/ElementalAreasExtension.php
```diff
@@ -278,17 +280,21 @@ public function supportsElemental()
...
-                $area->OwnerClassName = get_class($this->owner);
-                $area->write();
-                $this->owner->$areaID = $area->ID;
+        $owner = $this->getOwner();
+        // Only create the elemental area on draft stage
+        if (Versioned::get_reading_mode() === 'Stage.' . Versioned::LIVE) {
+            return $owner;
+        }
+        foreach ($elementalAreaRelations as $relation) {
+            if ($owner->$relation()->exists()) {
```
GuySartorelli: 1. Why is a new loop being added that wasn't there before?
2. Why are we changing to potentially fetching the record from the DB instead of using the `<relation>ID` column
3. Why are we calling `exists()` that wasn't called before?

None of these seem related to the cause of the bug as described in the PR description.

tests/Extensions/ElementalAreasExtensionTest.php
```diff
@@ -48,6 +49,22 @@ protected function setUp(): void
             ->set('disallowed_elements', []);
     }
 
+    protected function tearDown(): void
+    {
+        // For whatever reason, tables are not being properly cleaned between tests
+        // This was noticed when asserting an empty database table in testSingleElementalAreaCreated()
+        // where there was still data in it from testRequireDefaultRecords()
+        $schema = (new DataObject)->getSchema();
```
GuySartorelli:
```suggestion
        $schema = DataObject::getSchema();
```
Static method should be called statically.

tests/Extensions/ElementalAreasExtensionTest.php
```diff
@@ -48,6 +49,22 @@ protected function setUp(): void
             ->set('disallowed_elements', []);
     }
 
+    protected function tearDown(): void
+    {
+        // For whatever reason, tables are not being properly cleaned between tests
+        // This was noticed when asserting an empty database table in testSingleElementalAreaCreated()
+        // where there was still data in it from testRequireDefaultRecords()
```
GuySartorelli, line 56: If that's actually the case, that's a bug that we need to fix, not work around.

GuySartorelli (changes requested):
Looks like this bug also affects CMS 5. Please target `5.4`.

src/Extensions/ElementalAreasExtension.php
emteknetnz: Updated to use get_stage(), though still checking against LIVE as this the method can return NULL for unit-tests, have added inline comment

src/Extensions/ElementalAreasExtension.php
emteknetnz: Updated to use get_stage(), though still checking against LIVE as this the method can return NULL for unit-tests, have added inline comment

src/Extensions/ElementalAreasExtension.php
emteknetnz: > Why is a new loop being added that wasn't there before?

It was there before, I just renamed the variable - was `foreach ($elementalAreaRelations as $eaRelationship) {`

> Why are we changing to potentially fetching the record from the DB instead of using the <relation>ID column
> Why are we calling exists() that wasn't called before?

Have changed back to checkng ID

tests/Extensions/ElementalAreasExtensionTest.php
emteknetnz: Since it's not related to the actual bug being fixed will keep as is and I'll will deal with it in a new issue https://github.com/silverstripe/silverstripe-framework/issues/11786

tests/Extensions/ElementalAreasExtensionTest.php
emteknetnz: Updated

src/Extensions/ElementalAreasExtension.php
GuySartorelli: If it's null, that might represent "archived" reading mode which is not draft stage and should not be used to create an elemental area... can you please confirm that using the "archived" reading mode won't result in a new orphaned elemental area?

emteknetnz:
Closed in favor of https://github.com/silverstripe/silverstripe-elemental/pull/1364

PR #1364, opened 2025-07-17
FIX Ensure only a single ElementalArea is created

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-elemental/issues/1362

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
         "silverstripe/framework": "^5.4",
         "silverstripe/cms": "^5",
         "silverstripe/admin": "^2.0.1",
-        "silverstripe/versioned": "^2",
+        "silverstripe/versioned": "^2.4.5",
         "silverstripe/versioned-admin": "^2",
         "silverstripe/graphql": "^5",
         "symbiote/silverstripe-gridfieldextensions": "^4",
@@ -31,6 +31,7 @@
     "require-dev": {
         "silverstripe/documentation-lint": "^1",
         "silverstripe/recipe-testing": "^3",
+        "silverstripe/behat-extension": "^5.6.1",
         "silverstripe/frameworktest": "^1",
         "silverstripe/standards": "^1",
         "phpstan/extension-installer": "^1.3"
--- a/docs/en/02_advanced_setup.md
+++ b/docs/en/02_advanced_setup.md
@@ -50,6 +50,12 @@ class BlockPage extends Page
 }
 ```
 
+> [!IMPORTANT]
+> If you programatically publish a `Page` or a `DataObject` that has [`ElementalPageExtension`](api:DNADesign\Elemental\Extensions\ElementalPageExtension)
+> applied to it, ensure that you call [`RecursivePublishable::publishRecursive()`](api:SilverStripe\Versioned\RecursivePublishable::publishRecursive())
+> and not [`Versioned::publishSingle()`](api:SilverStripe\Versioned\Versioned::publishSingle())
+> to ensure the `ElementalArea` and its child elements are published correctly.
+
 ### Block setup
 
 It is possible to share blocks between pages, but this may be a little bit tricky when it comes to content editing.
--- a/src/Extensions/ElementalAreasExtension.php
+++ b/src/Extensions/ElementalAreasExtension.php
@@ -19,6 +19,7 @@ use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\RelatedData\StandardRelatedDataService;
 use SilverStripe\Versioned\Versioned;
 use SilverStripe\View\ViewableData;
+use SilverStripe\Control\Director;
 
 /**
  * This extension handles most of the relationships between pages and element
@@ -226,24 +227,22 @@ class ElementalAreasExtension extends DataExtension
         if (!$this->supportsElemental()) {
             return;
         }
-
-        $elementalAreaRelations = $this->owner->getElementalRelations();
-
+        $owner = $this->getOwner();
+        $elementalAreaRelations = $owner->getElementalRelations();
         $this->ensureElementalAreasExist($elementalAreaRelations);
 
-        $ownerClassName = get_class($this->owner);
-
-        // Update the OwnerClassName on EA if the class has changed
-        foreach ($elementalAreaRelations as $eaRelation) {
-            $ea = $this->owner->$eaRelation();
-            if ($ea->OwnerClassName !== $ownerClassName) {
-                $ea->OwnerClassName = $ownerClassName;
-                $ea->write();
+        if ($this->allowAlteringElementalArea()) {
+            $ownerClassName = $owner->ClassName;
+            foreach ($elementalAreaRelations as $relation) {
+                $area = $owner->$relation();
+                if ($area->OwnerClassName !== $ownerClassName) {
+                    $area->OwnerClassName = $ownerClassName;
+                    $area->write();
+                }
             }
         }
-
         if (Config::inst()->get(ElementalAreasExtension::class, 'clear_contentfield')) {
-            $this->owner->Content = '';
+            $owner->Content = '';
         }
     }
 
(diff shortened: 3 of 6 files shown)


docs/en/02_advanced_setup.md
```diff
@@ -50,6 +50,12 @@ class BlockPage extends Page
 }
 ```
 
+> [!IMPORTANT]
```
emteknetnz, line 53: I haven't changed any logic in this PR that would break any sites that programatically call publishSingle() somewhere for something - however it's worth calling out that the intended way for elemental to be run with it's nested data hierarchy is publishRecursive()

GuySartorelli:
It looks like you've replaced #1363 with this PR.
When replacing a PR which has already had some review, please remember to link to the old PR so that the old review comments can be checked as part of reviewing the new PR.

Otherwise the reviewer has to start from scratch :sweat_smile: 

In this case it's especially important as the original PR description contains context this PR description lacks.

src/Extensions/ElementalAreasExtension.php
```diff
@@ -340,4 +344,23 @@ public function requireDefaultRecords()
...
     }
+
+    /**
+     * Whether it's OK to alter the ElementalArea based the current versioned reading mode
+     */
+    private function canAlterElementalArea(): bool
+    {
+        $readingMode = Versioned::get_reading_mode();
+        // Reading mode is Stage.Stage, which is the normal state on the CMS
+        if ($readingMode === 'Stage.' . Versioned::DRAFT) {
```
GuySartorelli: Why are we back to using the reading mode? Please use stage as per https://github.com/silverstripe/silverstripe-elemental/pull/1363#discussion_r2209049813

src/Extensions/ElementalAreasExtension.php
```diff
@@ -340,4 +344,23 @@ public function requireDefaultRecords()
 
         $this->owner->extend('onAfterRequireDefaultElementalRecords');
     }
+
+    /**
+     * Whether it's OK to alter the ElementalArea based the current versioned reading mode
+     */
+    private function canAlterElementalArea(): bool
```
GuySartorelli: Dont name this `can*` - methods starting with `can` in a `DataObject` or `Extension` usually imply they are checking if a member can perform an action.

src/Extensions/ElementalAreasExtension.php
```diff
@@ -340,4 +344,23 @@ public function requireDefaultRecords()
...
+        // Reading mode is Stage.Stage, which is the normal state on the CMS
+        if ($readingMode === 'Stage.' . Versioned::DRAFT) {
+            return true;
+        }
+        // When running unit tests he Versioned reading mode is empty string by default as a result
+        // of Versioned::reset() which is called because of the initial flush
+        // Treat this the same as Stage.Stage
+        if ($readingMode === '' && Director::is_cli()) {
+            return true;
+        }
```
GuySartorelli: CLI can also be `sake db:build` or `sake tasks:some-specific-task` - but in those cases the reading mode should be set. If the reading mode isn't set in those scenarios, we probably either have a bug in the way sake bootstraps versioning, or someone is doing something weird.

I assume this is specifically for tests. We should avoid having weird stuff like this in tests. Instead, just set a valid reading mode in the test.

GuySartorelli (changes requested):

src/Extensions/ElementalAreasExtension.php
emteknetnz: Yup will be Stage.Stage for a BuildTask

Will update tests

src/Extensions/ElementalAreasExtension.php
emteknetnz: Updated

src/Extensions/ElementalAreasExtension.php
emteknetnz: Updated

src/Extensions/ElementalAreasExtension.php
emteknetnz: Actually have reversed, will keep the `$readingMode === '' && Director::is_cli()` and not set in tests - reason being is that this is also required for behat :-/

I think we'll just need to keep this as is with a big inline explanation of why it's here

src/Extensions/ElementalAreasExtension.php
emteknetnz: Reverted back to reading mode - see https://github.com/silverstripe/silverstripe-elemental/pull/1364#discussion_r2218012093

src/Extensions/ElementalAreasExtension.php
GuySartorelli: I'll go find out why this is happening incorrectly in behat. Might be a bit, it's not my top priority.
We should not have weird stuff like this (which could affect actual projects running in live mode) in source code just to support tests unless it's actually unavoidable - which means we need to know _why_ this is happening.

GuySartorelli: Behat calls `reset()` on all classes that implement the `Resettable` interface in `SilverStripeContext::before()`.
In `Versioned::reset()` this sets `reading_mode` to an empty string.

This means that for the CLI context (which is used for creating fixture records), the reading mode is empty and the stage is empty.

Adding this to the bottom of that method seems to work:
```php
if (class_exists(Versioned::class)) {
    Versioned::set_stage(Versioned::DRAFT);
}
```

emteknetnz: Since this could easily affect other modules, I've split this off as a new card https://github.com/silverstripe/silverstripe-versioned/issues/519

For now, will keep this as is

GuySartorelli: I'd prefer not to merge this as-is.
I'll get the other card done. All it really needs is a sink run so it's not like it's going to result in much delay.

emteknetnz: I've merged the other card and have silverstripe/versioned 2.4.x-dev 859e9d0 and silverstripe/behat-extension 5.6.1 locally

I removed the workaround, behat seems to work, though phpunit tests still fails e.g. ElementalAreasExtensionTest::testRequireDefaultRecords()

When that particular unit test fails Versioned::get_reading_mode is a blank string, which is set by Versioned::reset()

```
ElementalAreasExtensionTest::setUp()
SapphireTest::setup()
SapphireTestState::setUp()
FlushableTestState::setUp()
Versioned::reset()
=> Versioned::$reading_mode = '';
```

I'm keeping the workaround in place

composer.json
```diff
@@ -31,10 +31,14 @@
     "require-dev": {
         "silverstripe/documentation-lint": "^1",
         "silverstripe/recipe-testing": "^3",
+        "silverstripe/behat-extension": "^5.6.1",
```
emteknetnz, line 34: This includes the fix for the workaround https://github.com/silverstripe/silverstripe-behat-extension/pull/308

composer.json
```diff
@@ -31,10 +31,14 @@
...
         "silverstripe/frameworktest": "^1",
         "silverstripe/standards": "^1",
         "phpstan/extension-installer": "^1.3"
     },
+    "conflict": {
+        "silverstripe/versioned": "<2.4.4"
```
emteknetnz: This includes a fix for the workaround for phpunit - https://github.com/silverstripe/silverstripe-versioned/pull/520

Since it only for a unit test, and required API, I've put it in as a conflict rather than a requirement

src/Extensions/ElementalAreasExtension.php
GuySartorelli: Where is that call to `Versioned::reset()` happening?

emteknetnz: My comment above lists a pseudo stack trace

GuySartorelli: Oh I didn't realise that was meant to be a stack trace, thanks for clearing that up. I'll go see why that's interfering with setting the stage in the separate PRs.

composer.json
GuySartorelli: This module already has `silverstripe/versioned` as a dependency, so having this conflict does the same as setting the constraint in `require` to `^2.4.5` which is what we should do instead.

Composer doesn't know this is only for tests and there's no `conflict-dev`, so the end result is the same.

src/Extensions/ElementalAreasExtension.php
GuySartorelli: Changing this globally proved to be too disruptive - see https://github.com/silverstripe/silverstripe-versioned/issues/519#issuecomment-3105055016

Please set the stage to `Versioned::DRAFT` in this module for the tests that need it, and remove the unnecessary workaround (i.e. the change you made before you noticed behat was affected, since it isn't anymore that change is now valid).

composer.json
emteknetnz: Updated

GuySartorelli (approved):
LGTM

PR #1365, opened 2025-07-29
MNT Clarify why data is cleared between tests

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/11786

[SecurityDefaultAdminTest](https://github.com/silverstripe/silverstripe-framework/blob/5.4/tests/php/Security/SecurityDefaultAdminTest.php#L27) does a similar thing. Those tests also assert record counts, and also fail if clearAllData() is not called


GuySartorelli (approved):
LGTM. Good find in `clearAllData()`.

PR #1373, opened 2025-09-21
FIX: remove ->owner where not needed

sunnysideup:
## Description

I just noticed a weird reference to owner that is usually applied to extensions. I could be wrong, but it seems not needed here. At the same time, I simplified the code a little. 

## Manual testing steps

Not sure

## Issues

None AFAIK (I searched)

## Pull request checklist

- [ x] The target branch is correct
- [x ] All commits are relevant to the purpose of the PR (e.g. no debug statements, unrelated refactoring, or arbitrary linting)
- [x ] The commit messages follow our [commit message guidelines](https://docs.silverstripe.org/en/contributing/code/#commit-messages)
- [x ] The PR follows our [contribution guidelines](https://docs.silverstripe.org/en/contributing/code/)
- [ x] Code changes follow our [coding conventions](https://docs.silverstripe.org/en/contributing/coding_conventions/)
- [x ] This change is covered with tests (or tests aren't necessary for this change)
- [ x] Any relevant User Help/Developer documentation is updated; for impactful changes, information is added to the changelog for the intended release
- [ x] CI is green

### DIFF
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -744,8 +744,8 @@ JS
         $anchors = [$this->getAnchor()];
         $anchorRegex = "/\\s+(name|id)\\s*=\\s*([\"'])([^\\2\\s>]*?)\\2|\\s+(name|id)\\s*=\\s*([^\"']+)[\\s +>]/im";
         $allFields = DataObject::getSchema()->fieldSpecs($this);
-        foreach ($allFields as $field => $fieldSpec) {
-            $fieldObj = $this->owner->dbObject($field);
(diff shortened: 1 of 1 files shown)


emteknetnz (approved):
This is a safe change, though it has no functional impact, so it's fine to target the `6` branch

While `$this->owner` is wrong, in this context it will just resolve to `$this`

I've checked locally that [BaseElementTest::testGetAnchorsInContent()](https://github.com/silverstripe/silverstripe-elemental/blob/6/tests/BaseElementTest.php#L301) still passes with this change

PR #1376, opened 2025-10-02
ENH Hide JS font icons from screen readers

GuySartorelli:
- Replaces any `<i>` being used as an icon with `span` for consistency and semantic correctness
- Hides decorative icons from screen readers (usually by adding an inner `span`)
- In a few places, adds `aria-label` (will be noted with PR comments)

## Issue

- https://github.com/silverstripe/silverstripe-admin/issues/1957


client/src/components/ElementEditor/Header.js
```diff
@@ -192,7 +192,7 @@ class Header extends Component {
               expandable={expandable}
             />
           </div>
-          {!type.broken && <i className={expandCaretClasses} title={expandTitle} />}
+          {!type.broken && <span className={expandCaretClasses} aria-label={expandTitle} title={expandTitle} />}
```
GuySartorelli, line 195: Explicit `aria-label` here to match the `title` since some screen readers won't read a `title` attribute.

client/src/components/ElementEditor/AddNewButton.js
```diff
@@ -32,7 +32,7 @@ class AddNewButton extends Component {
       id: `ElementalArea${areaId}_AddButton`,
       color: 'primary',
       onClick: this.toggle,
-      className: 'font-icon-plus',
+      icon: 'plus',
```
GuySartorelli, line 35: `icon` prop already existed through old `IconHOC` so no need to bump composer dep

GuySartorelli:
CI failures are pre-existing - see https://github.com/silverstripe/silverstripe-elemental/actions/runs/18176530538/

emteknetnz (approved):

PR #1383, opened 2025-10-29
ENH Use javascript default parameters instead of defaultProps

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/2037


client/src/components/ElementActions/AbstractAction.js
```diff
@@ -7,11 +7,18 @@ import { elementTypeType } from 'types/elementTypeType';
...
-const AbstractAction = (props) => {
-  const { className, title, label } = props;
-
+const AbstractAction = ({
+  disabled = false,
+  className,
+  title,
+  label,
+  ...props
+}) => {
```
GuySartorelli, line 16: I don't really like the addition of `className, title, label` into the constructor but it doesn't matter enough to hold up merging.

client/src/components/ElementEditor/Element.js
```diff
@@ -475,10 +475,6 @@ Element.propTypes = {
   increment: PropTypes.number.isRequired,
 };
 
-Element.defaultProps = {
-  element: null,
```
GuySartorelli, line 479: Why wasn't this assigned in the constructor? Won't it be `undefined` if not set? Without spending effort looking into whether that prop is ever compared in a way where that distinction matters, I'd rather just pop it into the constructor.

GuySartorelli (changes requested):
Just the one thing below. Don't bother with the note above it's not consequential.

client/src/components/ElementEditor/Element.js
emteknetnz: It's kind of nonsensical as a default prop, and equally pointless as a default javascript parameter of null. It has a propType of elementType which is 

```js
const elementType = PropTypes.shape({
  id: PropTypes.number,
  title: PropTypes.string,
  blockSchema: PropTypes.object,
  inlineEditable: PropTypes.bool,
  published: PropTypes.bool,
  liveVersion: PropTypes.bool,
  version: PropTypes.number
});
```

Nothing in this js file will work if it's a non-object and missing all these keys

GuySartorelli: Probably the proptype should be "required" then but I bet we have a ton of incorrect, missing, unnecessary etc proptypes so that's a whole separate kettle of fish.

GuySartorelli (approved):
LGTM

PR #1389, opened 2025-11-26
ENH PHP 8.5 Support

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/455

The following changes were made based on the [PHP 8.5 changelog](https://www.php.net/ChangeLog-8.php):

- Using null as an array offset or when calling array_key_exists() is now deprecated.


GuySartorelli (approved):
LGTM

Issue #1393, opened 2025-12-15
Convert more elemental class components to functional


emteknetnz:
There are react class component(s) that needs to be converted to react functional component(s).

- [ ] Content
- [ ] ElementActions
- [ ] ElementEditor
- [ ] InlineEditForm
- [ ] Summary

## Acceptance criteria
- [ ] Component(s) are converted to a functional component using hooks
- [ ] Jest test coverage is increased
- [ ] All existing functionality is preserved

## PRs
- https://github.com/silverstripe/silverstripe-elemental/pull/1394

GuySartorelli:
PR merged

PR #1394, opened 2025-12-16
ENH Refactor more react components from class to functional

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-elemental/issues/1393


client/src/components/ElementEditor/ElementActions.js
```diff
@@ -11,30 +11,31 @@ import AbstractAction from 'components/ElementActions/AbstractAction';
  * Element actions is a dropdown menu containing links to inline editing forms for each
  * of the element's primary tabs, as well as operations such as save, publish, archive etc
  */
-class ElementActions extends Component {
-  constructor(props) {
-    super(props);
-    this.handleEditTabsClick = this.handleEditTabsClick.bind(this);
-  }
-
+const ElementActions = ({
```
emteknetnz, line 14: The following props were added - previously referred to via this.props (in render()) though not in propTypes:
- id
- ActionMenuComponent
- children (though it's always a prop)

client/src/components/ElementEditor/ElementEditor.js
```diff
@@ -19,55 +19,69 @@ export const ElementEditorContext = createContext(null);
...
-      elements: null,
-      isLoading: true,
-      dragging: false,
-    };
-
-    this.handleDragStart = this.handleDragStart.bind(this);
-    this.handleDragEnd = this.handleDragEnd.bind(this);
-    this.fetchElements = this.fetchElements.bind(this);
-  }
+const ElementEditor = ({
```
emteknetnz, line 22: The following props were added - previously referred to via this.props though not in propTypes:
- forceRefetchElements

client/src/components/ElementEditor/ElementEditor.js
```diff
@@ -19,55 +19,69 @@ export const ElementEditorContext = createContext(null);
...
+  elementTypes,
+  allowedElements,
+  sharedObject,
+  isLoading,
+  actions,
+  forceRefetchElements,
+}) => {
+  const [dragging, setDragging] = useState(false);
+  const [elements, setElements] = useState(null);
+  // isLoadingState is ununsed, possibly because isLoading prop is being incorrectly used
```
emteknetnz: While this state is unused, it's possibly incorrectly unused, though I also don't want to investigate it right now because I want to retain this existing logic

client/src/components/ElementEditor/ElementEditor.js
```diff
@@ -19,55 +19,69 @@ export const ElementEditorContext = createContext(null);
  * The ElementEditor is used in the CMS to manage a list or nested lists of
  * elements for a page or other DataObject.
  */
-class ElementEditor extends PureComponent {
-  constructor(props) {
-    super(props);
-
-    this.state = {
-      dragTargetElementId: null,
-      dragSpot: null,
```
emteknetnz, line 28: draftTargetElementId and dragSpot were unused in the old component, so were not copied over

client/src/components/ElementEditor/ElementEditor.js
```diff
@@ -84,108 +98,74 @@ class ElementEditor extends PureComponent {
...
-      sharedObject,
-      isLoading,
-    } = this.props;
-    const { dragging, elements } = this.state;
+  }, [forceRefetchElements]);
 
+  /**
+   * Fetch elements on mount
+   */
+  useEffect(() => {
```
emteknetnz: There wasn't previously componentDidMount(), however it did used to call fetchElements() in the render() method

client/src/components/ElementEditor/InlineEditForm.js
```diff
@@ -8,25 +8,22 @@ import i18n from 'i18n';
...
-    this.handleLoadingError = this.handleLoadingError.bind(this);
-    this.handleSubmit = this.handleSubmit.bind(this);
-
-    this.state = {
-      loadingError: null
-    };
-  }
-
-  handleLoadingError() {
+const InlineEditForm = ({
```
emteknetnz, line 11: The following props were added - previously referred to via this.props though not in propTypes:
- onFormInit
- formHasState

client/src/components/ElementEditor/InlineEditForm.js
```diff
@@ -8,25 +8,22 @@ import i18n from 'i18n';
...
+  onClick,
+  onFormInit,
+  formHasState,
+  notVisible,
+  handleLoadingError,
+  onFormSchemaSubmitResponse,
+}) => {
+  const [loadingError, setLoadingError] = useState(null);
+
+  const handleLoadingErrorFn = () => {
```
emteknetnz, line 32: Renamed function to because there's a handeLoadingError prop

client/src/components/ElementEditor/InlineEditForm.js
```diff
@@ -8,26 +8,23 @@ import i18n from 'i18n';
...
+  extraClass,
+  onClick,
+  onFormInit,
+  formHasState,
+  notVisible,
+  handleLoadingError,
+  onFormSchemaSubmitResponse,
+}) => {
+  const [loadingError, setLoadingError] = useState(null);
+  const onFormSchemaSubmitResponseRef = useRef(onFormSchemaSubmitResponse);
```
emteknetnz, line 26: This component was hard to refactor because it caused a failing behat test - the toast on saving the page was not showing, but only in behat, not in manual testing. 

https://github.com/silverstripe/silverstripe-elemental/blob/6/tests/Behat/features/edit-block-element.feature#L111 (ignore the double `""`, it makes no difference, and I didn't want to fix in this PR as I want to keep behat tests identical)

After much experimentation, the following worked:
- Wrapping the onFormSchemaSubmitResponse in useRef()
- Adding useCallback() to handleSubmit()
- Not wrapping the exported component in memo(), even though the class component extended PureComponent

I honestly don't fully understand why this works, or why it broke in the first place. I worked this solution out by trial and error, rather than comprehension.

client/src/components/ElementEditor/ElementEditor.js
GuySartorelli: My guess is the state is legacy, and changed after that to use the prop instead.

In any case instead of keeping the unused and renamed state here with a comment, please remove the state and comment, and instead raise a card so we can investigate and add the state in (and use it) if that's appropriate. We don't track comments, but we do track cards.

GuySartorelli: Oop nevermind, I see this is being set. Please still open a new card but leave the state here.
Please also update the comment to say it does get set.

client/src/components/ElementEditor/ElementEditor.js
GuySartorelli: This will now only try calling this on mount, but it used to try it on render whenever `elements` was null.
Are we 100% sure this change won't result in a regression?

client/src/components/ElementEditor/ElementEditor.js
```diff
@@ -84,108 +98,74 @@ class ElementEditor extends PureComponent {
...
-      </ElementEditorContext.Provider>
-    </div>;
+  }, []);
+
+  // Memoize provider value to avoid unnecessary re-renders
+  const providerValue = useMemo(() => ({
+    fetchElements,
+    actions,
+    // fetchElements will be recreated each time areraId changes, so
+    // it's needed in the dep array, otherwise fetchElements will be stale
```
GuySartorelli, line 129: Should `areaID` be in the dependency array instead, to prevent re-rendering child elements unnecessarily if `fetchElements` changes for some other reason?

client/src/components/ElementEditor/InlineEditForm.js
GuySartorelli: Presumably this needs an effect to update the ref in case the `onFormSchemaSubmitResponse` prop updates.

client/src/components/ElementEditor/InlineEditForm.js
```diff
@@ -8,26 +8,23 @@ import i18n from 'i18n';
...
+  onFormSchemaSubmitResponseRef.current = onFormSchemaSubmitResponse;
```
GuySartorelli: Doesn't `const onFormSchemaSubmitResponseRef = useRef(onFormSchemaSubmitResponse);` already set the `current` value to `onFormSchemaSubmitResponse`?

If not, what does passing the value into `useRef` do?

GuySartorelli: Or is this setting the value even when the prop changes, despite not being an effect? If that's the case it probably warrants an in-code comment to just clarify that's what's going on.

GuySartorelli (changes requested):

client/src/components/ElementEditor/ElementEditor.js
emteknetnz: Added comment and created new issue

https://github.com/silverstripe/silverstripe-elemental/issues/1395

client/src/components/ElementEditor/ElementEditor.js
emteknetnz: Combined with the forceRefetchElements useEffect block above

client/src/components/ElementEditor/InlineEditForm.js
emteknetnz: Updated

client/src/components/ElementEditor/InlineEditForm.js
emteknetnz: It seemed necessary to in order to get the behat test to pass. As mentioned this was discovered via experimentation rather than comprehension i.e. I don't know why it works :-)

I've added an inline comment to mention this ref is used in order to get a behat test to pass

client/src/components/ElementEditor/ElementEditor.js
emteknetnz: Seems convoluted doing it that way?

Better way seems like it would be to wrap fetchElements() in useCallback() and put areaId in the dependency array of that, though I'm trying to NOT wrap existing methods in useCallback() / useMemo() as while that's a good optimisation, it does make peer-reviewing these sorts of PRs harder

context provider values though do need to be wrapped in useMemo() otherwise eslint will complain about child re-renders

GuySartorelli (approved):
LGTM

Issue #1399, opened 2026-01-20
`BaseElement::getCMSEditLink()` returns invalid link for non-page parents


GuySartorelli:
If the parent of an elemental block is a non-SiteTree object, and the elemental block _is_ inline editable, the result of `getCMSEditLink()` has an extra `/ItemEditForm` on the end of the URL. This results in an error ("unable to handle sub-urls") instead of showing the edit form of the parent object.

`ItemEditForm` should only be appended to the URL when for non-inline-editable blocks or when `$directLink` is true, as it is part of the URL for the ElementalArea's gridfield.

## PRs

- https://github.com/silverstripe/silverstripe-elemental/pull/1401

PR #1401, opened 2026-01-20
FIX Get valid CMS edit link for non-page data object

GuySartorelli:
The edit link for an elemental block on a DataObject that didn't subclass SiteTree was incorrectly created.

## Issue

- https://github.com/silverstripe/silverstripe-elemental/issues/1399

### DIFF
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -916,7 +916,7 @@ JS
         if ($page instanceof SiteTree) {
             $link = $page->CMSEditLink();
         } elseif (ClassInfo::hasMethod($page, 'CMSEditLink')) {
-            $link = Controller::join_links($page->CMSEditLink(), 'ItemEditForm');
+            $link = $page->CMSEditLink();
         }
         // In-line editable blocks should just take you to the page.
         // Editable ones should add the suffix for detail form.
@@ -931,10 +931,12 @@ JS
                     $this->ID,
                     'edit'
                 );
-            } else {
-                // If $page is not a Page, then generate $link base on $page->CMSEditLink()
+            } elseif ($link) {
+                // If $page is not a Page, then generate $link based on $page->CMSEditLink()
+                // for a gridfield edit form
                 return Controller::join_links(
                     $link,
+                    'ItemEditForm',
                     'field',
                     $relationName,
                     'item',
--- a/tests/BaseElementTest.php
+++ b/tests/BaseElementTest.php
@@ -310,43 +310,51 @@ class BaseElementTest extends FunctionalTest
     public function getElementCMSLinkDataProvider()
     {
         return [
-            // Element in DataObject with $directLink === true
-            'element1' => [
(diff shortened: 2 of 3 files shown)


src/Models/BaseElement.php
```diff
@@ -916,7 +916,7 @@ private function getElementCMSLink(bool $directLink)
         if ($page instanceof SiteTree) {
             $link = $page->CMSEditLink();
         } elseif (ClassInfo::hasMethod($page, 'CMSEditLink')) {
-            $link = Controller::join_links($page->CMSEditLink(), 'ItemEditForm');
+            $link = $page->CMSEditLink();
```
GuySartorelli, line 919: This value is used directly if the block is inline editable, so it must no include the `ItemEditForm` part which is for the ElementalArea's gridfield.

tests/BaseElementTest.php
GuySartorelli, line 1: Several problems fixed with these test scenarios:

1. Now using meaningful keys for scenarios array
1. Using regex instead of `assertStringContainsString` so we can test more of the URL and to not rely on hardcoded ID values
1. Removed hardcoded domain name which is related to how `CMSEditLinkExtension` builds URLs - nothing to do with this module's logic, and not consistent with the URL for a non-inline-editable block on a SiteTree owner
1. Correctly expect no `ItemEditForm` in the URL for inline-editable blocks.
1. Added missing scenarios
    - non-inline-editable block on a SiteTree owner
    - direct link on an owner with no CMS edit link

tests/BaseElementTest.php
```diff
@@ -360,9 +368,9 @@ public function testCMSEditLink(string $class, string $element, ?string $link, b
         $editLink = $object->CMSEditLink($directLink);
 
         if ($link) {
-            $this->assertStringContainsString($link, $editLink);
+            $this->assertMatchesRegularExpression($link, $editLink);
         } else {
-            $this->assertNull($link);
+            $this->assertNull($editLink);
```
GuySartorelli, line 373: We should be asserting about the result of calling `CMSEditLink()`, not the scenario's expected link value.

tests/ElementalAreaDataObjectTest.yml
GuySartorelli, line 1: Changes here are so we can test a non-inline-editable block on a SiteTree owner

src/Models/BaseElement.php
```diff
@@ -931,10 +931,12 @@ private function getElementCMSLink(bool $directLink)
                     $this->ID,
                     'edit'
                 );
-            } else {
-                // If $page is not a Page, then generate $link base on $page->CMSEditLink()
+            } elseif ($link) {
```
GuySartorelli, line 934: This new condition is required because if there is no cms edit link (i.e. `$link` is falsy) we should just be returning `null`. Before it was returning a broken URL.

emteknetnz (approved):

Issue #1395, opened 2026-01-04
Investigate if ElementEditor isLoading should use prop or internal state


emteknetnz:
Follow on from https://github.com/silverstripe/silverstripe-elemental/pull/1394/changes#r2659976954

isLoading is a prop, though there is also an internal state

Work out if the prop or the internal state should be used, and remove the one that is not needed

## PRs
- https://github.com/silverstripe/silverstripe-elemental/pull/1404

GuySartorelli:
PR merged

PR #1404, opened 2026-01-22
ENH Remove unused prop from ElementEditor

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-elemental/issues/1395

Removed loading prop and using state var instead.
- `isLoadingState` was properly tracking loading (set to true  fetch, false when done) but never used
- `isLoading` prop was being passed to child `ElementList` to show the loading indicator, but never provided by its parent `entwine.js` - this meant the loading indicator never showed

Also added a 300 millisecond buffer before showing loading indicator to prevent FOUT now that the loading indicator actually works


client/src/components/ElementEditor/ElementEditor.js
```diff
@@ -111,6 +108,33 @@ const ElementEditor = ({
...
+    }
+    return () => {
+      if (timeoutId) {
+        clearTimeout(timeoutId);
+      }
+      if (timeoutId2) {
+        clearTimeout(timeoutId2);
+      }
+    };
+  }, [isLoadingState]);
```
GuySartorelli: Is is very hard to parse out what's happening. I know the comment says something about preventing "infinite loading state" but how would that happen, and what does that second timer actually do? 

If `isLoadingState` is set to `true`, we should have a visual loading indicator (if it's been that way for 300ms+) - no second timeout should be necessary.

client/src/components/ElementEditor/ElementEditor.js
```diff
@@ -111,6 +108,33 @@ const ElementEditor = ({
...
   };
 
+  /**
+   * Delay loading indicator to prevent FOUT
+   */
+  useEffect(() => {
+    let timeoutId;
+    let timeoutId2;
+    if (!isLoadingState) {
+      setShowLoadingIndicator(false);
```
GuySartorelli: We should also clear any pre-existing timeout. Otherwise won't we have this scenario?:

1. isLoadingState set to true. We start a timeout which will show the loading indicator after 300ms
1. isLoadingState set to false. Set `showLoadingIndicator` to false, which has no effect.
1. Because the timeout wasn't cleared, we still show the loading indicator after 300ms

That doesn't seem to happen, but it's not clear from this code _why_ that doesn't happen.

client/src/components/ElementEditor/ElementEditor.js
```diff
@@ -26,16 +26,13 @@ const ElementEditor = ({
...
-  isLoading,
   actions,
   forceRefetchElements,
 }) => {
   const [dragging, setDragging] = useState(false);
   const [elements, setElements] = useState(null);
-  // isLoadingState is unused, possibly because isLoading prop is being incorrectly used
-  // setIsLoadingState is still used to update the state internally
-  // eslint-disable-next-line no-unused-vars
   const [isLoadingState, setIsLoadingState] = useState(true);
```
GuySartorelli: We should rename this to `loading` or `isLoading` now that we don't have a naming conflict with the prop.

GuySartorelli (changes requested):

client/src/components/ElementEditor/ElementEditor.js
emteknetnz: Changed to loading

client/src/components/ElementEditor/ElementEditor.js
emteknetnz: Refactored

emteknetnz:
Failing elemental behat test is [existing](https://github.com/silverstripe/silverstripe-elemental/actions/runs/21267756239)

client/src/components/ElementEditor/ElementEditor.js
emteknetnz: Block has been refactored

GuySartorelli (approved):
LGTM

Issue #1405, opened 2026-01-23
Failing behat test


emteknetnz:
`tests/Behat/features/add-block-element-to-data-object.feature:26 (on line 35)`

https://github.com/silverstripe/silverstripe-elemental/actions/runs/21267293569/job/61209511796#step:12:436 (6.1 build)

## Kitchen sink CI
- https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/21273236624

## PRs
- https://github.com/silverstripe/silverstripe-frameworktest/pull/257
- https://github.com/silverstripe/silverstripe-elemental/pull/1406

GuySartorelli:
Please document in current (CMS 5 and 6.1) docs that a CMS edit link is required for this scenario as requested in https://github.com/silverstripe/silverstripe-elemental/pull/1406#issuecomment-3808433141

emteknetnz:
It's already documented in https://docs.silverstripe.org/en/5/optional_features/elemental/examples/elementalarea_with_dataobject/

```markdown
If you are using ElementalArea together with DataObject, it is important to define the CMSEditLink()
method in the class. BaseElement relies on a valid CMS link being available in the parent DataObject
- in this case, BlogPost. It is used to navigate directly to the editing section of the particular element.
```

PR #1406, opened 2026-01-23
MNT Fix behat test

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-elemental/issues/1405

Requires https://github.com/silverstripe/silverstripe-frameworktest/pull/257 for behat to go green

### DIFF
--- a/tests/Behat/features/add-block-element-to-data-object.feature
+++ b/tests/Behat/features/add-block-element-to-data-object.feature
@@ -25,6 +25,7 @@ Feature: Add elements in the CMS DataObject
 
   Scenario: I can add non-inline-editable elements to the DataObject
     Given content blocks are not in-line editable
+        And I add an extension "SilverStripe\FrameworkTest\Elemental\Model\ElementalBehatTestObjectCMSEditLinkExtension" to the "SilverStripe\FrameworkTest\Elemental\Model\ElementalBehatTestObject" class
         When I go to "/admin/elemental-behat-test-admin/"
         Then I press the "Add Elemental Behat Test Object" button
         Then I press the "Create" button


GuySartorelli (commented):
It looks like this PR would be hiding a bug - we shouldn't need a CMSEditLink in order to add a non-inline-editable block to a DataObject.

emteknetnz:
Perhaps? I don't have a lot of context on this one, I just traced it back to this [PR](https://github.com/silverstripe/silverstripe-frameworktest/pull/256) so I restored edit link functionality.

Do you have context on this one?

GuySartorelli:
I don't have enough context to know what the underlying cause of the failure is, but I do know that we shouldn't need an edit link on the parent in order for the steps in this behat feature to pass.

emteknetnz:
> It looks like this PR would be hiding a bug - we shouldn't need a CMSEditLink in order to add a non-inline-editable block to a DataObject.

Not to add, to edit the block - failing test is https://github.com/silverstripe/silverstripe-elemental/blob/6/tests/Behat/features/add-block-element-to-data-object.feature#L26 which has `Given content blocks are not in-line editable`

It needs the CMSEditLink() here https://github.com/silverstripe/silverstripe-elemental/blob/6.1/src/Models/BaseElement.php#L877

Note that $page the parent DataObject, not specifcally a Page, and I think that `// In-line editable blocks should just take you to the page.` is supposed to be `Non-inline editable`

If $link is null (which happens when we are missing that extension I added in this PR), then there's no logic to handle it .. which makes sense? Seems like you'd need a nested `/ItemEditForm` path to generate a form to edit the block?

Call stack to get to that code is

BaseElement::getElementCMSLink()
BaseElement::getCMSEditLink()
BaseElement::getEditLink()
BaseElement::provideBlockSchema()
BaseElement::getBlockSchema()
ElementaAreaController::apiReadElements()

GuySartorelli:
Ahhhh yup that makes sense. In that case I think it should fail with a hard exception that has a clear message, instead of just trying to navigate to... somewhere?

We should also ensure that the documentation calls our this requirement.

emteknetnz:
Have created a [new issue](https://github.com/silverstripe/silverstripe-elemental/issues/1407) for the hard exception - behavioral change so needs to be in a new minor

We still need these PRs to fix 6.1 CI

GuySartorelli (approved):
LGTM

Issue #1407, opened 2026-01-28
Throw exception if non-sitetree parent CMSEditLink is empty


emteknetnz:
Follow on from https://github.com/silverstripe/silverstripe-elemental/pull/1406#issuecomment-3808433141

If a non-inline editable elements non site tree parent has an empty CMSEditLink, then the form for the edit form cannot be created.

The current behaviour is the user will be redirected to the home page on the frontend. Instead, a hard exception should be thrown.

## PRs
- https://github.com/silverstripe/silverstripe-elemental/pull/1408
- https://github.com/silverstripe/developer-docs/pull/871

PR #1408, opened 2026-01-28
ENH Throw exception if empty parent CMSEditLink

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-elemental/issues/1407

Requires PRs in https://github.com/silverstripe/silverstripe-elemental/issues/1405 to be merged and merged up for behat to pass

### DIFF
--- a/src/Models/BaseElement.php
+++ b/src/Models/BaseElement.php
@@ -8,6 +8,7 @@ use DNADesign\Elemental\Services\ReorderElements;
 use DNADesign\Elemental\Extensions\TopPageElementExtension;
 use Exception;
 use InvalidArgumentException;
+use RuntimeException;
 use SilverStripe\CMS\Controllers\CMSPageEditController;
 use SilverStripe\CMS\Model\SiteTree;
 use SilverStripe\Control\Controller;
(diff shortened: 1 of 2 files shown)


GuySartorelli (changes requested):
With the exception in that spot, it will be thrown for any of these public method calls:

- `getCMSEditLink()`
- `getEditLink()`
- `getBlockSchema()`

That will happen no matter why they're called, and regardless of whether there even is a parent record for the block at the time of the call.

Probably the exception should be thrown from `ElementalAreaController` or some other place that's closer to the place where the link is actually needed.

Please also use `LogicException` instead of `RuntimeException`: https://www.php.net/manual/en/class.logicexception.php
> Exception that represents error in the program logic. This kind of exception should lead directly to a fix in your code.

GuySartorelli:
Thinking more about this actually - there's some edge cases we need to square away as well. For example what if you move a block to a new parent, but that new parent doesn't have a edit link? Suddenly we're throwing an exception so the elemental area probably won't load. You can't really catch that at dev time.

I know I originally suggested an exception but perhaps it would be better to instead do a combination of logging an error, and displaying an error on the block. That way we don't have to consider every possible edge case, and content authors won't have elemental areas failing to load.

GuySartorelli:
Closing for now as the attached card was moved to refinement

Issue #1378, opened 2025-10-03
Accessibility: Cannot move elements with space


emteknetnz:
Moved from https://github.com/silverstripe/silverstripe-elemental/issues/1374

Pressing space on a block to activate keyboard reordering expands the block. This means it is no longer possible to reorder blocks using the keyboard.

~Note: Since this is a bug for existing functionality, it should be fixed in patches where appropriate.~

Update: I don't think this is a regression, changing elemental to 5.0 and running `yarn dev` it still has the same behaviour i.e. space will open the inline edit form for the element, not activate 'move mode'

## PRs
- https://github.com/silverstripe/silverstripe-elemental/pull/1412
- https://github.com/silverstripe/developer-docs/pull/873

GuySartorelli:
PR(s) merged. This will be included in the next minor release.

PR #1412, opened 2026-02-02
ENH Allow reordering elements with space

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-elemental/issues/1378

I have no idea why the behat test for PHP 8.5 MySQL 8.4 is failing, using that setup locally behat passes. The same test for PHP 8.3 MySQL 8.0 passes in CI. I've tried rerunning multiple times though it fails. The screenshot in CI shows some sort of validation error which is unrelated to any work done here


client/src/components/ElementEditor/Element.js
```diff
@@ -463,16 +473,21 @@ Element.propTypes = {
   type: elementTypeType.isRequired,
   areaId: PropTypes.number.isRequired,
   link: PropTypes.string.isRequired,
+  HeaderComponent: PropTypes.elementType,
+  ContentComponent: PropTypes.elementType,
   // Redux mapped props:
   activeTab: PropTypes.string,
   tabSetName: PropTypes.string,
   onActivateTab: PropTypes.func,
-  isOver: PropTypes.bool.isRequired,
```
emteknetnz, line 470: isOver is not a prop - in Element.js it refers to a var that comes from the useSortable() hook - possible it being in propTypes is a left-over from the old dnd implementation before we switched to dnd-kit

client/src/components/ElementEditor/Element.js
```diff
@@ -360,17 +384,16 @@ const Element = (props) => {
...
     className={elementClassNames}
     onClick={handleExpand}
     onKeyUp={handleKeyUp}
     role="button"
     tabIndex={0}
     title={getLinkTitle(type)}
     key={element.id}
-    // sortable properties
     ref={setNodeRef}
-    {...attributes}
```
GuySartorelli, line 372: Why is `...attributes` being moved into the child element?

client/src/components/ElementEditor/Element.js
```diff
@@ -463,16 +490,21 @@ Element.propTypes = {
   type: elementTypeType.isRequired,
   areaId: PropTypes.number.isRequired,
   link: PropTypes.string.isRequired,
+  HeaderComponent: PropTypes.elementType,
+  ContentComponent: PropTypes.elementType,
```
GuySartorelli: Are these unrelated changes?

client/src/components/ElementEditor/Element.js
```diff
@@ -463,16 +490,21 @@ Element.propTypes = {
...
-  isOver: PropTypes.bool.isRequired,
+  onChangeHasUnsavedChanges: PropTypes.func.isRequired,
   saveElement: PropTypes.bool.isRequired,
   onBeforeSubmitForm: PropTypes.func.isRequired, // eslint-disable-line react/no-unused-prop-types
   onAfterSubmitResponse: PropTypes.func.isRequired,
   // Used to ensure form gets re-rendered on submission so it can be submitted again if there are validation errors
   increment: PropTypes.number.isRequired,
+  submitForm: PropTypes.func,
+  actions: PropTypes.object,
+  formDirty: PropTypes.bool,
```
GuySartorelli: Are these unrelated changes?
When making unrelated changes please both point them out and explain why the changes are made - otherwise I have to piece it together and build context from scratch that you could provide.

From the looks of it in this case these props already existed but were missing as declarations.
Are we sure none of these are required props?

tests/Behat/features/element-editor.feature
```diff
@@ -154,3 +156,38 @@ Feature: View types of elements in an area on a page
...
+    And I press the "Tab" key globally
+    And I press the "Tab" key globally
+    And I press the "Tab" key globally
+    And I press the "Tab" key globally
+    Then the ".element-editor-header__drag-handle" element should have focus
+    # Move the element with space
+    When I press the "Space" key globally
+    And I press the "Down" key globally
+    When I press the "Space" key globally
+    Then I should see "My other" in the ".element-editor-header__title" element
```
GuySartorelli: Can we assert something about "block 1" and "block 2" instead? That will be more robust and is easier to see what we're actually asserting just by reading the feature file.

tests/Behat/features/element-editor.feature
GuySartorelli, line 1: Can you please also add a behat test for this:
- I expand the inline editable block
- I make a change (but do not save yet)
- I use the drag handler to move the block

The current state of affairs is:
- the block moves
- the block is "modified"
- the unsaved changes are still unsaved - and still present in the inline form

We should have a test against regressions, because it would be easy to refactor in such a way that it saves the unsaved changes, which is not desirable.

tests/Behat/features/element-editor.feature
GuySartorelli: Please also assert something about the state of the block. i.e. if it was published, it should become modified.

client/src/components/ElementEditor/Element.js
```diff
@@ -268,9 +286,14 @@ const Element = (props) => {
...
   const handleKeyUp = (event) => {
     const { nodeName } = event.target;
+    const dragHandle = event.target.closest
+      ? event.target.closest('.element-editor-header__drag-handle')
+      : null;
     if ((event.key === ' ' || event.key === 'Enter')
       // Ignore presses while focusing inputs and textareas
       && !['input', 'textarea'].includes(nodeName.toLowerCase())
+      // Ignore presses while focused on the drag handle
+      && !dragHandle
```
GuySartorelli, line 303: This doesn't seem to be working - I can both expand inline-editable blocks as well as navigate to non-inline-editable blocks by clicking the drag handler.

Please also add a behat test to catch regressions of this.

client/src/components/ElementEditor/ElementList.js
```diff
@@ -35,6 +110,16 @@ function ElementList({
   const [increment, setIncrement] = useState(0);
   const [hasUnsavedChangesBlockIDs, setHasUnsavedChangesBlockIDs] = useState({});
   const [validBlockIDs, setValidBlockIDs] = useState({});
+  // Keyboard drag-drop ends on keydown; the following keyup can bubble to the element container
+  // and wrongly toggle inline edit, so we suppress that keyup for a short window.
+  const keyboardDragActiveRef = useRef(false);
+  const suppressKeyUpUntilRef = useRef(0);
+  const keyupSuppressWindowMs = 300;
```
GuySartorelli: 300ms is way too long. There's nothing like network requests etc that can take longer on weak computers, the two events fire one after another pretty quickly. We should be able to have a very short fuse on this.
```suggestion
  const keyupSuppressWindowMs = 50;
```

That said, I don't understand why this is needed anyway. Key up events on the drag handler are meant to skip the `handleExpand` call altogether.

client/src/components/ElementEditor/ElementList.js
```diff
@@ -35,6 +110,16 @@ function ElementList({
...
+  // Ref to the list container DOM node, used to attach a native capture-phase keyup listener
```
GuySartorelli: What is a "native capture-phase keyup listener"?
- "keyup listener" I understand.
- "native" I assume means vanilla JS, though I don't see why we'd want/need that (see comment below)
- "capture-phase" lost me completely.

client/src/components/ElementEditor/ElementList.js
```diff
@@ -35,6 +110,16 @@ function ElementList({
...
+  // that suppresses Space/Enter keyup events immediately after a keyboard drag ends.
+  // A React onKeyUpCapture handler would not work here because stopImmediatePropagation() is
+  // needed to prevent other native listeners (including React's delegated handler) from firing.
```
GuySartorelli: I don't understand what the goal is here. If it's to prevent `handleKeyUp()` in `Element.js`, then:
1. `event.stopPropagation()` should prevent it from bubbling, since in both cases it should be a react synthetic event
2. that handler has code that should prevent key events on the drag handler from doing anything anyway.

client/src/components/ElementEditor/ElementList.js
```diff
@@ -139,8 +224,59 @@ function ElementList({
...
     }),
+    useSensor(KeyboardSensor, {
+      coordinateGetter: keyboardCoordinateGetter
+    })
   );
 
+  // Capture phase ensures the keyup is cancelled before React's handlers on the element container,
+  // preventing the inline edit form from toggling open when Space/Enter is released after a
+  // keyboard drag-drop operation.
+  useEffect(() => {
```
GuySartorelli: I haven't reviewed the code in this effect, pending discussions further up that could render it unnecessary.

client/src/components/ElementEditor/Header.scss
```diff
@@ -84,14 +84,26 @@
...
     position: absolute;
     left: 5px;
     cursor: grab;
+
+    &:focus,
+    &:focus-visible {
+      opacity: 1;
+      pointer-events: auto;
+      outline: 2px solid #2A62BC;
+      outline-offset: 2px;
```
GuySartorelli: Why are you setting these explicitly? We have CSS variables for this which should automatically apply.

client/src/components/ElementEditor/Header.scss
```diff
@@ -84,14 +84,26 @@
...
   &__drag-handle {
-    display: none;
+    display: block;
+    opacity: 0;
+    pointer-events: none;
     position: absolute;
     left: 5px;
     cursor: grab;
+
+    &:focus,
```
GuySartorelli:
```suggestion
```
Don't use `:focus` for styling the focus outline - let the browser decide when styling should be applied by only using `:focus-visible`

GuySartorelli (changes requested):
CI failure is unrelated - it's also happening on the `6` branch already: https://github.com/silverstripe/silverstripe-elemental/actions/runs/21850125219/job/63054653309

client/src/components/ElementEditor/Element.js
emteknetnz: DnD kit expects its `attributes` (aria, tabIndex, data-* flags) to be applied to the activator element. Moving them to the drag handle keeps keyboard drag activation scoped to that handle (not the whole block), avoids Space/Enter on the block container being interpreted as a drag action, and keeps the outer container free to handle expand/navigation.

client/src/components/ElementEditor/Element.js
emteknetnz: Yeah they are sorry should have pointed out

These props already exist via injector and were used before adding them to make the API explicit.

client/src/components/ElementEditor/Element.js
emteknetnz: Yeah unrelated also

These props already exist and are used by the component from submit/toast/dirty-state flows

> Are we sure none of these are required props?

I don't know, though since it's unrelated I'd rather just remove the propTypes then enforce .isRequired.

client/src/components/ElementEditor/ElementList.js
emteknetnz: The suppression window exists becuase keyboard drag-drop completes on keydown but the keyup that follows can bubble to the block container and trigger `handleKeyUp()` andopen inline edit even though the drag handle was the activator

This isn't related to network request cos it is purely about the keyup event timing. A 50ms window might be to short cos typical key press duration might exceed that and 300ms is still short enough not to affect normal interactions and reliably catches the post-drag keyup - 300ms is kind of the 'default' ms delay where it's a "delay but not annoying"

client/src/components/ElementEditor/ElementList.js
emteknetnz: Updated comment

client/src/components/ElementEditor/ElementList.js
emteknetnz: event.stopPropagation() in a React handler is too late here because Reacts synthetic events are delegated at the document level so by the time the container’s `onKeyUp` runs, the event has already bubbled through native listeners

We need to stop the native event during capture and use `stopImmediatePropagation()` when available so React’s delegated `onKeyUp` for the element container never fires otherwise handleKeyUp() still runs even when the drag handle initiated the keyboard drag

client/src/components/ElementEditor/Header.scss
emteknetnz: Removed

client/src/components/ElementEditor/Header.scss
emteknetnz: Updated to use css variables

client/src/components/ElementEditor/Element.js
emteknetnz: Done, have added behat test

tests/Behat/features/element-editor.feature
emteknetnz: Done

tests/Behat/features/element-editor.feature
emteknetnz: Done

client/src/components/ElementEditor/Element.js
GuySartorelli: Please remove the new proptype declarations if you're not spending the time to identify whether they're required or not. Small unrelated changes are okay but they do need to be done with the same rigour as though they were being done in their own PR.

tests/Behat/features/element-editor.feature
```diff
@@ -154,3 +169,70 @@ Feature: View types of elements in an area on a page
...
+    And the "page" "Blocky" has a "Block 2" content element with "My other" content
+    And I am logged in as a member of "AUTHOR" group
+    And I go to "/admin/pages"
+    And I follow "Blocky"
+    When I press the "Publish" button
+    Then I should see a "Published Page "Blocky"" success toast
+    And I wait for 1 seconds
+    # Note that :nth-of-type() selectors are x2 what you'd expect them to be, (2) is the 1st block
+    Then I should not see a ".element-editor__element:nth-of-type(2) .badge" element
+    # Assert the title of the first content block
```
GuySartorelli:
```suggestion
    # Assert the order is correct
```

tests/Behat/features/element-editor.feature
```diff
@@ -154,3 +169,70 @@ Feature: View types of elements in an area on a page
...
+    And I press the "Tab" key globally
+    Then the ".element-editor-header__drag-handle" element should have focus
+    When I press the "Space" key globally
+    And I press the "Down" key globally
+    And I press the "Space" key globally
+    Then the element order should be "Bob's Block,Alice's Block"
+    # Expand the block again as it collapsed during move
+    And I press the "Shift-Tab" key globally
+    And I press the "Space" key globally
+    And the "Content" field for block 2 should contain "<p>Unsaved change</p>"
```
GuySartorelli, line 238: Please reload the page after this to show that the changes are indeed not saved.

client/src/components/ElementEditor/Header.scss
```diff
@@ -84,14 +84,25 @@
...
+    pointer-events: none;
     position: absolute;
     left: 5px;
     cursor: grab;
+
+    &:focus-visible {
+      opacity: 1;
+      pointer-events: auto;
+      outline: var(--focus-outline-width) var(--focus-outline-style) var(--focus-outline-color);
+      outline-offset: var(--focus-outline-offset);
```
GuySartorelli:
```suggestion
```
As I mentioned earlier, this should automatically apply. You shouldn't need to explicitly declare this unless the existing styling for this state needs to be overridden for some reason.

client/src/components/ElementEditor/ElementList.js
GuySartorelli: > the event has already bubbled through native listeners

What native listeners are causing problems here? I tried removing the code related to this temporary event propagation prevention but couldn't see anything unexpected happening at all. I'm really struggling to understand what problem this added complexity is trying to solve.

GuySartorelli (changes requested):

tests/Behat/features/element-editor.feature
emteknetnz: I don't know how to test this in behat, the 'you have unsaved changes' alert messes up the test, and adding "When I dismiss the dialog" doesn't fix. I think it's fine to leave untested (this is a keyboard nav test after all, the exact same functionality was untested previously with mouse drag and drop)

tests/Behat/features/element-editor.feature
emteknetnz: Done

emteknetnz:
> Please remove the new proptype declarations if you're not spending the time to identify whether they're required or not. Small unrelated changes are okay but they do need to be done with the same rigour as though they were being done in their own PR.

Done

client/src/components/ElementEditor/ElementList.js
emteknetnz: Weeeird, you're right the code isn't needed. I struggled for a long time with 'after moving the block the inline edit for would open', though I guess I resolved with a different code fix. Have removed all this extra code.

client/src/components/ElementEditor/Element.js
emteknetnz: Removed

tests/Behat/features/element-editor.feature
GuySartorelli: You need "When I confirm the dialog" - confirm is "ok", dismiss is "cancel".
```cucumber
    When I go to "/admin/pages"
    And I confirm the dialog
    And I follow "Blocks Page"
    Then I should see "Some content" as the summary for block 2
```

You also need to add the `@modal` tag to the scenario so it's not marked as an "unexpected alert".

GuySartorelli (changes requested):
https://github.com/silverstripe/silverstripe-elemental/pull/1412#discussion_r2795560337 was not addressed

client/src/components/ElementEditor/Header.scss
emteknetnz: Removed, I've kept the border-radius rule though otherwise it'll have ugly square corners

tests/Behat/features/element-editor.feature
```diff
@@ -154,3 +169,76 @@ Feature: View types of elements in an area on a page
...
+    # Move the element down and refresh the page to confirm it saved correctly after moving
+    When I press the "Enter" key globally
+    When I press the "Down" key globally
+    When I press the "Enter" key globally
+    Then the element order should be "Block 2,Block 1"
+    And I go to "/admin/pages"
+    And I follow "Blocky"
+    Then the element order should be "Block 2,Block 1"
+
+  @unsavedChanges @modal @sboyd
```
GuySartorelli:
```suggestion
  @unsavedChanges @modal
```
I wonder if we should add linting to catch these :sweat_smile:

GuySartorelli (changes requested):

tests/Behat/features/element-editor.feature
emteknetnz: Fixed

GuySartorelli (approved):
LGTM

